PR #107, opened 2020-02-24
Add dex score tiebreaker to initiative

### LINKED ISSUE #105: Add dex tiebreaker to initiative (optional)
In the 5e character sheets in roll 20, there is an option to add your dexterity score to initiative.

Based on my current experimenting, I haven't found a way to ensure that this also gets added to initiative rolls through Beyond20. Probably the easiest thing to do would be to add an option to add this within Beyond20. Another valid implementation would be to the check if the setting is enabled in roll 20 on the target token's sheet, but this seems more complicated.

To be explicit, how this works is that it adds your dexterity score as a decimal after your rolled score:
If you roll a 5 and have a 12 dexterity score = (5 + 1 + .12) = 5.12

This makes it easier to sort if your DM uses this house rule.

jaypoulz:
Closes #105

### DIFF
--- a/src/dndbeyond.pyj
+++ b/src/dndbeyond.pyj
@@ -613,9 +613,19 @@ class Monster(CharacterBase):
         for ability in self._abilities:
             if ability[1] == "DEX":
                 modifier = ability[3]
-                sendRoll(self, "initiative", "1d20" + modifier, {"initiative": modifier})
-                break
 
+                initiative = modifier
+                if self.getGlobalSetting("initiative-tiebreaker", False):
+                    tiebreaker = ability[2]
+
+                    # Add tiebreaker as a decimal
+                    initiative = float(initiative) + float(tiebreaker) / 100
+
+                    # Render initiative as a string that begins with '+' or '-'
+                    initiative = '+' + str(initiative) if initiative >= 0 else str(initiative)
+
+                sendRoll(self, "initiative", "1d20" + initiative, {"initiative": initiative})
+                break
 
     def rollSavingThrow(self, abbr):
         mod = self._saves[abbr]
--- a/src/dndbeyond.pyj
+++ b/src/dndbeyond.pyj
@@ -233,7 +233,7 @@ class Character(CharacterBase):
             hp = max_hp = temp_hp = None
             hp_items = $(".ct-health-summary__hp-group--primary .ct-health-summary__hp-item")
             for item in hp_items:
-                label = $(item).find(".ct-health-summary__hp-item-label").text() 
+                label = $(item).find(".ct-health-summary__hp-item-label").text()
                 if label == "Current":
                     # Make sure it's not an input being modified
                     number = $(item).find(".ct-health-summary__hp-item-content .ct-health-summary__hp-number")
@@ -303,7 +303,7 @@ class Character(CharacterBase):
             for option in options:
                 option_name = option.childNodes[0].textContent.trim()
                 feature_list.append(feat_name + ": " + option_name)
-                
+
         #console.log(name, feature_list)
         return feature_list
 
@@ -317,7 +317,7 @@ class Character(CharacterBase):
                 update = True
         elif self.getSetting("class-features", None):
             self._class_features = list(self.getSetting("class-features", None))
-            
+
         race_detail = $(".ct-features .ct-race-detail")
         if race_detail.length > 0:
             self._racial_traits = self.featureDetailsToList(race_detail, "Racial Traits")
@@ -350,7 +350,7 @@ class Character(CharacterBase):
                                          "racial-traits": self._racial_traits,
                                          "feats": self._feats,
                                          "actions": self._actions})
-            
+
     def hasClassFeature(self, name):
         return name in self._class_features
     def hasRacialTrait(self, name):
@@ -533,7 +533,7 @@ class Monster(CharacterBase):
                                 self.rollInitiative()
                         addIconButton(cb, initiative.find(base + "__attribute-data"))
 
-        
+
         tidbits = stat_block.find(base + "__tidbits " + base + "__tidbit")
         for tidbit in tidbits:
             label = $(tidbit).find(base + "__tidbit-label").text()
@@ -586,7 +586,7 @@ class Monster(CharacterBase):
                         data.append(skill + " " + self._skills[skill])
                         if add_dice:
                             addIconButton(makeCB(skill), data, append=True)
-                        
+
             elif label == "Challenge":
                 self._cr = value.split(" ")[0]
             self._tidbits[label] = value
@@ -708,7 +708,7 @@ class Monster(CharacterBase):
             if save:
                 roll_properties["save-ability"] = save[0]
                 roll_properties["save-dc"] = save[1]
- 
+
         if attackInfo or hitInfo:
             return roll_properties
         else:
@@ -800,7 +800,7 @@ class Monster(CharacterBase):
 
     def injectSpellRolls(self, element, url):
         icon16 = chrome.extension.getURL("images/icons/icon16.png")
-        
+
         roll_icon = $('<img class="ct-beyond20-spell-icon" x-beyond20-spell-url="' + url + '"></img>')
 
         $(element).after(roll_icon)
@@ -836,7 +836,7 @@ class Monster(CharacterBase):
         hp = max_hp = temp_hp = None
         groups = $(".ct-creature-pane .ct-collapsible__content .ct-creature-pane__adjuster-group")
         for item in groups:
-            label = $(item).find(".ct-creature-pane__adjuster-group-label").text() 
+            label = $(item).find(".ct-creature-pane__adjuster-group-label").text()
             if label == "Current HP":
                 hp = int($(item).find(".ct-creature-pane__adjuster-group-value").text())
             elif label == "Max HP":
--- a/src/dndbeyond_character.pyj
+++ b/src/dndbeyond_character.pyj
@@ -77,7 +77,7 @@ def rollInitiative():
         roll_properties["advantage"] = RollType.ADVANTAGE
     sendRollWithCharacter("initiative", "1d20" + initiative, roll_properties)
 
-        
+
 def rollHitDie(multiclass, index):
     #console.log("Rolling hit die index " + index)
     hitdie = $(".ct-reset-pane__hitdie").eq(index)
@@ -264,7 +264,7 @@ def rollItem(force_display=False):
         roll_properties["item-type"] = item_type
         if critical_limit != 20:
             roll_properties["critical-limit"] = critical_limit
-            
+
         # Asssassinate: consider all rolls as critical
         if character.hasClassFeature("Assassinate") and \
            character.getSetting("rogue-assassinate", False):
@@ -275,7 +275,7 @@ def rollItem(force_display=False):
     else:
         sendRollWithCharacter("item", 0, {"name": item_name,
                              "description" : description,
-                             "item-type": item_type})        
+                             "item-type": item_type})
 
 def rollAction(paneClass):
     nonlocal character
@@ -291,7 +291,7 @@ def rollAction(paneClass):
         sendRollWithCharacter("custom", superiority_die, {"name": action_name,
                                                           "description": description,
                                                           "modifier": superiority_die})
-                                                          
+
     elif action_name == "Bardic Inspiration" or action_parent == "Blade Flourish":
         bard_level = character.getClassLevel("Bard")
         inspiration_die = "1d6" if bard_level < 5 else ("1d8" if bard_level < 10 else ("1d10" if bard_level < 15 else "1d12"))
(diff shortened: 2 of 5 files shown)


jaypoulz:
@kakaroto Do you have a test strategy for new changes? I can test on FF & Chrome for r20, but I don't have access to Foundry VTT.

kakaroto:
Hey @jaypoulz, thanks for the PR and taking care of the issue #105.
I gave it a quick look and it generally looked good, other than some differences in style and `console.log` that need to be removed.
For testing strategy, I mostly test on one system (roll20 or FVTT) in Chrome and once the feature is done, I test on the other system if there was a need to write system specific changes, then I only go through a quick "let's make sure nothing is broken in Firefox" before a release.
For testing how it would be on FVTT, you can just close your Roll20 tab or click the quick settings there and set it to send rolls to the D&D Beyond dice roller, it's the same code/renderer between FVTT and the integrated Dice roller.
I'll do a quick review of the PR itself so you can fix things before you send an update and we get it merged.
Thanks again!

src/dndbeyond.pyj
```diff
@@ -613,7 +613,9 @@ class Monster(CharacterBase):
         for ability in self._abilities:
             if ability[1] == "DEX":
                 modifier = ability[3]
-                sendRoll(self, "initiative", "1d20" + modifier, {"initiative": modifier})
+                score = ability[2]
+                roll_modifier = float(modifier) + float(score / 100) if self.getGlobalSetting("initiative-with-dex-score", False) else modifier
```
kakaroto: Due to the length of the line, please split it into :
```python
if self.getGlobalSetting("initiative-with-dex-score", False):
  score = ability[2]
  roll_modifier = float(modifier) + float(score / 100)
else:
  roll_modifier = modifier
```
to make it easier to read. I usually reserve the one-liner `x if y else z` constructs for simpler lines only.

src/dndbeyond_character.pyj
```diff
@@ -66,12 +66,15 @@ def rollSavingThrow():
...
 def rollInitiative():
+    nonlocal character, settings
     initiative = $(".ct-initiative-box__value").text()
     advantage = $(".ct-initiative-box__advantage").length > 0
     if initiative == "":
         initiative = $(".ct-combat-mobile__extra--initiative .ct-combat-mobile__extra-value").text()
         advantage = $(".ct-combat-mobile__advantage").length > 0
-    #console.log("Initiative " + ("with" if advantage else "without") + " advantage : " + initiative)
+    console.log(character)
+    initiative = initiative if not settings.getSetting("initiative-with-dex-score", False) else initiative
```
kakaroto: Same as above, split the if into a non-tertiary operator.
Also, it should be `character.getGlobalSetting`, I think `settings` is only a dictionary, so `settings.getSetting` would not work.
In this case, you can also get the DEX abilities using : 
```python
   if character.getGlobalSetting("initiative-with-dex-score")
     tiebreaker = 0
     for ability in character._abilities:
        if ability[1] == "DEX":
           tiebreaker = ability[2]
```
And finally, remove the `console.log` lines

jaypoulz:
Awesome - I'll clean up the PR once I've implemented the character sheet based initiative roll, test it, and ping you for review when its ready. Thanks!

src/roll20.pyj
```diff
@@ -49,7 +49,8 @@ def escapeRoll20Macro(text):
 def genRoll(dice, modifiers={}):
     roll = "[[" + dice
     for m in modifiers:
-        mod = modifiers[m]
+        mod = str(modifiers[m])
```
kakaroto: shouldn't be necessary to cast it to a string as it should always be sent as a string from the dndbeyond side

src/dndbeyond.pyj
kakaroto: Seeing the later cast of the modifier into a string from the roll20 side, this reminds me that the roll modifier here should be a string, and not a float, so make sure you cast it into a string in here.

src/roll20.pyj
```diff
@@ -282,6 +284,7 @@ def rollInitiative(request, custom_roll_dice=""):
             dice = ROLL20_INITIATIVE_ADVANTAGE_QUERY
         else:
             dice = "1d20"
+        console.log(request.initiative)
```
kakaroto: remember the cleanup the `console.log` lines before final PR

src/settings.pyj
```diff
@@ -85,6 +85,11 @@ options_list = {
                            "type": "bool",
                            "default": True},
 
+    "initiative-with-dex-score": {"title": "Add dexterity score tiebreaker to initiative roll.",
```
kakaroto: I would rename the setting into `initiative-tiebreaker` rather than `initiative-with-dex-score` (what it does, rather than how it does it) and adjust the title accordingly. I'm not super good with naming stuff, but my suggestion would be : 
- title: Add tie breaker to initiative rolls.
- description: Adds the dexterity score as a decimal to initiative rolls to break any initiative ties.

src/dndbeyond.pyj
jaypoulz: Do you prefer the stylistic:
```
initiative = modifier
if self.getGlobalSetting("initiative-with-dex-score", False):
  score = ability[2]
  initiative = str(float(modifier) + float(score / 100))
```

src/dndbeyond_character.pyj
jaypoulz: Would you rather me fix this in both dndbeyond.pyj & dndbeyond_character.pyj (as is the current approach) or did you want me to have dndbeyond_character.pyj call the parent class method and pass along the initiative modifier?

src/dndbeyond.pyj
kakaroto: yes, actually, I do prefer it the way you just described. Leaving optional settings to just change the behavior rather than define it. Especially since the default would be `initiative = modifier`
Good catch. Thanks!

src/dndbeyond_character.pyj
kakaroto: The code is honestly a mess, so there isn't really any 'parent method' in this case. The 'rollInitiative' in dndbeyond.pyj is a method that exists only for the `Monster` class (the only common functions are in `CharacterBase` which is the `getSetting` and `getGlobalSetting` I think), so the change has to go in both files/classes.

jaypoulz: Ah, OK. Is there any value in doing a rolling refactor such that we move this function into CharacterBase with the intention of fixing the others as you go? Or would you rather keep them the way they are so that you don't find yourself in a 1/2 way refactored state?

kakaroto: A refactor at this point would be more of a rewrite. Statblocks and character sheets are also very different in D&D Beyond. A roll Initiative for a monster is about simply adding dexterity, but also modifying the statblock to add my own 'Roll Initiative' button for convenience.
For a character, it's about checking the initiative value in the sidebar, verifying if there are 'advantage on initiative' modifiers due to class features, etc.. 
The way the two work is too different to have anything merged. If you had a common function name for rolling initiative, you'd just have them override it both, so it's not worth pursuing at this time.

Note: Beyond20 was initially written as a thought experiment, a "would it be possible to... ?" question, from before I knew how to write javascript (explains the .pyj) and from before I knew html/css (I still don't). It worked so easily and quickly, and got so much traction that it quickly evolved, but the code is unfortunately still very messy. I apologize for that :)

jaypoulz: I think you've described the origin story for just about any project that's ever come to be. :)

No need for apologies - it works! That's a testament to itself!

Rewrites can be very good and very bad. If you actually have the time (and energy) you need to follow through, then it could become the new base and you could incorporate all you learned. However, its easier to iterate on working code then to hack on vaporware. Also, the longer you wait to rewrite, the more you'll have to write to achieve feature parity. Well-written software can often lose to worse written software that was good enough to work when people needed it.

You're the maintainer, so I obviously want to defer to your vision.

It might make sense to move towards a Model-View-Controller type organization. The abstraction I would make would be that characters and monsters are special cases of a creature, similar to how the game is structured. Creatures (models) would have abilities, speed, type, size, etc and corresponding (controller) methods to interact with them (e.g. rollInitiative). Your "views" would be the code that inject the buttons into the page. You would also need controllers to parse the dndbeyond page specifics.

I don't think the underlying tech you pick matters all that much once you have a portable/modular architecture.

jaypoulz:
@kakaroto ready for review!
Tested for on roll20 and using no-VTT renderer on Chrome.
Still need to sanity check ffox.

jaypoulz:
Also I didn't add the compiled js since my rapydscrypt seems to compile differently. One suggestion would be to build a release branch and only keep the compiled js in that. This way you can keep master clean for development, and you can still have the compiled js in the downloadable releases.

src/dndbeyond_dice.pyj
```diff
@@ -188,6 +188,7 @@ class DNDBRoll(Beyond20BaseRoll):
                 self._total += part.roll()
             else:
                 self._total += part
+        self._total = Math.round((self._total + Number.EPSILON) * 100) / 100
```
kakaroto: Why are you adding epsilon here?

src/dndbeyond.pyj
```diff
@@ -613,9 +613,14 @@ class Monster(CharacterBase):
         for ability in self._abilities:
             if ability[1] == "DEX":
                 modifier = ability[3]
-                sendRoll(self, "initiative", "1d20" + modifier, {"initiative": modifier})
-                break
 
+                initiative = modifier
+                if self.getGlobalSetting("initiative-tiebreaker", False):
+                    score = ability[2]
+                    initiative= str(float(modifier) + float(score) / 100)
```
kakaroto: Space missing between `initiative` and `=`

src/dndbeyond_character.pyj
```diff
@@ -66,12 +66,18 @@ def rollSavingThrow():
...
+    nonlocal character
     initiative = $(".ct-initiative-box__value").text()
     advantage = $(".ct-initiative-box__advantage").length > 0
     if initiative == "":
         initiative = $(".ct-combat-mobile__extra--initiative .ct-combat-mobile__extra-value").text()
         advantage = $(".ct-combat-mobile__advantage").length > 0
     #console.log("Initiative " + ("with" if advantage else "without") + " advantage : " + initiative)
+
+    if character.getGlobalSetting("initiative-tiebreaker", False):
+       dexterity_score = [x[2] for x in character._abilities if x[1] == 'DEX'][0]
```
kakaroto: This will crash if there is no page DEX ability, which could happen if user loads their sheet on a tab other than the abilities (not entirely sure, but i remember there were cases when it wouldn't find abilities, that's why it always checks elsewhere uf it finds it in the abilities array)

jaypoulz:
I can split the ws fix with the other commit - that's just me being lazy since my emacs is set to autostrip white space. Splitting the commit will take me just a moment.

As for the missing dex score problem - is there an alternative way of getting the dex score if it's missing from _abilities? If not, this sounds like I should just pass initiative as is if I encounter this situation.

kakaroto:
No problem. I use `git add -p` for whenever I want to commit something, allows me to review my stuff as well. But like I said, it's not a big deal since the changes are small.
For the dex, it was in my original comment : https://github.com/kakaroto/Beyond20/pull/107#discussion_r383442418
```python
 if character.getGlobalSetting("initiative-with-dex-score")
     tiebreaker = 0
     for ability in character._abilities:
        if ability[1] == "DEX":
           tiebreaker = ability[2]
```
That way if DEX isn't found, the tiebreaker is 0, and use that variable for the math.

I just saw what `Number.EPSILON` is (I thought it was Euler's number 2.71828), so I guess it makes no difference to the actual value to have it added, but I still don't understand the need to do it, along with the `* 100 / 100`.
If it's because JS causes the number to be `10.1200000000001` for example, instead of `10.12`, then how about doing instead : `parseFloat(self._total.toFixed(2))` with a comment above it to explain why it needs to be there?

src/dndbeyond_dice.pyj
jaypoulz: I actually made a note of this in the commit summary.
This is a rounding trick to ensure that edge cases are rounded correctly.
https://stackoverflow.com/questions/11832914/round-to-at-most-2-decimal-places-only-if-necessary

It's probably not necessary in this case.

kakaroto:
Ah, I just saw the commit log message that links to https://stackoverflow.com/questions/11832914/round-to-at-most-2-decimal-places-only-if-necessary
 (sorry for missing it yesterday, I did the review on my phone while I was out).
That makes sense, and it's good, it can stay as is, but I would put the stackoverflow link in a comment above that line, as it would be easier for the future to figure out when reading the code rather than trying to find the the link from the commit which added the line.
Thanks!

kakaroto:
Thanks for all the changes. Nicely done. I like the `reduce` solution for the ability modifier.
I've merged it now, thanks! I hope to see more from you in the future :)

jaypoulz:
Ready for final review!

- Separated the commits by whitespace & content.
- Removed the epsilon bit since it's covering edge cases we aren't going to hit
- Added some logic to ensure we render the same with and without the flag set.
- Tested {-x,0,+x} for characters & monsters are both d&d beyond and the beyond20/vtt renderer {Chrome, FFox}

jaypoulz:
LOL you beat me to it. :) Thanks!

kakaroto:
Yeah, I saw it pop in the discord #bots channel, so I did the quick review and merge :)

Issue #171, opened 2020-04-07
donate icon too big in popout outside of dndb


kakaroto:
![image](https://user-images.githubusercontent.com/27990/78691161-54d00c80-78c6-11ea-98da-d786053ba2cc.png)

harrisrl2023:
Everything should be working now

PR #172, opened 2020-04-08
Fixed large donate icon

harrisrl2023:
Fixed this issue. https://github.com/kakaroto/Beyond20/issues/171
It seems that the wrong size image was used. I tested it out with chrome developer and everything looks good


kakaroto:
Thank you @harrisrl2023 for the contribution. I appreciate you trying to help with fixing the open issues.

Unfortunately, I can't accept this PR because you've only made changes to the .js file, while Beyond20 is written in rapydscript and is generated from the .pyj files, so your changes would be overwritten the next time I recompiled it.
Also, the image was the 64x64 because that's what it needs to be in the settings (see `settings.pyj`) but it only needs to be modified for the popups.

I have pushed the appropriate fix earlier today which you can see here : https://github.com/kakaroto/Beyond20/commit/1c0620fd1d81fcef6b3b4a57849be4af304c6592

Thanks again.

PR #192, opened 2020-04-13
Ranger special ability support

Brunhine:
Added support for Ranger ability "Hunter Slayer"

### DIFF
--- a/src/dndbeyond_character.pyj
+++ b/src/dndbeyond_character.pyj
@@ -227,6 +227,9 @@ def rollItem(force_display=False):
             damages.append("1d8")
             damage_types.append("Ambush")
             character.mergeCharacterSettings({"ranger-dread-ambusher": False})
+        if character.hasClassFeature("Hunter’s Prey: Colossus Slayer"):
+            damages.append("1d8")
+            damage_types.append("Colossus Slayer")
         if properties["Attack Type"] == "Melee" and \
            character.hasClassFeature("Improved Divine Smite") and \
            character.getSetting("paladin-improved-divine-smite", True):
--- a/src/dndbeyond_character.pyj
+++ b/src/dndbeyond_character.pyj
@@ -223,13 +223,21 @@ def rollItem(force_display=False):
                     rite_die = "1d10"
                 damages.append(rite_die)
                 damage_types.append("Crimson Rite")
-        if character.getSetting("ranger-dread-ambusher", False):
-            damages.append("1d8")
-            damage_types.append("Ambush")
-            character.mergeCharacterSettings({"ranger-dread-ambusher": False})
-        if character.hasClassFeature("Hunter’s Prey: Colossus Slayer"):
-            damages.append("1d8")
-            damage_types.append("Colossus Slayer")
+
+        #Ranger abilities    
+        if "Ranger" in character._classes:
+            rangerlevel = character.getClassLevel("Ranger")
+            if character.getSetting("ranger-dread-ambusher", False):
+                damages.append("1d8")
+                damage_types.append("Ambush")
+                character.mergeCharacterSettings({"ranger-dread-ambusher": False})
+            if character.hasClassFeature("Hunter’s Prey: Colossus Slayer"):
+                damages.append("1d8")
+                damage_types.append("Colossus Slayer")
+            if character.hasClassFeature("Planar Warrior"):
+                damages.append("1d8" if rangerlevel < 11 else "2d8")
+                damage_types.append("Planar Warrior")
+
         if properties["Attack Type"] == "Melee" and \
            character.hasClassFeature("Improved Divine Smite") and \
            character.getSetting("paladin-improved-divine-smite", True):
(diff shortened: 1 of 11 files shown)


Brunhine:
Recalling to fix prompt text

src/roll_renderer.pyj
```diff
@@ -462,6 +462,12 @@ class Beyond20RollRenderer:
                 elif request.name == "Toll the Dead":
                     ttd_dice = v'await self.queryGeneric(request.name, "Is the target missing any of its hit points?", {"d12": "Yes", "d8": "No"}, "ttd_dice", ["d12", "d8"])'
                     damages[0] = damages[0].replace("d8", ttd_dice)
+                elif damage_types.includes("Colossus Slayer"):
+                    cs_index = damage_types.indexOf("Colossus Slayer")
+                    cs_dice = ttd_dice = v'await self.queryGeneric(request.name, "Is the target missing any of its hit points?", {"d8": "Yes", "0": "No"}, "ttd_dice", ["d8", "0"])'
```
kakaroto, line 467: Other than the prompt which I suggest you to change it to : 
```
Add Colossus Slayer damage?
(It can only be applied once per turn, if enemy has sustained damage)
```
There's also that you did `cs_dice = ttd_dice = ` which should just be `cs_dice` and the 3rd argument to `queryGeneric` is "ttd_dice" when it should be "cs_dice".

src/roll20.pyj
```diff
@@ -362,6 +363,12 @@ def rollAttack(request, custom_roll_dice=""):
         damage_types = list(request["damage-types"])
         crit_damages = list(request["critical-damages"])
         crit_damage_types = list(request["critical-damage-types"])
+
+        # Ranger Colossus Slayer Support
+        if damage_types.includes("Colossus Slayer"):
```
kakaroto, line 368: It might be better to do the indexOf, and check if it's `!= -1` :
```js
    cs_index = damage_types.indexOf("Colossus Slayer")
    if cs_index != -1:
        damages[cs_index] = ROLL20_COLOSSUS_SLAYER_QUERY
```
Also, this being python, it technically should have been `if "Collossus Slayer" in damage_types` but doing it this way is better since I want to move it to JS

src/dndbeyond_character.pyj
```diff
@@ -223,10 +223,21 @@ def rollItem(force_display=False):
...
                 damages.append(rite_die)
                 damage_types.append("Crimson Rite")
-        if character.getSetting("ranger-dread-ambusher", False):
-            damages.append("1d8")
-            damage_types.append("Ambush")
-            character.mergeCharacterSettings({"ranger-dread-ambusher": False})
+
+        #Ranger abilities    
+        if "Ranger" in character._classes:
+            rangerlevel = character.getClassLevel("Ranger")
```
kakaroto, line 229: No need to grab this variable unless it's for Planar Warrior, so better to put it inside the if

src/dndbeyond_character.pyj
```diff
@@ -223,10 +223,21 @@ def rollItem(force_display=False):
...
+            if character.getSetting("ranger-dread-ambusher", False):
+                damages.append("1d8")
+                damage_types.append("Ambush")
+                character.mergeCharacterSettings({"ranger-dread-ambusher": False})
+            if character.hasClassFeature("Hunter’s Prey: Colossus Slayer"):
+                damages.append("1d8")
+                damage_types.append("Colossus Slayer")
+            if character.hasClassFeature("Planar Warrior"):
+                damages.append("1d8" if rangerlevel < 11 else "2d8")
+                damage_types.append("Planar Warrior")
```
kakaroto, line 239: Planar Warrior also changes the damage type to Force, for all weapon damages.
I think this should only apply to the damage types that are standard ("Bludgeoning", "Fire", etc..) not custom ones, like "Sneak Attack" for example.

src/dndbeyond_character.pyj
```diff
@@ -223,10 +223,21 @@ def rollItem(force_display=False):
...
+        if "Ranger" in character._classes:
+            rangerlevel = character.getClassLevel("Ranger")
+            if character.getSetting("ranger-dread-ambusher", False):
+                damages.append("1d8")
+                damage_types.append("Ambush")
+                character.mergeCharacterSettings({"ranger-dread-ambusher": False})
+            if character.hasClassFeature("Hunter’s Prey: Colossus Slayer"):
+                damages.append("1d8")
+                damage_types.append("Colossus Slayer")
+            if character.hasClassFeature("Planar Warrior"):
```
kakaroto, line 237: Planar Warrior also works as a bonus action, so you need to have that toggleable as an option, because it can't always be applied.

src/roll20.pyj
```diff
@@ -362,6 +363,12 @@ def rollAttack(request, custom_roll_dice=""):
         damage_types = list(request["damage-types"])
         crit_damages = list(request["critical-damages"])
         crit_damage_types = list(request["critical-damage-types"])
+
+        # Ranger Ability Support
+        for dmgIndex, dmgType in enumerate(damage_types):
+            if damage_types[dmgIndex] in ["Colossus Slayer", "Planar Warrior"]:
+                damages[dmgIndex] = ROLL20_ADDL_DMG_QUERY.replace(/%1/, damage_types[dmgIndex]).replace(/%2/, damages[dmgIndex])
```
kakaroto, line 370: Not a fan, but I think it's a smart way of doing it as a generic query. But I'd prefer `%dmg%` and `%dmgtype%` to just `%1` and `%2`
Also, I don't want to have a query on every roll for planar warrior, so it needs to just be a character sheet setting.

src/roll_renderer.pyj
```diff
@@ -425,7 +425,7 @@ class Beyond20RollRenderer:
                 is_critical = self.isCriticalHitD20(to_hit, critical_limit)
 
 
-            if request.damages?:
+            if request.damages?:
```
kakaroto, line 428: trailing spaces

src/roll_renderer.pyj
```diff
@@ -462,6 +462,16 @@ class Beyond20RollRenderer:
                 elif request.name == "Toll the Dead":
                     ttd_dice = v'await self.queryGeneric(request.name, "Is the target missing any of its hit points?", {"d12": "Yes", "d8": "No"}, "ttd_dice", ["d12", "d8"])'
                     damages[0] = damages[0].replace("d8", ttd_dice)
+                
+                # Ranger Ability Support
+                for dmgIndex, dmgType in enumerate(damage_types):
+                    if damage_types[dmgIndex] in ["Colossus Slayer", "Planar Warrior"]:
+                        dVal = damages[dmgIndex].toString()
+                        if dVal:
+                            dmg_dice = v'await self.queryGeneric(request.name, ("Add %1 damage?").replace(/%1/, dmgType), {"0": "No", [dVal]: "Yes"}, "dmg_dice", ["0", dVal])'
```
kakaroto, line 471: It makes more sense here to do `"Add " + dmgType + " damage"` than using the `%1` and a `replace`

Issue #201, opened 2020-04-16
make use of the character/item avatars


kakaroto:
Can use on roll20, a `[name](url)` macro to display an image.
We already have the character avatar and weapon avatar from the discord implementation, would need to show it. Probably via an option due to the size of the images sometimes...

![image](https://user-images.githubusercontent.com/27990/79505057-cfed9d00-8001-11ea-860a-8fb4869ecf85.png)

PR #202, opened 2020-04-17
Add functionality to display monster image in VTT

Brunhine:
partial support for #201

### DIFF
--- a/src/dndbeyond.pyj
+++ b/src/dndbeyond.pyj
@@ -537,7 +537,7 @@ class Monster(CharacterBase):
             if (avatarImg):
                 cb = def():
                     self.displayAvatar()
-                addRollButton(self, cb, avatarImg, small=True, image=True, text="Show in chat")
+                addRollButton(self, cb, avatarImg, small=True, image=True, text="Display in VTT")
         attributes = stat_block.find(base + "__attributes " + base + "__attribute")
         for attr in attributes:
             label = $(attr).find(base + "__attribute-label").text().trim()
--- a/src/dndbeyond.pyj
+++ b/src/dndbeyond.pyj
@@ -533,6 +533,11 @@ class Monster(CharacterBase):
         avatar = $(".details-aside .image a")
         if avatar.length > 0:
             self._avatar = avatar[0].href
+            avatarImg = $(".details-aside .image")
+            if (avatarImg):
+                cb = def():
+                    self.displayAvatar()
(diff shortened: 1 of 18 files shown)


src/dndbeyond.pyj
```diff
@@ -534,6 +533,11 @@ class Monster(CharacterBase):
         avatar = $(".details-aside .image a")
         if avatar.length > 0:
             self._avatar = avatar[0].href
+            avatarImg = $(".details-aside .image")
+            if (avatarImg):
```
kakaroto, line 537: using parenthesis here is not very python-y

src/dndbeyond.pyj
```diff
@@ -534,6 +533,11 @@ class Monster(CharacterBase):
         avatar = $(".details-aside .image a")
         if avatar.length > 0:
             self._avatar = avatar[0].href
+            avatarImg = $(".details-aside .image")
+            if (avatarImg):
+                cb = def():
+                    self.displayAvatar()
+                addRollButton(self, cb, avatarImg, small=True, image=True, text="Display in VTT")
```
kakaroto, line 540: for one liners like this, you can use rapydscript's lambda functions : 
```python
addRollButton(self, def() self.displayAvatar();, avatarImg, small=True, image=True, text="Display in VTT")
```

src/roll20.pyj
```diff
@@ -215,6 +215,10 @@ def template(request, name, properties):
 
     return result
 
+def rollAvatarDisplay(request):
+    return "[x](" + request.character.avatar + "#.png)"
+    #postChatMessage("[x](" + request.character.avatar + "#.png)", request.character.name)
```
kakaroto, line 220: oh, that's smart. You're overriding Roll20's image detection in case the URL doesn't have the right extension, but adding it as a query string.
I'd love to have seen that as a comment to clarify for future maintenance needs.

src/fvtt_script.pyj
```diff
@@ -140,6 +140,16 @@ def rollInitiative(request, custom_roll_dice=""):
             addInitiativeToCombat(roll)
     )
 
+def popAvatar(request):
+    # Check for the beyond20 module, if it's there, we can pop the image.
+    module = game.modules.get("beyond20")
```
kakaroto, line 145: Why beyond20 module? We don't need it. `ImagePopout` is a native FVTT API function.

src/roll20.pyj
kakaroto: Just noticed, the `postChatMessage` is commented out :p

PR #206, opened 2020-04-18
Fix for "custom dice aren't cleared from features page"

### LINKED ISSUE #183: custom dice aren't cleared from features page
ezgif-6-5980f6d4b17f

Brunhine:
fixes #183


kakaroto:
Sorry, while this fixes the specific issue, it will cause other issues because now you're not ignoring the 'beyond20-roll' classes. Which means, it will start causing the same behavior in other places that are not using custom dices (i.e: switching from one spell to another for example).
I'll fix it properly on my side.

PR #207, opened 2020-04-18
Clean Roll Strings

### LINKED ISSUE #191: Monk's Deflect Missiles rolling incorrectly
From Support questions on chrome store : https://chrome.google.com/webstore/detail/beyond-20/gnblbpbepfbfmoobegdogkglpbhcjofh

> I'm seeing an issue with how Deflect Missiles is rolled from the sidebar (the base ability, not the "Deflect Missiles Attack" you get when you reduce damage to zero). It should roll 1d10 and add your DEX modifier and monk level (so, for my 8th-level, 20 DEX monk, 1d10 + 13), but it is showing up in Roll20 as 1d10 + a DEX check (1d20+DEX) + 8 I poked about in the page inspector a bit, and the link has the attribute x-beyond20-roll="1d10 + +5 + 8". If I take off one of the duplicate "+" signs (i.e. "1d10 + 5 + 8" or "1d10 +5 + 8") the roll shows up correctly in Roll20. So I can manually correct it, but when I click off the sidebar and come back, it reverts to the original roll. I also tried "/r 1d10 + +5 + 8" directly in Roll20, and that parses correctly. But for some reason what gets sent across by Beyond20 is not right.

Brunhine:
Added support to clean roll strings for various imperfections due to edge case formatting in some abilities (Monk's deflect missiles)

fixes #191

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -2558,7 +2558,8 @@ function all(iterable) {
 if (!all.__argnames__) Object.defineProperties(all, {
     __argnames__ : {value: ["iterable"]}
 });
-var define_str_func, ρσ_unpack, ρσ_orig_split, ρσ_orig_replace;
+var decimal_sep, define_str_func, ρσ_unpack, ρσ_orig_split, ρσ_orig_replace;
+decimal_sep = 1.1.toLocaleString()[1];
 function ρσ_repr_js_builtin(x, as_array) {
     var ans, b, keys, key;
     ans = [];
@@ -2935,7 +2936,7 @@ define_str_func("format", function () {
                     value = value.toExponential(prec - 1);
                 }
                 value = value.replace(/0+$/g, "");
-                if (value[value.length-1] === ".") {
+                if (value[value.length-1] === decimal_sep) {
                     value = value.slice(0, -1);
                 }
                 if (ftype === "G") {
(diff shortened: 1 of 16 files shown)


src/utils.pyj
```diff
@@ -19,6 +19,14 @@ def replaceRolls(text, replaceCB):
         return replaceRollsCallback(m, replaceCB)
     , text)
 
+# Used to clean various imperfections in dice roll strings
+def cleanRoll(rollText):
+    #clean adjacent '+'s (Roll20 treats it as a d20)
+    #eg: (1d10 + + 2 + 3) -> (1d10 + 2 + 3)
+    rollText = rollText.replace(/\+ \+/g, '+')
```
kakaroto, line 26: Same should apply to '+ -' becoming '-'.

src/background.js
```diff
@@ -5542,6 +5543,14 @@ var str = ρσ_str, repr = ρσ_repr;;
             __argnames__ : {value: ["text", "replaceCB"]}
         });
 
+        function cleanRoll(rollText) {
+            rollText = rollText.replace(/\+ \+/g, "+");
```
kbuzsaki, line 5547: Heads up, I think this implementation breaks the "display in vtt" functionality because rollText will be `0` which then causes a type error.

Seems to be fixed by just guarding against the condition:

```
if rollText == 0:
  return 0
```

But up to y'all how you'd best like to fix it.

kakaroto: Yeah, it does, I've fixed it locally for now, by making it use a `.toString()` on the rollText and changed the `0` to `"0"`
Almost done converting this extension into pure javascript, then will commit the whole thing

kakaroto: FYI: the conversion is done and this fix was included with it

PR #256, opened 2020-05-09
Fix for Advantage/Disadvantage Dropped Rolls

### LINKED ISSUE #255: DDB Digital Dice Integration fails to determine dropped die w/ Disadvantage/Advantage
image

Due (I believe) to changes for DDB Dice, there is no longer anything to compare at this point in code (both results being 0).

This results in the right side of the equation ALWAYS receiving .setDiscarded(true)

Advantaged Attack
image

Disadvantaged Attack
image

Aeristoka:
Fixes #255

Digital Dice implementation changed when rolls take place for all rolls
 (Digital Dice or not) this fix determines dropped rolls
 After Rolls have actually taken place in the new paradigm

### DIFF
--- a/dist/dndbeyond_character.js
+++ b/dist/dndbeyond_character.js
@@ -1574,52 +1574,59 @@ class Beyond20RollRenderer {
         if ([RollType.DOUBLE, RollType.ADVANTAGE, RollType.DISADVANTAGE].includes(advantage)) {
             const roll_1 = this.createRoll("1d20" + modifier, data);
             const roll_2 = this.createRoll("1d20" + modifier, data);
-
-            if (advantage == RollType.ADVANTAGE) {
-                if (roll_1.total >= roll_2.total) {
-                    roll_2.setDiscarded(true);
-                } else {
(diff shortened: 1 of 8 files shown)


Aeristoka:
Tested and looks to be working Advantage/Disadvantage/Super Advantage/Super Disadvantage on DnDBeyond Integrated/Roll20/FVTT

kakaroto:
Good catch!
This won't always work though because the advantage could be QUERY in the request, and so the getToHit would need to return the advantage result that the query gave.
Also, I believe there's another place where getToHit is called (skill check for example, where it's not building an attack)

Aeristoka:
> Good catch!
> This won't always work though because the advantage could be QUERY in the request, and so the getToHit would need to return the advantage result that the query gave.
> Also, I believe there's another place where getToHit is called (skill check for example, where it's not building an attack)

I was worried there was a part of this I had missed. Where else does it need to be fixed up?

kakaroto:
haven't checked, still waking up, just search for 'gettohit' in the file to see where else it's used :p

PR #279, opened 2020-05-16
Alt Fix Tiebreaker Initiative

### LINKED ISSUE #277: Tiebreaker Initiative doesn't work on Dev Version
In src/dndbeyond/content-scripts/character.js the following:
image

returns 0 for all values except Dex (as it should), but the final return is Charisma, so tiebreaker is set to 0.

Aeristoka:
Fixes #277 - Tiebreaker Initiative doesn't work on Dev Version


Aeristoka:
I think this way is a lot more efficient, as it works through abilities, then breaks when it hits Dex, instead of continuing to loop over all abilities, returning each time.

PR #280, opened 2020-05-16
KaKaRoTo suggested Tiebreaker Initiative Fix

### LINKED ISSUE #277: Tiebreaker Initiative doesn't work on Dev Version

Aeristoka:
Fixes #277 - Tiebreaker Initiative doesn't work on Dev Version


kakaroto:
just realized the commit message 🤦‍♂️ it's ok (and too late) :p

PR #281, opened 2020-05-17
Implement Skill Badges on DnDBeyond

### LINKED ISSUE #229: Skill advantage granted by equipped item
When equipped the boots of Elevnkind grant "You also have advantage on Dexterity (Stealth) checks that rely on moving silently."

It would be nice if this was added to Stealth rolls (probably in addition to regular roll, since it's only for sound),it's just the thing to keep forgetting to do. :)

Not sure if this is in scope / technically possible, searched through tracker, but didn't find anything related to such items. :)

Aeristoka:
Fixes #229

### DIFF
--- a/dist/background.js
+++ b/dist/background.js
@@ -217,6 +217,7 @@ class RollType {
     static get SUPER_DISADVANTAGE() { return 7; }
     // If a feat overrides it to have advantage;
     static get OVERRIDE_ADVANTAGE() { return 8; }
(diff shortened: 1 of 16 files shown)


Aeristoka:
Could use an additional bit of code to show the proper Advantage/Disadvantage Icon on the Quick Roll and Side Panel. Not sure how to do that at the moment.

kakaroto:
well... that was much easier than I thought it would be to add!
I'll test/merge later when I get the time. Thanks!

PR #289, opened 2020-05-22
Beyond20 inside frames

Ainias:
added `"all_frames": true` to the content scripts inside the manifest. This allows beyond20 to work inside iframes (which I want to have, because I'm writing a DM-Screen Plugin with an optional iframe for the character sheet)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -58,7 +58,8 @@
 				"libs/alertify.min.js",
 				"libs/jquery-3.4.1.min.js",
 				"dist/dndbeyond_character.js"
-			]
+			],
+			"all_frames":true
 		},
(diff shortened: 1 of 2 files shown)


Aeristoka:
Looks like this also includes a CRLF -> LF conversion, if I'm looking at it right

kakaroto:
Looks like it. I'll fix the CRLF locally as a separate commit then rebase so I can review properly then merge.
Thanks!

Ainias:
Thanks! 
And sorry for the CRLF->LF conversion

kakaroto:
Fixed the CRLF, then merged your change. Thanks!

Issue #296, opened 2020-05-26
Missing package.json?


moritonal:
Hi @kakaroto, I was looking to provide a PR for the permissions issue #179  but I ran into the issue that there isn't a `package.json` in the main to describe the development process.

Rough gist is that the project might be improved by adding a "package.json" for main repo and adding commands for `npm run build-firefox` or `npm run build-chrome`, along with allowing `npm install` to simply install the dev-dependencies.

It might make more sense to compile all the files into a "./build/firefox" and "./dist/chrome" to avoid having to rename files and makes things more obvious.

I'll look into this happily, just wanted to raise it so I can get your feedback and thoughts.

kakaroto:
Hi Tom,
Thanks for the help!
I don't really see the point of having a package.json, but it might just be because I'm used to doing things differently, and this isn't a node application, so I'm not sure how why it would be necessary. The build process has changed recently and I haven't gone through the release process with the new build system yet.
The README has instructions though on how to build it. You need to have `gulp` installed and just run `gulp` to build the dist files.

The build is done with 'make build' which runs the webext program (which I then modify manually the generated zip to remove the extraneous stuff from the docs folder and the src/ dir), and the only difference between firefox and chrome is the `manifest_ff.json` is renamed to `manifest.json`.

I suppose one run command that the package.json could have is for launching the standalone docs server, I have a run.sh file in the docs folder with this in it :
`~/bin/bundle exec jekyll serve --incremental --host 0.0.0.0`
I don't even remember what `bundle` is or where it came from, but it works and I don't touch it once it works 🤷 
If you can improve this process, that'd be great!

moritonal:
So that makes sense, but can I suggest a more developed build pipeline? I see now there's a Make file, but honestly I didn't even notice that when I first started looked at this project, and have no idea how I'd run it on a windows box.

You already have gulp (which means we can assume node is installed) which means we could use that build a much clearer process?

If you look at the PR (#300) I submitted you'll see I've used a `package.json` to wrap up all the dependencies, build-process and other meta things, which make the process of building and packaging a version for Chrome and Firefox as simple as:

`npm install`
`npm run build`
`npm run package`

Along with things like `npm run start:chrome` for development. It's in no ways perfect, but I've used it many times before on projects and thought you might be interested in seeing it.

kakaroto:
Hey, since the merge of #300, this is fixed, thanks!

PR #300, opened 2020-05-27
Added Package.json and other scripting to help issue #296

moritonal:
Okay, so this is bit a of a big change for a newbie to the repo so I'm mostly just putting this forward so you can see what I meant in issue #296. I'd recommend pulling this locally and playing around with `npm run` to see how it compares.

By using `package.json` for building and packaging related tasks, it clearly shows how a developer can work with the project. Given we assume gulp is available, it's easier just to use node. Dev-dependencies are added to the PATH when running `npm run` commands so we can use that to keep the environment cleaner.

`npm-run-all` lets us create a clean scripting environment, and we can use `gulp` to build induvidual directories for each browser so there is a clear "here are all the files for x" setup.

I'll understand of course if you're more comfortable with the current setup.


kakaroto:
Thanks for this. I'll have a look later, this is going to get bumped until after 2.0 release. From a quick look, I'd say my main issues at the moment are : 
- I don't like the 'compiled' directory, I like it being 'dist', but I understand the need to change. 'build' might be nicer though.
- While I don't like having built binaries in a repository, I also want to make it easy for people to try the dev version without having to mess with compiling. So either having to keep the built files in dist or having a system setup for auto builds (with github actions) and provided somewhere might be something I'd need to do before implementing this change.

kakaroto:
Hey @moritonal,
Thanks for the PR. I've copied your changes manually to the repo, and fixed some stuff then committed it under your name/authorship.
Main changes are : 
- keep the dist directory and instead create a build directory for the built chrome/firefox builds
- Moved the requires to the top of gulpfile and use const instead of var
- The `web-ext build` was wrong as it would use current directory for source, so I had to add `--source-dir=` argument instead of the `--artifacts-dir=` argument you used
- Added missing package.json, gulpfile.js and LICENSE.MIT files to the base build which would be needed to build from source
- Renamed the 'firefox-build' and 'chrome-build' tasks into 'build-firefox' and 'build-chrome'
- Made the 'package' npm script depend on the 'build' script first
- in package.json, changed the homepage to the point to the actual https://beyond20.here-for-more.info website

Realizing now, that the package.json version is wrong, and I need to check the license field, if "GNU" is a proper license name for npm, or if it should be 'GPLv3' instead (with some parts being licensed under MIT)

Thanks again for doing the heavy lifting in making this a much more streamlined process.

PR #316, opened 2020-05-30
Cleaner Psychic Blade Sneak Attack Fix

### LINKED ISSUE #311: Psychic Blades does not roll Sneak Attack Damage
https://www.dndbeyond.com/classes/rogue#SoulknifeUA

"You can manifest your psionic power as shimmering blades of psychic energy. When you are about to make a melee or ranged weapon attack against a creature, you can manifest a psychic blade from your free hand and make the attack with that blade. This magic blade is a simple melee weapon with the finesse and thrown properties."

Aeristoka:
Fixes #311

Rogue Soulknife (UA): Psychic Blades
3rd-level Soulknife feature

You can manifest your psionic power as shimmering blades of psychic energy. When you are about to make a melee or ranged weapon attack against a creature, you can manifest a psychic blade from your free hand and make the attack with that blade. This magic blade is a simple melee weapon with the finesse and thrown properties.

Sneak Attack
Beginning at 1st level, you know how to strike subtly and exploit a foe’s distraction. Once per turn, you can deal an extra 1d6 damage to one creature you hit with an attack if you have advantage on the attack roll. The attack must use a finesse or a ranged weapon

Sneak Attack should therefore apply to Psychic Blades (handled through rollAction).

### DIFF
--- a/dist/dndbeyond_character.js
+++ b/dist/dndbeyond_character.js
@@ -4503,7 +4503,8 @@ function rollAction(paneClass) {
             character.getSetting("warlock-hexblade-curse", false))
             critical_limit = 19;
         // Polearm master bonus attack using the other end of the polearm is considered a melee attack.
-        if (action_name == "Polearm Master - Bonus Attack" || action_name.includes("Unarmed Strike") || action_name == "Tavern Brawler Strike") {
+        if (action_name == "Polearm Master - Bonus Attack" || action_name.includes("Unarmed Strike") || action_name == "Tavern Brawler Strike"
+            || action_name.includes("Psychic Blade")) {
             if (character.hasClassFeature("Fighting Style: Great Weapon Fighting"))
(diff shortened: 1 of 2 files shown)


kakaroto:
I'd have cleaned the indentation first, because now the psychic blade change is non working on that specific commit until the next commit comes into place.
Also, why the "Reach" check for sneak attack. You already check for psychic blades, so it shouldn't be needed.
Looks good otherwise

Aeristoka:
> I'd have cleaned the indentation first, because now the psychic blade change is non working on that specific commit until the next commit comes into place.
> Also, why the "Reach" check for sneak attack. You already check for psychic blades, so it shouldn't be needed.
> Looks good otherwise

Reach was a forgotten holdover from the other implementation. I'll pull that out, but do you want both commits un-done and reversed?

kakaroto:
Well, the first "indent fix" commit also adds support for psychic blades... but yeah, sure, I'll let that slide :p

PR #328, opened 2020-06-02
Implement Silver Tongue

### LINKED ISSUE #327: Silver Tongue: Bard College of Eloquence
From the new Mythic Odysseys of Theros
https://www.dndbeyond.com/classes/bard#CollegeofEloquence

Silver Tongue
3rd-level College of Eloquence feature

You are a master at saying the right thing at the right time. When you make a Charisma (Persuasion) or Charisma (Deception) check, you can treat a d20 roll of 9 or lower as a 10.

Similar to Rogue Reliable Talent, but specific to ONLY two Skills

Aeristoka:
From Mythic Odysseys of Theros
Bard: College of Eloquence
https://www.dndbeyond.com/classes/bard#CollegeofEloquence

Silver Tongue
3rd-level College of Eloquence feature

You are a master at saying the right thing at the right time. When you make a Charisma (Persuasion) or Charisma (Deception) check, you can treat a d20 roll of 9 or lower as a 10.

Similar to Rogue Reliable Talent, but specific to ONLY two Skills

Fixes #327

### DIFF
--- a/dist/dndbeyond_character.js
+++ b/dist/dndbeyond_character.js
@@ -2032,7 +2032,9 @@ class Beyond20RollRenderer {
             }
         } else {
             const data = { [request.ability]: request.modifier, "custom_dice": custom_roll_dice }
-            const d20_modifier = request.reliableTalent ? "min10" : "";
+            let d20_modifier = request.reliableTalent ? "min10" : "";
+            if (request.silverTonge && (request.skill === "Deception" || request.skill === "Persuasion"))
+                d20_modifier = "min10";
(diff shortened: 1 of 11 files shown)


kakaroto:
Starting to dislike the way I did it (moving the handling of these cases in the renderer/VTT side), but I don't think there's a better way of doing it... unless! How about instead of setting 'reliableTalent' and 'silverTongue' we instead set 'd20Modifier' in the request from the DNDB side and then the renderer can just append it to the d20 and it would always be whatever it needs to be.
Yep, I'll want to change the design that way! I'll merge this as is for now, and file an issue for the refactor of that portion.

PR #337, opened 2020-06-02
Implement Path of the Zealot Divine Fury

### LINKED ISSUE #336: Barbarian Divine Fury
Path of the Zealot Divine Fury is not considered

https://www.dndbeyond.com/classes/barbarian#PathoftheZealot

Divine Fury
Starting when you choose this path at 3rd level, you can channel divine fury into your weapon strikes. While you’re raging, the first creature you hit on each of your turns with a weapon attack takes extra damage equal to 1d6 + half your barbarian level. The extra damage is necrotic or radiant; you choose the type of damage when you gain this feature.

Aeristoka:
Fixes #336

https://www.dndbeyond.com/classes/barbarian#PathoftheZealot

Divine Fury
Starting when you choose this path at 3rd level, you can channel divine fury into your weapon strikes. While you’re raging, the first creature you hit on each of your turns with a weapon attack takes extra damage equal to 1d6 + half your barbarian level. The extra damage is necrotic or radiant; you choose the type of damage when you gain this feature.

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -226,7 +226,7 @@ function rollItem(force_display = false) {
         if (character.hasClassFeature("Rage") && character.getSetting("barbarian-rage", false) && character.hasClassFeature("Divine Fury") &&
             properties["Attack Type"] == "Melee") {
             const barbarian_level = character.getClassLevel("Barbarian");
-            damages.push(String(`1d6+${Math.floor(barbarian_level / 2)}`));
+            damages.push(`1d6+${Math.floor(barbarian_level / 2)}`);
             damage_types.push("Divine Fury");
         }
         if (character.getSetting("sharpshooter", false) &&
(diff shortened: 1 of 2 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -223,6 +223,12 @@ function rollItem(force_display = false) {
             damages.push(String(rage_damage));
             damage_types.push("Rage");
         }
+        if (character.hasClassFeature("Rage") && character.getSetting("barbarian-rage", false) && character.hasClassFeature("Divine Fury") &&
+            properties["Attack Type"] == "Melee") {
+            const barbarian_level = character.getClassLevel("Barbarian");
+            damages.push(String(`1d6+${Math.floor(barbarian_level / 2)}`));
```
kakaroto, line 229: Using ``` `${}` ``` already makes it into a string, so you don't need to add a call to `String()`

kakaroto:
Please do the change (in both areas) and will merge it. Looks good. I wish there was a wish to know if it's radiant or necrotic damage that the user chose.

src/dndbeyond/content-scripts/character.js
Aeristoka: I think that was a holdover from how I was doing it at first that I didn't clean up... Standby...

kakaroto:
Thanks for the quick fix

PR #351, opened 2020-06-03
Fix for Chaos Bolt using new resolveRolls

### LINKED ISSUE #344: Chaos bolt
chaos bolt has a box asking for a damage type before the roll My assumption is it's picking up the same thing chromatic Orb does but that conflicts with the special case stuff you've had to do for Chaos bolt in the past because the drop down is empty

image

Aeristoka:
Moves the processing for Chaos Bolt after resolveRolls
Allows Chaos Bolt to process the roll results properly

Fixes #344

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -595,6 +595,9 @@ class Beyond20RollRenderer {
                 }
             }
 
+            await this._roller.resolveRolls(request.name, all_rolls)
+            
+            //Moved after the new resolveRolls so it can access the roll results
             if (request.name == "Chaos Bolt") {
                 for (let [i, dmg_roll] of damage_rolls.entries()) {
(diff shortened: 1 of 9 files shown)


src/common/roll_renderer.js
```diff
@@ -595,6 +595,9 @@ class Beyond20RollRenderer {
                 }
             }
 
+            await this._roller.resolveRolls(request.name, all_rolls)
+            
+            //Moved after the new resolveRolls so it can access the roll results
```
kakaroto, line 600: This comment is not needed, because it explains what this specific commit does, not the code. So reading the code back in a year, we see "moved after... " and we wonder "why is this saying moved?" 
A comment like "Handle Chaos Bolt after the rolls are resolved" would be nicer. Also, space after `//`  :p

PR #358, opened 2020-06-03
Change to "Dragon's Breath" Spell

### LINKED ISSUE #356: Dragon's Breath Doesn't Ask for Damage Type
I've only just started checking out Beyond20 so apologies if I've managed to miss a report about this. Also I wanted to say a huge thank you for working on this, and I'm impressed at the speed with which integration with D&D Beyond's dice roller has been added!

The issue I'd like to report is with the spell Dragon's Breath from Xanathar's Guide to Everything; this basically gives a character a Dragonborn breath weapon attack, with a choice of five damage types (acid, cold, fire, lightning or poison).

When casting this with Beyond20 though no choice is given, it instead rolls a very generous 15d6; thankfully it recognises the different damage types and breaks them down in its results on Roll20, so we can still get the correct value. My understanding is that similar spells (such as Chromatic Orb) ask for the correct damage type before rolling, so if it's possible to do the same that would be ideal I think?

Thanks again!

Edit: This also seems to be affecting Hunter's Mark (rolls damage for bludgeoning, piercing and slashing all at once).

Aeristoka:
Follows Chromatic Orb, in that it asks for which damage type you choose for the Spell Damage

Fixes #356

### DIFF
--- a/dist/dndbeyond_character.js
+++ b/dist/dndbeyond_character.js
@@ -2163,6 +2163,16 @@ class Beyond20RollRenderer {
                     critical_damages.splice(0, 0, crit_damage);
                     critical_damage_types.splice(0, 0, chromatic_type);
                 }
+            } else if (request.name === "Dragon's Breath") {
+                const damage_choices = {}
+                for (let dmgtype of ["Acid", "Cold", "Fire", "Lightning", "Poison"]) {
+                    let idx = damage_types.findIndex(t => t === dmgtype);
+                    damage_choices[damage_types.splice(idx, 1)[0]] = damages.splice(idx, 1)[0];
+                }
+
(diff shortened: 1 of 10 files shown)


Haravikk:
Thanks!

Does this also address the same issue with Hunter's Mark? I added that as a note to the end of #356, but perhaps I should have posted a separate issue for it?

Aeristoka:
> Thanks!
> 
> Does this also address the same issue with Hunter's Mark? I added that as a note to the end of #356 shortly after I posted it, but perhaps I should have posted a separate issue for it?

Hunter's Mark is a little different, as this kind of change would require you knowing the weapon damage you already rolled, and inputting that properly. I'll look at it.

PR #359, opened 2020-06-03
Bugfix for Brutal Critical/Savage Attacks

### LINKED ISSUE #357: Brutal Critical doesn't work with Perfect Rolls Crit option
I was playing around with a level 20 Barbarian character, and I noticed that when I rolled a critical hit, it did not roll any extra damage for my Brutal Critical class feature.  I double checked to see if the "Brutal Crit/Savage Attacker" feature was toggled on, and it was, so I went to discord to check for help, and while I was waiting for a reply, I played around with a few other settings, and noticed that if I turned Critical Hit rule to "Standard PHB Rule" it suddenly started working correctly (I had "Homebrew: Perfect Rolls" turned on before that because that is what rule our gaming group uses.)

Here is a screen shot of what showed up when it didn't work:

https://i.imgur.com/AlfAvOt.png?1 (Yes, I rolled 2 20s on that roll :D )

Aeristoka:
Fixes Homebrew "Perfect Dice" for these options

Fixes #357

### DIFF
--- a/dist/dndbeyond_character.js
+++ b/dist/dndbeyond_character.js
@@ -2899,21 +2899,37 @@ function buildAttackRoll(character, attack_source, name, description, properties
                 }
             }
             if (brutal > 0) {
+                const rule = parseInt(character.getGlobalSetting("critical-homebrew", CriticalRules.PHB));
                 let highest_dice = 0;
-                for (let dmg of crit_damages) {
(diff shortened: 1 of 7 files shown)


kakaroto:
Not a fan of the way it's done, but nothing wrong with it, I just would have done it differently. Looks good otherwise!

PR #375, opened 2020-06-05
Rectify missed icon path fix from new Icon change

Aeristoka:
Missed in recent FVTT 1.1 update


kakaroto:
No, this is correct. The message has icon20.png when it gets posted (for the other players) but the extension replaces it with the path to the extension's own url on render. This allows it to display properly if the module is not installed on the foundry instance.

Issue #204, opened 2020-04-18
Display custom modifiers on the modifier field


spisin:
I hope rolls with modifiers show as this figure.
![roll](https://user-images.githubusercontent.com/63902131/79633363-f6e1d580-819f-11ea-9089-eff83b4ee3d5.jpg)

It displays custom modifiers on the modifier field. Then we don't have to mouseover on the results to check modifiers from now.

I have roll20.pyj for this. May I send my code?

Aeristoka:
Send the PR, he'll look it over to see if he wants it.

spisin:
I don't have a permission to push.

Aeristoka:
You should be able to open a Pull Request

kakaroto:
Thanks! That sounds like a useful change. You would need to send a Pull Request for it.
See github's normal workflow for contributing changes, it's explained here : 
https://gist.github.com/Chaser324/ce0505fbed06b947d962
If you **really** have issues doing it, you could send a patch (`git format-patch origin/master` I believe)

spisin:
I could not select my branch for a Pull Request.

The patch is here.
[0001-Display-custom-modifiers-on-modifier-field-of-roll20.patch.tar.gz](https://github.com/kakaroto/Beyond20/files/4498080/0001-Display-custom-modifiers-on-modifier-field-of-roll20.patch.tar.gz)

PR #381, opened 2020-06-07
Bugfix for Rolling Hit Die (including custom die)

### LINKED ISSUE #378: Hit die not rolling in Roll20 5e Sheet
The D&D Beyond Hit Die (under Short Rest) isn't showing the roll in Roll20 when it's set to the 5e sheet. It works if it's set to the Beyond 20 Roll Renderer though.

Aeristoka:
Fixes #378

### DIFF
--- a/dist/dndbeyond_character.js
+++ b/dist/dndbeyond_character.js
@@ -2054,9 +2054,12 @@ class Beyond20RollRenderer {
         return this.rollD20(request, "Initiative" + "(" + request.initiative + ")", data);
     }
 
-    rollHitDice(request) {
+    rollHitDice(request, custom_roll_dice = "") {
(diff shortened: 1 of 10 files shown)


kakaroto:
That's wrong. The bug was me pasting the "format_plus_mod(custom_roll_dice)" where I shouldn't and the line : `        "mod": request["hit-dice"] + format_plus_mod(custom_roll_dice),` should not have it. Removing that will fix the bug.
The custom roll dice applies to d20 rolls (Bless or other d20 custom modifiers), a hit dice is a d8, d10, d12, it's not a d20, so it should not have the custom roll dice applied to it. That's why it never had the custom die apply to hit dice, but I broke it when I applied #204 manually.

PR #382, opened 2020-06-07
Implement Magical Modifiers for Death Saves

### LINKED ISSUE #163: death saves should include magical modifiers on saving throws
Cloak of protection or other magics should affect death saves

https://rpg.stackexchange.com/questions/94927/does-a-paladins-aura-of-protection-affect-allies-death-saving-throws

Aeristoka:
Covers things like:
"Cloak of Protection"
"Ring of Protection"
Can also be used with Custom Die

Fixes #163

### DIFF
--- a/dist/dndbeyond_character.js
+++ b/dist/dndbeyond_character.js
@@ -2060,7 +2060,8 @@ class Beyond20RollRenderer {
     }
 
     rollDeathSave(request, custom_roll_dice = "") {
-        return this.rollD20(request, "Death Saving Throw", { "custom_dice": custom_roll_dice });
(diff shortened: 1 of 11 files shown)


kakaroto:
It looks for all positive bonus anywhere, so if you have one somewhere other than what would affect death saves, then it would cause an issue too. Also, it doesn't consider the fact that someone might be on tablet or mobile mode and the saving throws area isn't displayed on the page. That information needs to be cached first for those situations.
There are also modifiers that are situational, like for example (in my recent game) a tortle has adv/disadv to saving throws "if within its shell".
So there's more to fixing that than what this patch does.

PR #386, opened 2020-06-08
Implement Artificer: Battlesmith Arcane Jolt

Aeristoka:
Applies to Magical Weapon Attacks
Based on weapon not containing ", Common" in item_type
May be a better way to account for "Magical Weapon" as this also does not cover infusions.

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -587,6 +587,12 @@ const character_settings = {
         "description": "Your pact weapon drips with necrotic energy, lending extra damage to your strikes",
         "type": "bool",
         "default": false
+    },
+    "artificer-arcane-jolt": {
+        "title": "Artificer: Use Arcane Jolt",
+        "description": "Apply an Arcane Jolt to you or your Steel Defender's Weapon Attacks. Deals extra 2d6 damage, or 4d6 at Artificer Level 15+",
+        "type": "bool",
+        "default": false
     }
 }
 
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -254,13 +254,6 @@ const options_list = {
         "choices": { "all": "All components", "material": "Only material components", "none": "Do not display anything" }
     },
 
-    "roll20-spell-description-display": {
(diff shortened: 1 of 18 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -369,6 +369,17 @@ function rollItem(force_display = false) {
                     brutal += 1;
             }
         }
+
+        //Artificer Battlemaster Arcane Jolt
+        // TODO: Implement for Steel Defender at later date
+        if (damages.length > 0 &&
+            character.hasClassFeature("Arcane Jolt") &&
+            character.getSetting("artificer-arcane-jolt", false) &&
+            item_type.indexOf(", Common") === -1) {
```
kakaroto, line 378: 😬 I'm unsure about the ", Common" thing (I think it already broke since now it's ", common" lowercase.
How about checking for "magic" in the description? I noticed most/all magic weapons will say "this magic <thing>" in its description.

Aeristoka: Yeah, I'll have a look again, I just remember coding this and going "how the crap are you supposed to figure out what to apply this to..."

Aeristoka: Honestly, I think checking for anything but it being not common isn't going to work. I found discrepancies between Named weapons and +X weapons, as well as just named weapons with no +X to hit and attack.

Testing for not-common means anything magical will get the application, as they're magical.

src/common/settings.js
```diff
@@ -239,6 +239,13 @@ const options_list = {
         "choices": { "all": "All components", "material": "Only material components", "none": "Do not display anything" }
     },
 
+    "roll20-spell-description-display": {
+        "title": "Display Spell Descriptions in spell attacks",
+        "description": "When doing a spell attack, display the spells full description (Roll20 only toggle)",
+        "type": "bool",
+        "default": false
+    },
```
kakaroto, line 247: this shouldn't be in here.

PR #389, opened 2020-06-09
Implement Champion Fighter Remarkable Athlete

Aeristoka:
Does not stack with Bard: Jack of All Trades
https://www.sageadvice.eu/2017/01/27/is-it-safe-to-assume-that-remarkable-athlete-and-jack-of-all-trades-cannot-be-applied-to-the-same-ability-check/

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -43,9 +43,20 @@ function rollAbilityOrSavingThrow(paneClass, rollType) {
     const ability_name = ability_string.split(" ")[0];
     const ability = ability_abbreviations[ability_name];
     let modifier = $("." + paneClass + "__modifier .ct-signed-number,." + paneClass + "__modifier .ddbc-signed-number").text();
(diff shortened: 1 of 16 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -43,9 +43,20 @@ function rollAbilityOrSavingThrow(paneClass, rollType) {
...
+    let remarkable_athlete_used = false;
+
+    if (rollType == "ability" &&
+        character.hasClassFeature("Remarkable Athlete") &&
+        character.getSetting("champion-remarkable-athlete", false) &&
+        (ability == "STR" || ability == "DEX" || ability == "CON")) {
+        const remarkable_athlete_mod = Math.ceil(character._proficiency / 2);
+        modifier += " + " + remarkable_athlete_mod;
+        remarkable_athlete_used = true;
+    }
```
kakaroto, line 55: I would just put the JoaT in the `else` with a comment that they don't stack, instead of having a `remarkable_athlete_used` flag.

PR #390, opened 2020-06-09
Implement Oath of Conquest: Invincible Conquerer

Aeristoka:
https://www.dndbeyond.com/classes/paladin#OathofConquest

Invincible Conqueror
At 20th level, you gain the ability to harness extraordinary martial prowess. As an action, you can magically become an avatar of conquest, gaining the following benefits for 1 minute:

You have resistance to all damage.
When you take the Attack action on your turn, you can make one additional attack as part of that action.
Your melee weapon attacks score a critical hit on a roll of 19 or 20 on the d20.

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -451,7 +451,14 @@ const character_settings = {
         "description": "Unleash your divine soul to deal extra radiant damage equal to your level.",
         "type": "bool",
         "default": false
+    },
+    "paladin-invincible-conquerer": {
+        "title": "Oath of Conquest: Invincible Conquerer",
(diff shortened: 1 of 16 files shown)


src/common/settings.js
```diff
@@ -451,7 +451,14 @@ const character_settings = {
         "description": "Unleash your divine soul to deal extra radiant damage equal to your level.",
         "type": "bool",
         "default": false
+    },
+    "paladin-invincible-conquerer": {
+        "title": "Oath of Conquest: Invincible Conquerer",
```
kakaroto, line 456: "Conqueror" instead of "Conquerer"

Aeristoka:
Re-done in a different PR

PR #392, opened 2020-06-11
Bugfix for custom_damage

Aeristoka:
Improves parsing to split on ":" instead of ": "
Bug found & reported by n00ch on Discord


kakaroto:
Technically, you found the bug, n00ch triggered it 🙄

PR #400, opened 2020-06-13
Bugfix for Skill Badges and "Always Ask" rolls

Aeristoka:
I don't think I saw any cleaner way to get the setting


kakaroto:
Yep, that looks good to me

PR #401, opened 2020-06-13
Bugfix for Rolling Hit Die (including custom die)

### LINKED ISSUE #378: Hit die not rolling in Roll20 5e Sheet

Aeristoka:
Proper fix

Fixes #378


kakaroto:
Too complicated to review, I'll just merge it as is

PR #405, opened 2020-06-14
Roll20 token hp

shadow7412:
Elaborated on the process to link the D&D Beyond HP (and temp hp) to a Roll20 character sheet, and the relevant token.


kakaroto:
Awesome, thanks!

Issue #406, opened 2020-06-14
Biased die rolls


taekahn:
The way the rolls are calculated is biased and will not produce equally likely results across the entire range.    
Suggestion:
Replace 
```
let die = Math.floor(Math.random() * this.faces) + 1
```
With 
```
let Random = require("random-js");
let random_engine = Random.engines.mt19937().autoSeed();
...
let inclusive_range = Random.integer(1, this.faces);
let die = inclusive_range(random_engine);
```
To make the rolls fair.

kakaroto:
Hi,
This was explored already and you are wrong.
First, this is using pure JS, I won't be using any sort of `require` statement and switching to browserify or whatever is needed to generate the files. There are also much simpler and self container Mersenne Twister implementation, like the one included in Foundry VTT's code.

Second, this may have been true over 5 years ago, but it's not true anymore.

Here's part of the discussion that happened about this a few months ago on Discord : 
> Mersenne Twister is worse than almost any random generator made after 2015. Math.Random doesn't actually generator numbers, it tells the browser to, and most modern browsers use Xorshift128+ which is on the very high end of Pseudo Random Number Generators.
Point being, if you're using a browser made in 2016 or earlier, you're probably fine.
Unless some hipster made their browser use some really silly code,  Mersenne Twister is actually worse than the core

You can read about the algorithm used by v8 (Chrome and all Chromium based browsers) since 2015 which is xorshift128+ : https://v8.dev/blog/math-random
And you can see that Mozilla also switched to it 5 years ago : https://bugzilla.mozilla.org/show_bug.cgi?id=322529
If someone is using a browser from 2014 and hasn't updated it since, then the fact that the dice rolls may be biased will be the least of their problem.

taekahn:
Mersene twister was just an example. I was not raising it up as an ultimate random number generator.  
The fact that that is what you're pointing to is betraying your ignorance.  
It has nothing to do with the method with which a random number is generated and everything to do with how it is selected.  

**The fact is** using multiplication for fitting a randomly generated number to a range introduces bias. It is unequivocally an incorrect way to select a random number within a range.  
You can find out plenty about it by googling, but here is a site that explains it.  
https://pthree.org/2018/06/13/why-the-multiply-and-floor-rng-method-is-biased/  
This will be true forever.  

Your rolls _are_ biased. If you're ok with that, then you're ok with that.

kakaroto:
[...]

PR #407, opened 2020-06-15
Refactored dice rolling logic into function.

shadow7412:
…that should be slightly less biased.

Decided to have a go at fixing this up despite the concerted effort from the raiser of #406 to get us to ignore it forever.

This version splits 1 into equal chunks according to the dice being rolled, which probably is better?

### DIFF
--- a/src/common/dice.js
+++ b/src/common/dice.js
@@ -133,18 +133,24 @@ class DNDBDice {
         }
         this._rolls = [];
     }
-
+    rollSingleDie () {
+        // Borrowed from https://pthree.org/2018/06/13/why-the-multiply-and-floor-rng-method-is-biased/
+        const max = Math.floor(2 ** 32 / this.faces) * this.faces; // make "max" a multiple of "faces"
+        let x;
+        do {
+            x = Math.floor(Math.random() * 2 ** 32); // pick a number of [0, 2^32).
+        } while (x >= max); // try again if x is too big
+        return (x % this.faces) + 1; // uniformly picked in [1, faces] (inclusively)
+    }
     async rollDice() {
         this._rolls = [];
         for (let i = 0; i < this.amount; i++) {
-            let die = Math.floor(Math.random() * this.faces) + 1;
-            this._rolls.push({ "roll": die });
+            this._rolls.push({ "roll": this.rollSingleDie() });
         }
     }
     async rerollDice(amount) {
         for (let i = 0; i < amount; i++) {
-            let die = Math.floor(Math.random() * this.faces) + 1;
-            this._rolls.push({ "roll": die });
+            this._rolls.push({ "roll": this.rollSingleDie() });
         }
     }
     async roll() {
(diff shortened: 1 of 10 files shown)


src/common/utils.js
```diff
@@ -12,6 +12,10 @@ function replaceRollsCallback(match, replaceCB) {
     return result;
 }
 
+function getDieRollResult(faces) {
+    return Math.floor(Math.random() / (1 / faces)) + 1
```
kakaroto, line 16: That's in no way better, you're changing `x * y`  into `x / (1 / y)` which is the exact same mathematical equation.
The trick with the division in the examples would only work for integer random values between 0 and 2**32, not for floating points between 0 and 1.0.
See the suggested link which makes it unbiased by making the range into an integer and discarding values that fall in that small interval between where the modulo wouldn't be uniform anymore. https://pthree.org/2018/06/13/why-the-multiply-and-floor-rng-method-is-biased/

Also, I'd prefer if this function was a method inside the `DNDBDice` class, instead of being a function in the utils file.

shadow7412: Yeah, I figured I probably should have googled it rather than just make something up.

When I first wrote it, I noticed it was being used everywhere so I wanted to keep it in a common place. After updating it to use the function I then realised that most of those duplicates were in the dist folder, and weren't _really_ duplicates.

I'll have a look at that link.

kakaroto:
I see a lot of changes in dist files that shouldn't be there, there's changes to utils.js which are only changing whitespaces, the `getDieResult` is in a dist file, but not appearing in any of the src files. And there are some other changes (a big block) in dist files that don't appear in src either.
Please clean it all up, make sure you make your changes in the right file and have gulp running continuously so it keeps the dist files properly generated.
I'd also suggest renaming the `getDieResult` into `rollSingleDie` perhaps because it makes the roll, it doesn't just "get" a stored value.
Also, please squash all commits into one in this case since it's all the same change.
Thanks!

src/common/dice.js
```diff
@@ -133,18 +133,24 @@ class DNDBDice {
...
     }
-
+    getDieResult () {
+        // Borrowed from https://pthree.org/2018/06/13/why-the-multiply-and-floor-rng-method-is-biased/
+        const max = Math.floor(2 ** 32 / this.faces) * this.faces; // make "max" a multiple of "range"
+        let x;
+        do {
+            x = Math.floor(Math.random() * 2 ** 32); // pick a number of [0, 2^32).
+        } while (x >= max); // try again if x is too big
+        return x % this.faces; // uniformly picked in [0, range)
```
kakaroto, line 143: "uniformly picked in [0, range)"
What we want is the range `[1, range]` instead. As it is, rolls are going to become 0-19 instead of 1-20 on a d20.
Also change the comments to say 'faces' instead of 'range' (both here and earlier in this function).

src/common/dice.js
```diff
@@ -133,18 +133,24 @@ class DNDBDice {
         }
         this._rolls = [];
     }
-
+    getDieResult () {
```
kakaroto, line 136: as suggested in the conversation, rename the method into `rollSingleDie` instead.

kakaroto:
Nice, thanks! Looks good to me :)

PR #413, opened 2020-06-19
Attach Roll to ChatMessage as generated for FVTT

Aeristoka:
Roll is a roll that contains a DicePool of all damage and attack rolls

Change non-whispered Beyond20 roll to Roll type message

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -326,7 +326,7 @@ class Beyond20RollRenderer {
         if (request.sendMessage && this._displayer.sendMessage)
             this._displayer.sendMessage(request, title, html, buttons, character, request.whisper, play_sound, source, attributes, description, attack_rolls, roll_info, damage_rolls, total_damages, open)
         else
-            this._displayer.postHTML(request, title, html, buttons, character, request.whisper, play_sound);
+            this._displayer.postHTML(request, title, html, buttons, character, request.whisper, play_sound, attack_rolls, damage_rolls);
 
         if (attack_rolls.length > 0) {
             return attack_rolls.find((r) => !r.isDiscarded());
--- a/src/fvtt/page-script.js
+++ b/src/fvtt/page-script.js
@@ -7,7 +7,7 @@ var settings = null;
 var extension_url = "/modules/beyond20/";
 
 class FVTTDisplayer {
-    postHTML(request, title, html, buttons, character, whisper, play_sound) {
+    postHTML(request, title, html, buttons, character, whisper, play_sound, attack_rolls, damage_rolls) {
         Hooks.once('renderChatMessage', (chat_message, html, data) => {
             const icon = extension_url + "images/icons/badges/custom20.png";
             html.find(".ct-beyond20-custom-icon").attr('src', icon);
@@ -20,10 +20,10 @@ class FVTTDisplayer {
                 buttons[button]();
             });
         });
-        return this._postChatMessage(html, character, whisper, play_sound);
+        return this._postChatMessage(html, character, whisper, play_sound, attack_rolls, damage_rolls);
     }
 
-    _postChatMessage(message, character, whisper, play_sound = false) {
+    _postChatMessage(message, character, whisper, play_sound = false, attack_rolls, damage_rolls) {
         const MESSAGE_TYPES = CONST.CHAT_MESSAGE_TYPES || CHAT_MESSAGE_TYPES;
         const data = {
             "content": message,
@@ -32,8 +32,8 @@ class FVTTDisplayer {
         }
         const rollMode = this._whisperToRollMode(whisper);
         if (["gmroll", "blindroll"].includes(rollMode)) {
-            data['type'] = MESSAGE_TYPES.WHISPER;
             data["whisper"] = ChatMessage.getWhisperIDs("GM");
+            data['type'] = MESSAGE_TYPES.WHISPER;
             if (rollMode == "blindroll")
                 data["blind"] = true;
         } else {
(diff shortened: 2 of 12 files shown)


src/fvtt/page-script.js
```diff
@@ -37,7 +46,7 @@ class FVTTDisplayer {
             if (rollMode == "blindroll")
                 data["blind"] = true;
         } else {
-            data['type'] = MESSAGE_TYPES.OOC;
+            data['type'] = MESSAGE_TYPES.ROLL;
```
kakaroto, line 49: How about whispers a couple of lines above it ?

src/fvtt/page-script.js
```diff
@@ -20,13 +20,22 @@ class FVTTDisplayer {
...
             });
         });
-        return this._postChatMessage(html, character, whisper, play_sound);
+        return this._postChatMessage(html, character, whisper, play_sound, attack_rolls, damage_rolls);
     }
 
-    _postChatMessage(message, character, whisper, play_sound = false) {
+    _postChatMessage(message, character, whisper, play_sound = false, attack_rolls, damage_rolls) {
         const MESSAGE_TYPES = CONST.CHAT_MESSAGE_TYPES || CHAT_MESSAGE_TYPES;
+        const pool = new DicePool([...attack_rolls, ...damage_rolls.map(d => d[1])].map(r => r._roll));
```
kakaroto, line 28: Does it work with the DNDBDice class when using digital dice ?

src/fvtt/page-script.js
```diff
@@ -20,13 +20,22 @@ class FVTTDisplayer {
...
+        pool.roll();
+        const pool_roll = new Roll("");
```
kakaroto, line 30: Would be nice to make the formula here, like (to be tested) :
```js
const formulas = pool.dice.map(d => d.formula);
new Roll(`{${formulas.join(",")}}`);
```

src/fvtt/page-script.js
Aeristoka: This has been resolved

src/fvtt/page-script.js
Aeristoka: Added as requested

src/fvtt/page-script.js
Aeristoka: Changed.

src/fvtt/page-script.js
```diff
@@ -25,9 +25,25 @@ class FVTTDisplayer {
 
     _postChatMessage(message, character, whisper, play_sound = false, attack_rolls, damage_rolls) {
         const MESSAGE_TYPES = CONST.CHAT_MESSAGE_TYPES || CHAT_MESSAGE_TYPES;
-        const pool = new DicePool([...attack_rolls, ...damage_rolls.map(d => d[1])].map(r => r._roll));
+        let pool;
```
kakaroto, line 28: Dont' know why it's :
```
let pool;
pool = new DicePool ...
```
instead of `const pool = new DicePool ...`

src/fvtt/page-script.js
```diff
@@ -25,9 +25,25 @@ class FVTTDisplayer {
...
+        pool = new DicePool([...attack_rolls, ...damage_rolls.map(d => d[1])].map(r => {
+            if (r instanceof FVTTRoll) { return r._roll; }
+            r.class = "Roll";
+            r.dice = [];
+            r.parts = r.parts.map(p => {
+                if (p.formula) {
```
kakaroto, line 34: sounds good to me, alternative would be `if (typeof(p) === "string") return p;` but either way works. I don't have any strong preference actually.

src/common/roll_renderer.js
```diff
@@ -326,7 +326,7 @@ class Beyond20RollRenderer {
         if (request.sendMessage && this._displayer.sendMessage)
             this._displayer.sendMessage(request, title, html, buttons, character, request.whisper, play_sound, source, attributes, description, attack_rolls, roll_info, damage_rolls, total_damages, open)
         else
-            this._displayer.postHTML(request, title, html, buttons, character, request.whisper, play_sound);
+            this._displayer.postHTML(request, title, html, buttons, character, request.whisper, play_sound, attack_rolls, damage_rolls);
```
kakaroto, line 329: I would prefer if all the other arguments get sent as well, since they are available and they might be needed at some point in the future, so I'd prefer if the arguments to `sendMessage` and `postHTML` would just be identical at this point.

src/fvtt/page-script.js
Aeristoka: Old holdover from a broken way to handle DnDBeyond Digital Dice Off vs. On

src/common/roll_renderer.js
Aeristoka: Fixed

kakaroto:
Remaining issue that I see is that you build and set the dice pool and set the message type to `MESSAGE_TYPES.ROLL` in all circumstances. That shouldn't be true for things like "display spell card" for example.
I'd suggest you do something like : 
```js
const rolls = [...attacks_rolls, ...damage_rolls.map(foobar)]
if (rolls.length > 0) {
  // build the dice pool roll
  data.type = MESSAGE_TYPES.ROLL;
  data.roll = pool_roll;
}
```

Aeristoka:
Just pushed a fix for this. Tiny bit of code duplication, but I'm not sure how you'd want to untangle the code otherwise.

kakaroto:
> Just pushed a fix for this. Tiny bit of code duplication, but I'm not sure how you'd want to untangle the code otherwise.

euhh... no. You don't need code duplication. You're currently creating the pool dice even if it's not to be set, and then doing the same code in the `if` statement (for whisper) and `else` statement.

You should do code I pasted in the comment above *after* that if/else, so let it set the data.type to OOC or WHISPER and *then* if there's a roll, override the value. Dice pool gets created only if needed, and message type and roll gets set once.

Aeristoka:
Just ran a re-structure of the logic, it appears to work nicely, and be MUCH prettier now

kakaroto:
Nice, that's what I had in mind... now squash the 7 commits into one and I'll merge it :D

Aeristoka:
Squashed! (And references the issue you have surrounding this, for auto-close purposes)

kakaroto:
Thanks! But you didn't squash it correctly.
You need to squash it clockwise, not counter-clockswise. Man, there's always some issue with your PRs!

PR #414, opened 2020-06-20
Fix for adding damages in some cases

Aeristoka:
Proper addition (with multiple damages):
image

In some cases, the damages are not added (I believe due to the logic for tripping the addition being incorrect):
image

Corrects to this:
image

### DIFF
--- a/dist/dndbeyond_character.js
+++ b/dist/dndbeyond_character.js
@@ -1872,7 +1872,7 @@ class Beyond20RollRenderer {
             }
             html += "<div class='beyond20-roll-result beyond20-roll-cells'>" + this.rollsToCells(roll_html) + "</div>";
         }
-        const add_totals = damage_rolls.filter((r) => (r[2] & DAMAGE_FLAGS.CRITICAL) == 0).length > 1;
+        const add_totals = damage_rolls.filter((r) => (r[2] & DAMAGE_FLAGS.CRITICAL) == 0).length > 1 || damage_rolls.filter((r) => (r[2] & DAMAGE_FLAGS.CRITICAL) != 0).length > 1;
         const total_damages = {}
         for (let [roll_name, roll, flags] of damage_rolls) {
             const is_total = !add_totals && (flags & DAMAGE_FLAGS.CRITICAL) == 0;
--- a/dist/dndbeyond_encounter.js
+++ b/dist/dndbeyond_encounter.js
@@ -1872,7 +1872,7 @@ class Beyond20RollRenderer {
             }
             html += "<div class='beyond20-roll-result beyond20-roll-cells'>" + this.rollsToCells(roll_html) + "</div>";
         }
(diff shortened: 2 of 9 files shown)


kakaroto:
That's the wrong fix.
What it did was : 
"if there is more than one damage, ignoring criticals, then add totals"
and you made it into : 
"if there is at least one damage, ignoring criticals, then add totals"
This means that it will now show things like : 
"Slashing damage : 5
Total Damage: 5"
which is unnecessary.
The proper fix should be : 

"if there is more than one damage, ignoring criticals,  OR if there is more than one critical damage, then add totals".

Aeristoka:
> That's the wrong fix.
> What it did was :
> "if there is more than one damage, ignoring criticals, then add totals"
> and you made it into :
> "if there is at least one damage, ignoring criticals, then add totals"
> This means that it will now show things like :
> "Slashing damage : 5
> Total Damage: 5"
> which is unnecessary.
> The proper fix should be :
> 
> "if there is more than one damage, ignoring criticals, OR if there is more than one critical damage, then add totals".

That's what I observed in my game on Saturday. I don't know what the heck logic you're using in your comparison though. Any info there?

kakaroto:
LOL

(quick lesson) :
The `r[2]` is a bitflag (see https://en.wikipedia.org/wiki/Bit_field) so it's a number, in which each bit represents a boolean. That's why the DAMAGE_FLAGS are all powers of 2, as they need to be different bits in the binary representation of the number from the other flags. The '&' is a binary AND which is used to retrieve a specific bit in the number. If it's 0, then it's false, if it's 1, then it's true.
Yes, this is not very 'javascripty' and is much more 'C code' (or even network-optimized/embedded code). But basically, where in javascript you'd write : 
```js
flags = ["critical", "versatile", "additional-damage"];
```
I instead do : 
```
flags = CRITICAL | VERSATILE | ADDITIONAL;  // That's 2 | 4 | 0x10 = 0x16 (22 in decimal) = 0b10110
// the `|` is a binary OR
```
Where in javascript, I can check if the damage came from a critical hit by doing : 
```js
flags.includes("critical")
```
I instead do : 
```
flags & CRITICAL; // it's 0 if it's not, or it's CRITICAL (0x10) if it is, because of the binary AND operation
```

So, quick lesson done.
```js
const add_totals = damage_rolls.filter((r) => (r[2] & DAMAGE_FLAGS.CRITICAL) == 0).length > 1;
```
This checks all damages, finds the ones which don't have the CRITICAL flag set. Adds totals if the number of damages without CRITICAL flag set is > 1
Fix would be : 
```js
const add_totals = damage_rolls.filter((r) => (r[2] & DAMAGE_FLAGS.CRITICAL) == 0).length > 1 || damage_rolls.filter((r) => (r[2] & DAMAGE_FLAGS.CRITICAL) != 0).length > 1;
```
So this checks, if the number of damage rolls that are not critical are > 1, or the number of damage rolls that are critical is > 1.

PR #441, opened 2020-07-09
Implement Wizard Bladesinger Bladesong

### LINKED ISSUE #436: [Request] Add Wizard (Bladesinger) Bladesong-options, similar to how you implemented a Barbarian's rage

Aeristoka:
Fixes #436

Still could use a query for the Concentration check Constitution Save
Leaving that to @kakaroto

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -452,6 +452,12 @@ const character_settings = {
         "description": "Unleash your divine soul to deal extra radiant damage equal to your level.",
         "type": "bool",
         "default": false
+    },
+    "wizard-bladesong": {
+        "title": "Wizard: Bladesong",
+        "description": "Activate your Bladesong and make your weapon sing with magic",
+        "type": "bool",
+        "default": false
     }
 }
 
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -32,6 +32,9 @@ function rollSkillCheck(paneClass) {
             (character.hasClassFeature("Giant Might") && character.getSetting("fighter-giant-might", false)))) {
         roll_properties["advantage"] = RollType.OVERRIDE_ADVANTAGE;
     }
+    if (skill_name == "Acrobatics" && character.hasClassFeature("Bladesong") && character.getSetting("wizard-bladesong", false)) {
+        roll_properties["advantage"] = RollType.OVERRIDE_ADVANTAGE;
+    }
     // Set Reliable Talent flag if character has the feature and skill is proficient/expertise or a custom skill that needs to be queried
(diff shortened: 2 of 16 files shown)


src/common/settings.js
```diff
@@ -452,6 +452,12 @@ const character_settings = {
         "description": "Unleash your divine soul to deal extra radiant damage equal to your level.",
         "type": "bool",
         "default": false
+    },
+    "wizard-bladesong": {
+        "title": "Bladesong: Your magic causes your weapon to sing",
```
kakaroto, line 457: Format for every other special feat is `${class}: ${feature name}`, so I would do `"Wizard: Bladesong"` here instead, and in the description say something like "Activate your Bladesong and make your weapon sing with magic" or something like that. No need for the details of the mechanics

src/dndbeyond/content-scripts/character.js
```diff
@@ -527,6 +536,10 @@ function rollAction(paneClass) {
                     damage_types.push("Sneak Attack");
                 }
             }
+            if (character.hasClassFeature("Song of Victory") && character.getSetting("wizard-bladesong", false)) {
+                damages.push(String(character.getAbility("INT").mod));
```
kakaroto, line 540: I would verify that the value is available (I don't know, in case it never parsed the abilities?) and isn't 0. There's no need to have a constant damage of 0 if the character has no intelligence

src/common/settings.js
Aeristoka: Fixed in PR update that is incoming.

src/dndbeyond/content-scripts/character.js
Aeristoka: Fixed

src/dndbeyond/content-scripts/character.js
```diff
@@ -527,6 +538,13 @@ function rollAction(paneClass) {
                     damage_types.push("Sneak Attack");
                 }
             }
+            // Wizard: Bladesong
+            if (character.hasClassFeature("Song of Victory") && character.getSetting("wizard-bladesong", false)) {
+                const intelligence = character.getAbility("INT") || {mod: 0};
+                const mod = parseInt(intelligence.mod) || 0;
+                damages.push(String(Math.max(mod, 1)));
+                damage_types.push("Wizard: Bladesong");
```
kakaroto, line 546: ah, I see the misunderstanding when you told me in discord that sneak attack doesn't say "Rogue: Sneak Attack"... 
The comment I made about the `"${class}: ${feature_name}"` was not about the damage, it was about the settings title :p Here it should just say Bladesong

Aeristoka: Well make up your mind!

kakaroto: My mind's made up, I choose the egg

Issue #409, opened 2020-06-17
Initiative Tracker


shadow7412:
Synchronize the state of the DDB initiative tracker to VTTs.

# Core Requirements

* [X] The initiative score entered in DDB should replicate to the relevant VTT.
* [ ] VTT entered intiatives (particularly those by players, either via this extension or manually) should replicate to DDB.
* [X] Moving to the next unit's turn should be synchronised such that the combat can be progressed from ~either VTT or~ DDB.
* [X] Undo feature in DDB should be honored (whatever that does)

# Bonus Requirements

* ~HP bars on relevant token should be updated when tracking damage in DDB (and possibly vice versa) particularly for enemies.~ (not enough information to reliably link token with character, especially with monsters)
* ~Changing the HP of a token in the VTT should replicate back to the DDB tracker.~ (as above)
* [ ] Reduction to 0hp should cause a marker to appear on the token, to clearly show the creature is out of the fight.

shadow7412:
```javascript
$(".combatant-card.in-combat").map((i, x) => [
    $(".combatant-summary__name", x).text(),
    $(".combatant-card__initiative-value", x).text(),
    $(x).hasClass("is-turn"),
])
```

This snippet extracts the names and initiative value of all players in the combat, and whether it's currently their turn - which is more or less all of the information we'd need from DDB.

hypergig:
I would pay for this feature. Also, most folks click the beyond20 initiative button in DDB, which spits it to the VVT, that should somehow synchronize back to my encounter running and the VVT trackers.

I can't tell you how many times I accidentally closed out the tab on my encounter losing everyone's initiative... it would be great if those backfilled or there was a button that read them from the beyond20 chat tables.

I would put a bounty on this feature, honestly It would save a ton of time

shadow7412:
I've started playing with the concept - I'm about 60% through a rough prototype which syncs the DDB tracker to Roll 20.

I've mentioned capturing initiative scores already, but reviving an accidentally refreshed tracker is an interesting idea worth considering. It's not hard to imagine that happening. Of course, some data will be lost (such as what turn we're up to) but getting the initiative scores back should be painless enough...

hypergig:
Especially if they are already in the chat.. and came from the beyond20 in the first place

Daekesh:
[...]

PR #444, opened 2020-07-10
#409 First mostly working implementation of turn tracker.

shadow7412:
I've added a setting for this, which is off by default. Those interested in trying it out can enable it from the settings menu - but it might be a good idea to move this into the quick settings once it's been tested a bit. Because of this, I'm not too worried if this gets merged in.

Although this solution seems workable, there are some limitations that should be addressed;
* Ideally we shouldn't unlink tokens from the tracker. At the moment the list is replaced with a fresh version built from data in the DDB tracker.
* There's no support for hidden units - I see no such support in DDB's combat tracker anyway.
* Initiative values need to be entered by the person running the encounter. There is a way to do this automatically for monsters, but players will need manual entry.

But it's a start. Also looks like my IDE picked up some superfluous whitespace issues again. I can fix it if I need to, but it's probably going to bite me every single time I contribute, so I'd rather just fix them unless you have any specific reason why they should be trimmed.

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -290,10 +290,10 @@ const options_list = {
     },
 
     "sync-combat-tracker": {
-        "title": "Sync D&D Beyond Encounter with VTT turn tracker",
-        "description": "Currently only supports roll 20.",
+        "title": "Synchronize the Combat Tracker with the VTT",
+        "description": "Overwrites the VTT's combat tracker with the details from D&D Beyond's Encounter tool (Roll20 only)",
         "type": "bool",
-        "default": false
+        "default": true
     },
 
     "donate": {
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -289,6 +289,13 @@ const options_list = {
         "default": ""
     },
 
+    "sync-combat-tracker": {
+        "title": "Sync D&D Beyond Encounter with VTT turn tracker",
+        "description": "Currently only supports roll 20.",
+        "type": "bool",
+        "default": false
+    },
+
     "donate": {
         "short": "Buy rations (1 day) to feed my familiar",
         "title": "Become a patron of the art of software development!",
@@ -933,7 +940,7 @@ function setDiscordChannelsSetting(name, settings) {
     if (!dropdowns.find(d => d.active)) dropdowns[0].active = true;
     if (dropdowns.find(d => d.secret)) dropdowns.push({ name: "Delete selected channel", action: "delete" })
     dropdowns.push({ name: "Add new channel", action: "add" })
-    
+
     console.log("Added new options", dropdowns);
     fillDisordChannelsDropdown(name, dropdowns);
 }
@@ -951,7 +958,7 @@ function fillDisordChannelsDropdown(name, dropdowns, triggerChange=false) {
     const active = dropdowns.find(d => d.active);
     input.text(active.name);
     input.attr("data-secret", active.secret.slice(0, 12));
-    
+
     $("#beyond20-option-discord-channels li").off('click').click(ev => {
         ev.stopPropagation();
         ev.preventDefault()
@@ -971,7 +978,7 @@ function fillDisordChannelsDropdown(name, dropdowns, triggerChange=false) {
     $("#beyond20-option-discord-channels li[data-action=add]").off('click').click(ev => {
         ev.stopPropagation();
         ev.preventDefault()
-        
+
         dropdown_menu.removeClass('open');
         button_group.removeClass('open');
         m.set('triangle').size(10);
--- a/src/dndbeyond/content-scripts/encounter.js
+++ b/src/dndbeyond/content-scripts/encounter.js
@@ -13,6 +13,9 @@ function documentModified(mutations, observer) {
 
     const monster = $(".encounter-details-monster-summary-info-panel,.encounter-details__content-section--monster-stat-block,.combat-tracker-page__content-section--monster-stat-block,.monster-details-modal__body");
     const monster_name = monster.find(".mon-stat-block__name").text();
+    if (settings["sync-combat-tracker"]) {
+        updateCombatTracker();
+    }
     console.log("Doc modified, new mon : ", monster_name, " !=? ", last_monster_name);
     if (monster_name == last_monster_name)
         return;
@@ -20,6 +23,23 @@ function documentModified(mutations, observer) {
     removeRollButtons();
     character = new Monster("Monster", null, settings);
     character.parseStatBlock(monster);
+}
+
+function updateCombatTracker() {
+    if ($(".turn-controls__next-turn-button").length) {
+        $(".ui-dialog-buttonpane").hide(); // get rid of turn controls, to make it more obvious to do things from DDB.
+        const combat = Array.from($(".combatant-card.in-combat")).map(x => [
+            $(".combatant-summary__name", x).text(),
+            $(".combatant-card__initiative-value", x).text(),
+            $(x).hasClass("is-turn"),
+        ]);
+        const req = {
+            action: "combat-tracker",
+            combat,
+        };
+        console.log("Sending combat update", combat);
+        chrome.runtime.sendMessage(req, resp => beyond20SendMessageFailure(character, resp));
+    }
 }
 
 function updateSettings(new_settings = null) {
@@ -48,5 +68,5 @@ updateSettings();
 injectCSS(BUTTON_STYLE_CSS);
 chrome.runtime.onMessage.addListener(handleMessage);
 const observer = new window.MutationObserver(documentModified);
-observer.observe(document, { "subtree": true, "childList": true });
+observer.observe(document, { "subtree": true, "childList": true, attributes: true, });
 chrome.runtime.sendMessage({ "action": "activate-icon" });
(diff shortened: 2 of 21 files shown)


src/extension/background.js
```diff
@@ -148,10 +148,19 @@ function onRollFailure(request, sendResponse) {
     });
 }
 
+
+const acceptedActions = [
```
kakaroto, line 152: This is more of `forwardedActions` than `acceptedActions`

src/common/settings.js
```diff
@@ -289,6 +289,13 @@ const options_list = {
         "default": ""
     },
 
+    "sync-combat-tracker": {
+        "title": "Sync D&D Beyond Encounter with VTT turn tracker",
+        "description": "Currently only supports roll 20.",
```
kakaroto, line 294: I would make this also into a quick setting for the encounters page itself (see how popup.js checks for the page and adds custom stuff to the options, like the monster whispers option)
I'd change the title to : "Synchronize the Combat Tracker with the VTT" (maybe?)
Change description to : "Synchronizes D&D Beyond's Combat Tracker from your Encounters with the VTT's combat tracker (Roll20 only)"

src/common/settings.js
```diff
@@ -289,6 +289,13 @@ const options_list = {
...
+        "type": "bool",
+        "default": false
```
kakaroto, line 296: maybe default it to true?

src/dndbeyond/content-scripts/encounter.js
```diff
@@ -22,6 +25,23 @@ function documentModified(mutations, observer) {
     character.parseStatBlock(monster);
 }
 
+function updateCombatTracker() {
+    if ($(".turn-controls__next-turn-button").length) {
```
kakaroto, line 29: Use early return paradigm instead : 
```js
if ($(".turn-controls__next-turn-button").length === 0) return;
```

src/dndbeyond/content-scripts/encounter.js
```diff
@@ -22,6 +25,23 @@ function documentModified(mutations, observer) {
...
+        $(".ui-dialog-buttonpane").hide(); // get rid of turn controls, to make it more obvious to do things from DDB.
+        const combat = Array.from($(".combatant-card.in-combat")).map(x => [
+            $(".combatant-summary__name", x).text(),
+            $(".combatant-card__initiative-value", x).text(),
+            $(x).hasClass("is-turn"),
+        ]);
```
kakaroto, line 35: I see what you do, but it's a little difficult to read. I suggest naming it `combatant` instead of `x` and using `$(x).find(selector)` instead of `$(selector, x)`
I also know that you probably used an array like this because I also did that in other parts of the code (when I was still using python's tuple mentality), but I think it would be much better and easier to maintain now if we use an object.

How about : 
```js
const combatants = $(".combatant-card.in-combat")
const combat = combatants.toArray().map(combatant => {
  combatant = $(combatant)
  return {
    name: combatant.find(".combatant-summary__name").text(),
    initiative: combatant.find(".combatant-summary__name").text(),
    turn: combatant.hasClass("is-turn")
  };
});
```

src/dndbeyond/content-scripts/encounter.js
```diff
@@ -22,6 +25,23 @@ function documentModified(mutations, observer) {
...
+        const req = {
+            action: "combat-tracker",
+            combat,
+        };
+        console.log("Sending combat update", combat);
```
kakaroto, line 40: I would name the action "update-combat" to be consistent with the other actions, and would only send the message if the setting is enabled (i.e: check the setting here rather than in the VTT)

src/roll20/page-script.js
```diff
@@ -28,6 +28,23 @@ function updateHP(name, current, total, temp) {
     }
 }
 
+function updateCombatTracker(combat) {
+    const index = combat.findIndex(x => x[2]);
```
kakaroto, line 32: Using object as suggested above would make this into `findIndex(combatant => combatant.turn)` which makes it more readable for future maintenance

src/roll20/page-script.js
```diff
@@ -28,6 +28,23 @@ function updateHP(name, current, total, temp) {
...
+    if (index === -1) {
+        console.warn("It's apparently nobody's turn :/");
+    } else {
+        const c = combat.splice(index, combat.length);
+        combat.splice(0, 0, ...c);
```
kakaroto, line 37: Nice, so you're splicing the combatant whose turn it is and anyone after them and placing them at the beginning. I'd just add a comment to explain it 
`// Re-order list with the combatant whose turn it is at the top of the list, as that's how Roll20 needs it displayed`

src/dndbeyond/content-scripts/encounter.js
shadow7412: Oops, I deleted the message here.

Even renamed to update-combat, and that is already done in the if statement protecting whether this function gets called.

src/dndbeyond/content-scripts/encounter.js
```diff
@@ -26,20 +26,24 @@ function documentModified(mutations, observer) {
...
-            $(x).hasClass("is-turn"),
-        ]);
-        const req = {
-            action: "combat-tracker",
-            combat,
-        };
-        console.log("Sending combat update", combat);
-        chrome.runtime.sendMessage(req, resp => beyond20SendMessageFailure(character, resp));
+    if (!$(".turn-controls__next-turn-button").length) {
+        return;
```
kakaroto, line 32: oufff, I don't like one liners that use accolades.. I'll let it slide though...

src/dndbeyond/content-scripts/encounter.js
```diff
@@ -26,20 +26,24 @@ function documentModified(mutations, observer) {
...
     }
+    $(".ui-dialog-buttonpane").hide(); // get rid of turn controls, to make it more obvious to do things from DDB.
```
kakaroto: oh yeah, right, I wanted to check this, what are you hiding here? From the comment it sounds like you're hiding controls from roll20, but that wouldn't work since this is the encounter's page content script. Also, I would not hide any UI element since there's no way to unhide it

src/dndbeyond/content-scripts/encounter.js
```diff
@@ -25,11 +25,12 @@ function documentModified(mutations, observer) {
     character.parseStatBlock(monster);
 }
 
+let lastCombatUpdate = null;
```
kakaroto, line 28: I'd use var in this case. As it's a global variable. I consider 'let' to be for small scope temp variables. I may be wrong in how people do it, but that's the convention I've been using.

src/dndbeyond/content-scripts/encounter.js
```diff
@@ -38,6 +39,12 @@ function updateCombatTracker() {
             turn: $combatant.hasClass("is-turn"),
         };
     });
+    const json = JSON.stringify(combat);
+    if (lastCombatUpdate === json) {
+        return;
```
kakaroto, line 44: another one liner if with accolades 😭

src/dndbeyond/content-scripts/encounter.js
shadow7412: Took me a bit to work out what you meant by accolades - but here's some light reading: https://medium.com/@jonathanabrams/single-line-if-statements-2565c62ff492

src/dndbeyond/content-scripts/encounter.js
shadow7412: Yeah I realised that when I was doing the other changes, and have since removed it. I'm still on the fence as to whether we should be hiding them, especially without a way to bring them back should they decide to stop using the tracker mid-combat.

src/dndbeyond/content-scripts/encounter.js
shadow7412: I'm personally against using deprecated keywords as a rule... but if you insist further I can do that (but it'll hurt).

src/dndbeyond/content-scripts/encounter.js
shadow7412: Always :)

src/dndbeyond/content-scripts/encounter.js
kakaroto: Oh sorry, 'accolades' is the french term, I forget it's called braces in english.
And yep, I'm aware of the risk of it and the possibility of making mistakes because of that. It's a personal choice though, I really don't like seeing one liners with the braces because visually it feels too heavy. For early return like that with the single return statement, I'd have put it on the same line as well (I also used to hate if statements with their statement on the same line but gotten used to it recently for short return statements only).

I'm used to it though, so I don't fall into the trap of editing code and forgetting to add the braces.
Like I said, I'll let it slide though. :p

src/dndbeyond/content-scripts/encounter.js
kakaroto: Yeah, I'd definitely say not to hide it. I don't see a point in it, and it could only cause problems (due to the software) rather than eliminate them (due to the user). I prefer that if there is a problem, it's the user's fault rather than the software's :p

src/dndbeyond/content-scripts/encounter.js
kakaroto: My understanding is that `var` isn't deprecated, but rather it's used for something different. Specifically, `let` is for the current block scope, same as `const` but `var` is for the function's scope or for global variables. People say to use `let` in favor of `var` because it makes it less error prone, as you may get confused about your variables and the variable may be in the wrong scope, etc... totally understandable and good avice. In this case however, we are declaring an actual global variable, which is why I've taken to declaring them with `var` to make it clear that **hey look, this is a global variable right here**.

PR #446, opened 2020-07-11
Limit combat updates to roll 20 instances running in GM mode.

shadow7412:


kakaroto:
Sorry, already did that locally, I'm writing release notes now.

PR #458, opened 2020-07-14
Fix ajs-content overlapping with ajs-header in "Beyond20 Quick Settings"

John-Paul-R:
'Tis but a small visual fix to the "Beyond20 Quick Settings." (Tested on Chrome version 83.0.4103.116 (64 bit) on Windows 10)

## Before
Before
## After
After

### DIFF
--- a/libs/css/alertify-themes/beyond20.css
+++ b/libs/css/alertify-themes/beyond20.css
@@ -27,3 +27,9 @@
   -webkit-animation-duration: 100ms;
           animation-duration: 100ms;
 }
+
+/* Prevent ajs-content from overlapping with ajs-header in "Beyond20 Quick Settings" */
+.alertify.ajs-resizable .ajs-body .ajs-content,
+.alertify.ajs-maximized .ajs-body .ajs-content {
+  top: 64px;
+}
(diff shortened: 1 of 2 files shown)


kakaroto:
Woohoo, thanks! That has bugged me for a long time but I couldn't figure out why it happened (stupid css!)!
Can you make the change as an override in the beyond20.css file instead though so the original alertify themes remain unmodified ?
https://github.com/kakaroto/Beyond20/blob/master/libs/css/alertify-themes/beyond20.css
Thanks!

John-Paul-R:
Done! The property has been changed back to its original value in alertify.css and an override has been added in beyond20.css instead.

kakaroto:
Awesome, thanks! 
Merged, and welcome to the contributors list! 🥳

hypergig:
One suggestion: maybe make the donation button a bit more predominant in this page? Put it on to in it's own div on top, give it it's own color, something that sticks out like pink, or orange?

kakaroto:
I don't think it's necessary. I don't like donate buttons that are too aggressive. Having it both in the quick settings and full settings is enough as is already. I'd like people to donate because they want to, not because it's annoyingly **in their face** :p

hypergig:
hahahaha, fair enough

PR #460, opened 2020-07-17
Feature: Send text from element Notes or Description as message to VTT (before, after, or replacing normal roll action).

### LINKED ISSUE #112: Trigger macros on attacks/spells
This can be used to call /fx for example.
From Discord @spideyct :
> Hey there, just curious, with the integration of dndbeyond and roll20, is there a place i can edit a spell description to include the spell special effects to be called?
https://wiki.roll20.net/Text_Chat#Special_Effects_.28.2Ffx.29

> either on the dndbeyond char sheet by editing spell descriptions, or integrated on roll20 side to call a macro after it spits out info?

### LINKED ISSUE #455: Pass along Notes section from DND beyond to Roll20
If it's not a big lift, would it be possible to pass along the notes section to the roll for a spell attack?
Specifically, i'm looking for a way to add images to my attack rolls for a bit of flair :-)  Like "[ri](https://j.gifs.com/yozQ7W.gif)"  which in roll20 i'd normally put in the "Description" field.

John-Paul-R:
The changes in this branch add a feature that will parse any text in the "Notes" or "Description" blocks in the pane of an item, action, or spell that is in the format displayed below and send it as a message to the VTT. (In this version, only Roll20 is supported, mostly because I have no idea how FVTT works.)

```text
```[msg-type] Place your text that you wish to send to the VTT here!

this text can span any number of lines.```
```
Replace 'msg-type' with "before", "after", or "replace" depending on how you want to affect the message or action that would normally be sent to the VTT. (They do what they sound like.)

This also adds an 'info' settings option type, which displays only 'title' and 'description' text. (and adds its own entry, briefly outlining how to use the feature.)

*Resolves #455 (Original Feature Request)*
*Indirectly resolves #112*


src/common/settings.js
```diff
@@ -613,6 +623,13 @@ function createHTMLOptionEx(name, option, short = false) {
                 )
             )
         );
+    } else if (option.type == "info") {
+        e = E.li({ class: "list-group-item beyond20-option beyond20-option-info" },
+            E.label({ class: "list-content", for: name },
+                E.h4({}, title),
+                ...description_p
+            )
+        );
```
kakaroto, line 632: oh nice, I was going to say, you shouldn't need a setting for this, then realized it's a new info type. The \`\`\` in the description would look weird since this isn't markdown though.

src/dndbeyond/content-scripts/character.js
```diff
@@ -857,24 +857,97 @@ function displayAction(paneClass) {
...
+            rollItem();
+        else if (["ct-action-pane", "ct-custom-action-pane"].includes(paneClass))
+            rollAction(paneClass);
+        else if (paneClass == "ct-spell-pane")
+            rollSpell();
+        else
+            displayPanel(paneClass);  
+    }
+
+    //TODO allow adding additional custom roll modifiers within custom text?
```
kakaroto, line 938: It might make more sense to have the execute just pass along the callback to handleCustomText, but I'm not sure about the design here, something seems off to me.

src/dndbeyond/content-scripts/character.js
```diff
@@ -857,24 +857,97 @@ function displayAction(paneClass) {
     });
 }
 
+function handleCustomText(paneClass) {
+    // Function Objective: Search for and handle block of VTT Message text...
+    //...in pane Notes or Description (Priority descending in that order.)
+    const regexp = /```\[([^\]]*)\]\s*((\S|\s)*?)\s*```/g;
```
kakaroto, line 863: The regexp would match `[]` on its own. Also, I don't see the need for the \`\`\` I think it's just confusing, if we use `[before-beyond20] as the mark, and anything on that line gets sent as a macro basically, that should suffice to make it clear.
I'd replace the regexp with this : 
`/^\[((?:before-beyond20)|(?:after-beyond20)|(?:replace-beyond20))\]\s*(.*)$/`
(did not test)
Note that with `/g`, you'll catch all occurences, and you're not handling multiple cases correctly.

src/dndbeyond/content-scripts/character.js
```diff
@@ -857,24 +857,97 @@ function displayAction(paneClass) {
...
+    
+    let pane = $("."+paneClass);
+    let elemText;
+    let matchedText;
+    
+    // Search for msg text in "Notes" section, if it exists
+    //Notes: Iterate through classes in pane: "ddbc-property-list__property" that contain the text "Notes:"
+    //  If exists: load text from "ddbc-property-list__property-content" (Last child of "ddbc-property-list__property")
+    let msgElement = pane.find(".ddbc-property-list__property:contains('Notes:')").get(0)
+    if (msgElement!=undefined && msgElement!=null) {
```
kakaroto, line 877: can just be `if (msgElement)`
Also, please stay consistent with the rest of the coding style, where I would put spaces in between all operators.

src/dndbeyond/content-scripts/character.js
```diff
@@ -857,24 +857,97 @@ function displayAction(paneClass) {
...
+        elemText = msgElement.lastChild.innerText;
+        // console.log("Notes Text: "+elemText);
+        matchedText = [...elemText.matchAll(regexp)];
+    }
```
kakaroto, line 881: This whole section can be replaced with : 
```js
const notes = descriptionToString(`.${paneClass} .ddbc-property-list__property:contains('Notes:')`)
const notesMatch = notes.match(regexp);
```
By the way, this `:contains()` thing is amazing!

src/dndbeyond/content-scripts/character.js
```diff
@@ -857,24 +857,97 @@ function displayAction(paneClass) {
...
+    // If no message was found, search for msg in the "Description" section, if it exists
+    if (matchedText==undefined || matchedText==null) {
+        msgElement = pane.find(".ct-action-detail__description").get(0);
+        console.log(msgElement);
+        if (msgElement!=undefined) {
+            elemText = msgElement.innerText;
+            // console.log("Description Text: "+elemText);    
+            matchedText = [...elemText.matchAll(regexp)]; 
+        }
+    }
```
kakaroto, line 892: Seems you're only looking for `.ct-action-detail__description`, there's also `.ct-spell-detail__description` and same for item, also using `ddbc` instead of `ct`. Look for `__description` in the file to see the other things that could match. Along with previous notes search, I'd rewrite it like this : 
```js
const pane = $(`.${paneClass}`);
const notes = descriptionToString(pane.find(".ddbc-property-list__property:contains('Notes:')")
const notesMatch = notes.match(regexp);
const description = descriptionToString(pane.find(".ct-action-detail__description, .ct-spell-detail__description, .ct-item-detail__description, .ddbc-action-detail__description, .ddbc-spell-detail__description, .ddbc-item-detail__description,")
const descriptionMatch = description.match(regexp);
const customMacro = notesMatch || descriptionMatch; // Use this
```

src/dndbeyond/content-scripts/character.js
```diff
@@ -857,24 +857,97 @@ function displayAction(paneClass) {
...
+            if (rollOrderTypes.includes(cMatchType))
+                rollOrder = cMatchType;
+            rollFunc = function () { 
+                sendRollWithCharacter("action", 0, {
+                    "type": "raw-text",
+                    "text": cMatchText
+                })
+            };
+        }
+    }
```
kakaroto, line 912: This block doesn't break when it finds one match, so if you have multiple, it will use the last but might use the roll order from another line if the last one didn't match a valid order type, Also the rollFunc gets set and rollOrder remain as 'before' if you simply have `[]`.
The modified regexp which checks for specific valid values, and the use of `match` instead of `matchAll` should fix those issues. Here, I would just replace it with : 
```js
if (customMatch) {
  const sendChat = function () { 
                sendRollWithCharacter("action", 0, {
                    "type": "chat-message",
                    "text": customMatch[2]
                })
            };
  switch(customMatch[1]) {
   case "before-beyond20":
      sendChat();
      executeRoll();
   break;
   case "after-beyond20":
      executeRoll();
      sendChat();
   break;
   case "replace-beyond20":
      sendChat();
   break;
  default:
   // Shouldn't happen, but just in case, fallback
  executeRoll()
   break;
  }
} else {
  executeRoll();
}
```

src/roll20/content-script.js
```diff
@@ -798,6 +798,8 @@ function handleMessage(request, sender, sendResponse) {
             roll = rollSpellCard(request);
         } else if (request.type == "spell-attack") {
             roll = rollSpellAttack(request, custom_roll_dice);
+        } else if (request.type == "raw-text") {
```
kakaroto, line 801: I'd prefer "chat-message" instead of 'raw-text'

John-Paul-R:
I should be able to work on implementing these changes later this week. I will update here when the changes are done, or if I have any questions about your comments. Thanks for the review! o7

src/dndbeyond/content-scripts/character.js
John-Paul-R: The goal of using the ` ``` ` delimiters was to allow users to send multi-line content to the VTT without needing to send many individual messages. 
Example multiline message: 
```
PlayerName attacks with flavour text!
AttackRoll Here
It does DamageRoll damage!
More optional flavour text and inline content here!
```
Using ` ``` ` as the delimiter was an arbitrary choice. In theory, it could be anything. (Although I avoided using `[]`, `{}`, and `()`, as Roll20 uses them in their macros, and I was trying to distinguish between Beyond20 syntax and Roll20 (and keep parsing simple)).

I do think that some sort of multi-line support would be a good idea though. What do you think?

John-Paul-R:
I've addressed most of the requested revisions in 0d72d7b (changes outlined in commit message).

I'm still debating how to denote the start/end of a message w/in a description. (more info [here](https://github.com/kakaroto/Beyond20/pull/460#discussion_r473541896))

src/dndbeyond/base/utils.js
```diff
@@ -54,12 +54,12 @@ function propertyListToDict(propList) {
...
-        .replace("&apos;", "\'")
-        .replace("&lt;", "<")
-        .replace("&gt;", ">");
+        .replace(/&nbsp;/g, " ")
+        .replace(/&amp;/g, "&")
+        .replace(/&quot;/g, "\"")
+        .replace(/&apos;/g, "\'")
+        .replace(/&lt;/g, "<")
+        .replace(/&gt;/g, ">");
 }
```
kakaroto, line 63: Good catch, though I'd prefer that as an independent commit (no need to redo it separately)

src/dndbeyond/content-scripts/character.js
kakaroto: Good point about multiline, I hadn't thought of that. I don't like the \`\`\` though, it would be confusing to a lot of people I think who aren't used to markdown or aren't tech savvy (I expect half the users will complain that `'` doesn't work not realizing \` is a different character).
How about `[before] some test [/before]` ?

src/dndbeyond/content-scripts/character.js
```diff
@@ -873,24 +873,79 @@ function displayAction(paneClass) {
...
+            }
+        },
+    };
+    const regexp = /```\[([^\]]*)\]\s*((\S|\s)*?)\s*```/g;
+    const rollOrderTypes = ["before", "after", "replace"];//(relative to normal roll msg)
+    const paneText = descriptionToString($(`.${paneClass}`).find(".ddbc-property-list__property:contains('Notes:'), .ct-action-detail__description, .ct-spell-detail__description, .ct-item-detail__description, .ddbc-action-detail__description, .ddbc-spell-detail__description, .ddbc-item-detail__description"));
+    console.log("Pane Text: " + paneText);
+    const matchedText = [...paneText.matchAll(regexp)];
+    //Handle Matched Text
+    if (matchedText) {
```
kakaroto, line 899: `matchedText` is the array you defined above so it will always evaluate to true. You'd  want to test for `matchedText.length`

src/dndbeyond/content-scripts/character.js
```diff
@@ -873,24 +873,79 @@ function displayAction(paneClass) {
...
+        console.log("Matched Text:\n"+matchedText);
```
kakaroto, line 900: coding style: spaces around `+` operator

src/dndbeyond/content-scripts/character.js
```diff
@@ -873,24 +873,79 @@ function displayAction(paneClass) {
...
+            const cMatchType = matchedText[i][1].toLowerCase();
+            const cMatchText = matchedText[i][2];
+            // Determine if we should prevent the default action/message from being sent to the VTT
+            console.log("Note Match #"+i+" (Provided Type: "+cMatchType+"):\n"+cMatchText);
+            let rollOrder = rollOrderTypes[0];
+            if (rollOrderTypes.includes(cMatchType)) {
+                rollOrder = cMatchType;
+            } else {
+                console.warn(`Invalid CustomText roll order specified ("${cMatchType}"). Defaulting to "${rollOrder}" instead.`)
+            }
```
kakaroto, line 911: I don't like that you match all `[anything]` content and default to `before`. If it's invalid, then it should be ignored.

kakaroto (approved):
Thanks for the changes you've made. I did another quick review, and had some concerns but nothing major. I'll be fixing those comments I had right now, test on FVTT and merge it, so you don't have anything to do.
Thanks for the contribution!

src/dndbeyond/content-scripts/character.js
```diff
@@ -873,24 +873,79 @@ function displayAction(paneClass) {
...
+                    customTextFuncList[i]();
+                }
+            } else if (normalRollFunction) {
+                normalRollFunction();
+            }
+        },
+    };
+    const regexp = /```\[([^\]]*)\]\s*((\S|\s)*?)\s*```/g;
+    const rollOrderTypes = ["before", "after", "replace"];//(relative to normal roll msg)
+    const paneText = descriptionToString($(`.${paneClass}`).find(".ddbc-property-list__property:contains('Notes:'), .ct-action-detail__description, .ct-spell-detail__description, .ct-item-detail__description, .ddbc-action-detail__description, .ddbc-spell-detail__description, .ddbc-item-detail__description"));
```
kakaroto, line 895: Forgot to mention on this one, it won't work, because the `selector.html()` will return the html of the first element if multiple elements are matching, so the single query to find notes and description and passing it to `descriptionToString` will only return the Notes.

kakaroto:
It's merged, I've found a few more issues while testing and I had to change some stuff. You can see my changes here and let me know if you feel any of that is wrong : https://github.com/kakaroto/Beyond20/commit/21e78e0765167cc2c8ff793411bc34040ad07721
Thanks again.

PR #462, opened 2020-07-17
Enhancement: Support rolling instruments as tools.

### LINKED ISSUE #387: support for more tools (instruments)
Would it be possible to add the option to roll for instruments as well, just like other tools (for example Thieves' Tools, etc). They pretty much have the same description as the other tools which already have that option _(If you have proficiency with a given musical instrument, you can add your proficiency bonus to any ability checks you make to play music with the instrument.)_.

Thanks in advance :)

John-Paul-R:
Adds the ability to roll instruments as tools (1d20+ability+prof+custom) as suggested in #387.
Resolves #387.


src/dndbeyond/content-scripts/character.js
```diff
@@ -124,6 +124,7 @@ function rollItem(force_display = false) {
     //console.log("Properties are : " + String(properties));
     const item_name = $(".ct-item-pane .ct-sidebar__heading .ct-item-name,.ct-item-pane .ct-sidebar__heading .ddbc-item-name")[0].firstChild.textContent;
     const item_type = $(".ct-item-detail__intro").text();
+    const item_tags = $(".ct-item-detail__tags-list").find(".ct-item-detail__tag").map(function(){ return $(this).text().toLowerCase(); }).toArray();
```
kakaroto, line 127: `$(".ct-item-detail__tags-list").find(".ct-item-detail__tag")` should be `$(".ct-item-detail__tags-list .ct-item-detail__tag")`

`.map(function(){ return $(this).text().toLowerCase(); }).toArray()`  should be `.toArray().map(el => el.textContent)`
Didn't realize we can `.map` on the jquery object itself, but since you're doing a `toArray` anyways, might as well do it first. Also, I don't see a reason to lowercase the tags, or to use jquery to just get the text content.
This also means the check in the roll renderer should check for the non-lowercase version.
Smart idea to send the tags, I thought I'd have to keep a list of hardcoded instrument names

src/roll20/renderer.js
```diff
@@ -1,47 +1,47 @@
...
+
+class Roll20Prompter {
+    async prompt(title, html, ok_label = "OK", cancel_label = "Cancel") {
+        return new Promise((resolve, reject) => {
+            alertify.Beyond20Prompt(title, html, ok_label, cancel_label, resolve);
+        });
+    }
+}
+
+var roll_renderer = new Beyond20RollRenderer(new Roll20Roller(), new Roll20Prompter(), new Roll20Displayer());
```
kakaroto, line 46: any changes here? or it's all whitespace/CRLF changes ?

John-Paul-R: No content changes were made to oll20/renderer.js. Just CRLF > LF.

src/dndbeyond/content-scripts/character.js
John-Paul-R: Good points! I've now implemented these suggestions.

and ty! :P

kakaroto:
Sorry for the delay, I thought I had merged this already!

PR #473, opened 2020-07-24
Spell Description on Spell Attack for Roll20

Aeristoka:
Adds a toggle to display a Spell Description on a Spell Attack on Roll20
Logic to interact with the different Material Display Options for Roll20

Request from MaxPat on Beyond20 Discord

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -254,6 +254,13 @@ const options_list = {
         "choices": { "all": "All components", "material": "Only material components", "none": "Do not display anything" }
     },
 
+    "roll20-spell-description-display": {
+        "title": "Display Spell Descriptions in spell attacks",
+        "description": "When doing a spell attack, display the spells full description (Roll20 only toggle)",
+        "type": "bool",
+        "defult": false
+    },
+
     "component-prefix": {
         "title": "Component Prefix",
         "description": "Prefix to the components display of a spell attack.\nIf displaying material components only, you may want to set it to 'Materials used :' for example",
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -258,7 +258,7 @@ const options_list = {
         "title": "Display Spell Descriptions in spell attacks",
         "description": "When doing a spell attack, display the spells full description (Roll20 only toggle)",
         "type": "bool",
-        "defult": false
+        "default": false
     },
 
     "component-prefix": {
--- a/src/roll20/content-script.js
+++ b/src/roll20/content-script.js
@@ -544,7 +544,19 @@ function rollSpellAttack(request, custom_roll_dice) {
             }
         }
     }
-
+    if (settings["roll20-spell-description-display"] === true) {
+        if (settings["components-display"] === "all") {
+            properties["desc"] += "\r\n\r\nDescription: " + request.description;
+        } else if (settings["components-display"] === "material") {
+            if (properties["desc"] != undefined) {
+                properties["desc"] += "\r\n\r\nDescription: " + request.description;
+            } else {
+                properties["desc"] = "Description: " + request.description;
+            }
+        } else {
+            properties["desc"] = "Description: " + request.description;
+        }
+    }
     if (request.rollDamage && !request.rollAttack) {
         template_type = "dmg";
         dmg_props["charname"] = request.character.name;
--- a/src/roll20/content-script.js
+++ b/src/roll20/content-script.js
@@ -545,17 +545,8 @@ function rollSpellAttack(request, custom_roll_dice) {
         }
     }
     if (settings["roll20-spell-description-display"] === true) {
-        if (settings["components-display"] === "all") {
-            properties["desc"] += "\n\nDescription: " + request.description;
-        } else if (settings["components-display"] === "material") {
-            if (properties["desc"] != undefined) {
-                properties["desc"] += "\n\nDescription: " + request.description;
-            } else {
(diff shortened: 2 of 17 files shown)


src/common/settings.js
```diff
@@ -229,6 +229,13 @@ const options_list = {
         "choices": { "all": "All components", "material": "Only material components", "none": "Do not display anything" }
     },
 
+    "roll20-spell-description-display": {
+        "title": "Display Spell Descriptions in spell attacks",
+        "description": "When doing a spell attack, display the spells full description (Roll20 only toggle)",
+        "type": "bool",
+        "defult": false
```
kakaroto, line 236: `defult`? :p

src/roll20/content-script.js
```diff
@@ -583,7 +583,21 @@ function rollSpellAttack(request, custom_roll_dice) {
...
+        } else if (settings["components-display"] === "material") {
+            if (properties["desc"] != undefined) {
+                properties["desc"] += "\r\n\r\nDescription: " + request.description;
+            } else {
+                properties["desc"] = "Description: " + request.description;
+            }
+        } else {
+            properties["desc"] = "Description: " + request.description;
+        }
+    }
```
kakaroto, line 599: I don't really understand the logic here with the `\r\n` and all that. I would say that it should instead do a display (i.e: spellcard template) before the actual roll.

src/common/settings.js
Aeristoka: spulleng is herd

src/roll20/content-script.js
Aeristoka: Could be done, the \r\n is to get two returns set after the Spell Components are displayed, and it makes it display in a very Roll20-reminiscent way

Aeristoka: Roll20 Native:
image

Beyond20 with the above code:
image

kakaroto: Yeah, I figured `\r\n` was for adding newline (also only need `\n`), but I didn't understand the logic for why it checks for the component display and had all this logic instead of just checking if `desc` is set or not, regardless of the settings. Like : 
```js
properties["desc"] = `${properties["desc"] ? properties["desc"] + "\n\n" : ""}Description: ${request.description}`;
```

Anyway, some spells are really **huge**, and since the To-Hit is separated from the damage with that potential wall of text, I think it would be a lot less issue-prone to have it the spell card posted instead.
Something like (pseudo code in the main onMessage function) : 
```
if (roll20 description display) { 
   spell_description = displaySpell(request)
   roll = spell_description + "\n" + roll
}
```

Aeristoka: I think I still like my way better, because it lets you display All (Description + Components), Components only, or the regular none, and it looks super duper Roll20.

kakaroto:
Ready to merge this, I'll accept the change as you suggested, but can you do 2 things.
1 - Fix the code so it's cleaner and easier to read, changing this : 
```js

    if (settings["roll20-spell-description-display"] === true) {
        if (settings["components-display"] === "all") {
            properties["desc"] += "\n\nDescription: " + request.description;
        } else if (settings["components-display"] === "material") {
            if (properties["desc"] != undefined) {
                properties["desc"] += "\n\nDescription: " + request.description;
            } else {
                properties["desc"] = "Description: " + request.description;
            }
        } else {
            properties["desc"] = "Description: " + request.description;
        }
    }
```
into : 
```js

    if (settings["roll20-spell-description-display"] === true) {
		properties["desc"] = properties["desc"] ? properties["desc"] + "\n\n" : "";
		properties["desc"] += `\n\nDescription: ${request.description}`;
    }
```

2 - rebase on master and rebuild so I can easily merge it.

Aeristoka:
I've got conflicts here, and I'm not sure I understand what you changed here that broke it... (my merge is not super clear, is what I'm trying to say)

> Ready to merge this, I'll accept the change as you suggested, but can you do 2 things.
> 1 - Fix the code so it's cleaner and easier to read, changing this :
> 
> ```js
>     if (settings["roll20-spell-description-display"] === true) {
>         if (settings["components-display"] === "all") {
>             properties["desc"] += "\n\nDescription: " + request.description;
>         } else if (settings["components-display"] === "material") {
>             if (properties["desc"] != undefined) {
>                 properties["desc"] += "\n\nDescription: " + request.description;
>             } else {
>                 properties["desc"] = "Description: " + request.description;
>             }
>         } else {
>             properties["desc"] = "Description: " + request.description;
>         }
>     }
> ```
> 
> into :
> 
> ```js
>     if (settings["roll20-spell-description-display"] === true) {
> 		properties["desc"] = properties["desc"] ? properties["desc"] + "\n\n" : "";
> 		properties["desc"] += `\n\nDescription: ${request.description}`;
>     }
> ```
> 
> 2 - rebase on master and rebuild so I can easily merge it.

Rebased, but I'm not understanding your ask here:

Why do you want the logic cut down?
It's coded so you can do Spell Description only/Components only/Spell Description + Components.

kakaroto:
> Why do you want the logic cut down?
It's coded so you can do Spell Description only/Components only/Spell Description + Components.

In all situations all you do is either **set** the value, or **append** to it, but you set or append based on what you expect it to be (either empty or has components in it). All I changed is that I make it always append, and if it was empty, I set it to an empty string before I append to it.
I did make a mistake (which I'll fix post-merge) the second line shouldn't have included `\n\n`

PR #477, opened 2020-07-29
Implement Halfling Luck Racial Feature

### LINKED ISSUE #475: Feature request: Halfling Luck
Having to reroll every time you roll a nat 1 with a halfling _can_ be automated. Roll20 uses the 1d20ro<1 command for practically all d20 rolls, including but not limited to death saves and initiative. I'd love to do this myself but I don't know how to code...

Aeristoka:
Fixes #475

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -78,20 +78,25 @@ class Beyond20RollRenderer {
         let advantage = request.advantage;
         if (advantage == RollType.QUERY)
             advantage = await this.queryAdvantage(title);
+        
+        let d20_roll = "1d20";
+        // Halfling Luck affects Attack Rolls
+        if (request.character["racial-traits"].includes("Lucky"))
+            d20_roll = d20_roll.replace(/[0-9]*d[0-9]+/g, "$&ro<=1");
 
         if ([RollType.DOUBLE, RollType.ADVANTAGE, RollType.DISADVANTAGE].includes(advantage)) {
-            const roll_1 = this.createRoll("1d20" + modifier, data);
-            const roll_2 = this.createRoll("1d20" + modifier, data);
+            const roll_1 = this.createRoll(d20_roll + modifier, data);
+            const roll_2 = this.createRoll(d20_roll + modifier, data);
 
             return {advantage, rolls: [roll_1, roll_2]};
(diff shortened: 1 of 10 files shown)


src/common/roll_renderer.js
```diff
@@ -78,20 +78,25 @@ class Beyond20RollRenderer {
         let advantage = request.advantage;
         if (advantage == RollType.QUERY)
             advantage = await this.queryAdvantage(title);
+        
+        let d20_roll = "1d20";
+        // Halfling Luck affects Attack Rolls
+        if (request.character["racial-traits"].includes("Lucky"))
+            d20_roll = d20_roll.replace(/[0-9]*d[0-9]+/g, "$&ro<=1");
```
kakaroto, line 85: You've just set it to `"1d20"` so not sure why you need to regexp it, I assume it's a paste from somewhere else ?

src/roll20/content-script.js
```diff
@@ -387,6 +406,8 @@ function rollAttack(request, custom_roll_dice = "") {
         if (request["critical-limit"])
             d20_roll = "1d20cs>" + request["critical-limit"];
         properties["mod"] = request["to-hit"] + format_plus_mod(custom_roll_dice);
+        if (request.character["racial-traits"].includes("Lucky"))
+            d20_roll += "r1";
```
kakaroto, line 410: now you're simply appending instead of using the same regexp again.

src/roll20/content-script.js
```diff
@@ -248,17 +248,21 @@ function rollSkill(request, custom_roll_dice = "") {
             prof_val += "+[[" + request.character.proficiency + " * 2]]";
             reliableTalent = request.reliableTalent;
         }
-        const d20 = reliableTalent ? "{1d20, 0d0 + 10}kh1" : "1d20";
+        const d20_roll = reliableTalent ? "{1d20, 0d0 + 10}kh1" : "1d20";
+        if (request.character["racial-traits"].includes("Lucky"))
+            d20_roll = d20_roll.replace(/1d20+/g, "$&r1");
```
kakaroto, line 253: not using the same regexp as above, but also you added to it `ro<=1` above (in roll renderer) and now you add `r1`.. I assume it's because `r1` is the specific roll20 syntax? the regexp wouldn't need to be different though, and `1d20+` means it would match `1d200000000` so the regexp is wrong

PR #480, opened 2020-08-08
Create bug_report.md

FunnyLamma:
This and a bit more code will build an issue template.


kakaroto:
Thanks for bringing that to me attention.
I finally got to look into this more closely. The "and a bit more code" threw me off, so I had to dig into it. Found that the file wasn't in the correct path (had to be in `.github/ISSUE_TEMPLATES/` and I used github's template generator to take care of it automatically.

Issue #160, opened 2020-04-01
Add support for Astral Tabletop


ThatOhio:
I've been using Astral Tabletop lately and would love to use this for that game. I will likely switch back to roll20 because of how useful this extension is but would love to use Astral.

https://astraltabletop.com/

I have a good amount of experience with C# and .Net, and a bit of front end experience. I saw your explanation in the card for Foundry, I don't have a ton of time at the moment to attempt the PR, but if I get some time before you can get to it, I will try.

kakaroto:
Moving this to the 2.2 milestone, considering PR #484 has been reviewed and is looking pretty good so far.

kakaroto:
Closing as #484 was merged!

PR #484, opened 2020-08-09
Basic Astral TableTop support

adriangaro:
Hey,

Recently I and some of my friends moved from roll20 to astral vtt, which sadly didn't have support in beyond20. As such, I thought that I may try to implement the support myself. 

Astral seems to be implemented in React which caused it to be almost impossible to interact/interfere with the UI or its logic; luckily most of the chat commands are post requests to their API and all the needed credentials to authenticate to it, were easily extracted from the window object, localStorage, and Indexeddb. 

At the moment, I have tried to implement all rolls with the exception of rendered rolls and combat updates. HP updates are also implemented with the caveat that tempHp is added to maxHp and currentHp as there is no way to display otherwise on the tokens. Considering the templates/actions in Astral can't check for crits, as such, it adds delayed rolls for the crits, and in the case of HOMEBREW_DOUBLE, the delayed rolls are not displayed at all; and the modifier on the damage rolls isn't added to the roll so the player/dm can't easily double the damage die.

There are also some improvements that I added to the description parsing since I have observed some patterns in them which could be used to recreate the tables, bolded sections in the description, and bullet lists. I don't know id they can be ported to the roll20 implementation as I am not very knowledgable with their templates.

If there are changes/requests required for this implementation in the future, I am glad to provide help if I can.


src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
...
+`;
+    })
+}
+
+function formatNotesInDescription(description) {
+    return description.replace(/Notes: [^\n]*/g, (match) => `_${match}_`);
+}
+
+function formatBoldedSectionInDescription(description) {
+    return description.replace(/([^\n\.])*\.\&nbsp\;/g, match => `\n**${match}**`)
```
kakaroto, line 135: If I remember correctly, `&nbsp;` have been replaced by a space in the descriptions in a recent version, so this wouldn't match anything I expect. Also unsure if this would work, it might match for some monster blocks, but looking for a `. ` is going to match a lot of stuff and you'll end up with all sorts of things being bolded randomly, which will annoy users.

src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
...
+function formatTableInDescription(description) {
+    const matchTable =  new RegExp(/(^\n\n([^\n\r]+)\n([^\n\r]+)\n\n\n\n$)(^\n([^\n\r]+)\n([^\n\r]+)\n\n$)+/gm);
+    const matchHead =  new RegExp(/(^\n\n([^\n\r]+)\n([^\n\r]+)\n\n\n\n$)/gm);
+    const matchRow =  new RegExp(/(^\n([^\n\r]+)\n([^\n\r]+)\n\n$)/gm);
+
+    return description.replace(matchTable, (match) => {
+        let [_1, _2, firstHead, secondHead] = match.matchAll(matchHead).next().value;
+        
+        // Skip first as it is matched by the matchHead too
+        const [_, ...rows] = [...match.matchAll(matchRow)].map(([_1, _2, firstRow, secondRow]) => {
```
kakaroto, line 117: I really hate the ugliness of using spread operators everywhere. Please prioritize readability and maintainability vs one-liners.
```js
const allMatches = reMatchAll(match, matchRow];
// Skip first as it is matched by the matchHead too
const rows = allMatch.map(match => `| ${match[2]} | ${match[3]} |`).slice(1);
```

src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
...
+            reliableTalent = request.reliableTalent;
+        }
+        const d20 = reliableTalent ? "1d20min10" : "1d20";
+        return {
+            icon: "bunker-soldier-standing",
+            color: "#7ED321",
+            title: request.skill + " Skill Check",
+            message: template([
+                ...advantageRoll(request, "Skill Check", `!(${d20}${prof_val.length > 0 ? " + " + prof_val : ""}${custom_roll_dice.length > 0 ? " + " + custom_roll_dice : ""})`)
+            ])
```
kakaroto, line 190: why spread into a list here? I would also suggest defining the roll in a separate variable so it's not one huge line.

```js
          const roll = `!(${d20}${prof_val.length > 0 ? " + " + prof_val : ""}${custom_roll_dice.length > 0 ? " + " + custom_roll_dice : ""})`;
 ....
            message: template(advantageRoll(request, "Skill Check", roll))
```

src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
...
+}
+
+function rollInitiative(request, custom_roll_dice = "") {
+    return {
+        icon: "arrow-down-9",
+        color: "#7ED321",
+        title: "Initiative",
+        message:  template([
+            ["Initiative", `i!(1d20${request.initiative}${custom_roll_dice.length > 0 ? " + " + custom_roll_dice : ""})`]
+        ])
```
kakaroto, line 239: Does astral have a "add to tracker" sort of functionality ?

src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
...
+    }
+}
+
+function rollAbility(request, custom_roll_dice = "") {
+    return {
+        icon: "coin-star",
+        color: "#4A90E2",
+        title: request.ability + " Ability Check",
+        message:  template([
+            ...advantageRoll(request, "Ability Check", `!(1d20${request.modifier}${custom_roll_dice.length > 0 ? " + " + custom_roll_dice : ""})`)
```
kakaroto, line 216: I see a lot of duplication here, I would suggest you make a function to generate the roll.
In some places you check that the modifier has length > 0, in others, you don't, you always do that code for the custom dice roll as well. You also add ` + ` in front of the custom roll dice, but you don't check if it's something that doesn't need it. For example a custom modifier of "- 2" or "+ 1"would cause it to become `1d20 + - 2` or `1d20 + + 1` which would both fail (I just tested it). So you'd need to trim the value, check if it starts with `+` or `- and add the `+` only if it doesn't.
Check either roll20 or the roll renderer code, I know I do that somewhere already.
As this becomes complicated, and a fix here would need to be copied in every other function, that's why having it separate to generate the roll would make sense.
Same comment from above about setting the variable elsewhere and not using the spread operator as it's not necessary, applies here and to every other similar line.

src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
...
+    const source = request["item-type"].trim().toLowerCase();
+    if ((source === "tool, common" || (source === "gear, common" && request.name.endsWith("Tools"))) && request.character.abilities && request.character.abilities.length > 0) {
+        const ret = rollTrait(request);
+
+        return {
+            ...ret,
+            message: ret.message + '\n' + template([
+                ...advantageRoll(request, request.name, `!(1d20${custom_roll_dice.length > 0 ? " + " + custom_roll_dice : ""})`)
+            ])
+        }
```
kakaroto, line 275:
```js
 ret.message += "\n" + template(advantageRoll(request, request.name, genRoll(custom_roll_dice)))
 return ret;
```
is easier to read

src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
...
+            ...advantageRoll(request, "To Hit", `!(1d20cr>=${request["critical-limit"] || 20}${request["to-hit"]}${custom_roll_dice.length > 0 ? " + " + custom_roll_dice : ""})`),
+        ]
+    }
+    
+    if (request["save-dc"] !== undefined) {
+        rolls = [
+            ...rolls,
+            ["Save DC", `DC ${request["save-dc"]} ${request["save-ability"]}`],
+        ]
+    }
```
kakaroto, line 316: Please use `rolls.push()`

src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
...
+function rollSpellCard(request) {
+    let rolls = [
+        ["Level", request["cast-at"] !== undefined ? `${request["level-school"]} (Cast at ${request["cast-at"]} Level)` :  request["level-school"]],
+        ["Components", request.components],
+        ["Range", request.range],
+        ["Casting Time", request["casting-time"]],
+        ["Duration", request.duration],
+        ...(request.ritual ? [["Ritual", "Yes"]] : []),
+        ...(request.concentration ? [["Concentration", "Yes"]] : []),
+    ];
```
kakaroto, line 354: Same, use `if (request.ritual) rolls.push(["Ritual", "Yes"])` as a separate line.

src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
...
+                let idx = damage_types.findIndex(t => t === dmgtype);
+                chromatic_damage = damages.splice(idx, 1)[0];
+                damage_types.splice(idx, 1);
+                idx = critical_damage_types.findIndex(t => t === dmgtype);
+                if (idx >= 0) {
+                    crit_damage = critical_damages.splice(idx, 1)[0];
+                    critical_damage_types.splice(idx, 1)[0];
+                }
+            }
+            damages.splice(0, 0, chromatic_damage + "m2");
```
kakaroto, line 420: + "m2" ? Why are you adding that flag to the damage dice ? I couldn't see what it did.
Do you have a link to where I can see what the roll syntax is ?

kakaroto (changes requested):
Hi @adriangaro!
Thank you very much for doing this Pull Request (Which fixes long standard feature request #160) and also thank you very much for your patience while I tried to find the time to give this the time it deserved.

I've had a look at the code, and in general, it looks good. There's a few small issues that I've pointed out in the review, which I hope you can fix, and there's a big difference in coding style as well. You seem to love the spread operator while I hate it (not really, but I think it needs to be used only when it's the appropriate tool for the task).

There are of course some missing things, such as the roll renderer (i.e: anything using the digital dice), and queries (for advantage, whispers), I think whisper mode HIDE_NAME for monsters is not supported either, and a few other small things, but that's ok, I would still merge this as a first iteration even if it doesn't do everything right from the start.
I would like however to see the things I've mentioned in the comments resolved first as those are either bugs, issues with logic, or coding style inconsistencies.

Thanks again!

src/avtt/content-script.js
kakaroto: Found this : https://help.astraltabletop.com/dice/syntax/flags/
Looks like it's "match group" so it groups similar dice results together when you roll multiple dice.
In this case, I don't see why it's needed for the chromatic orb spell damage

src/avtt/page-script.js
```diff
@@ -0,0 +1,16 @@
...
+    const meta = document.createElement("meta");
+    meta.innerText = window.__ASTRAL_USER.ready ? window.__ASTRAL_USER.user : window.__ASTRAL_USER;
+    meta.id = "BEYOND20_ASTRAL_USER"
+    document.head.append(meta);
+
+    
+    const meta2 = document.createElement("meta");
+    meta2.innerText = JSON.stringify(__NEXT_DATA__);
+    meta2.id = "BEYOND20_NEXT_DATA"
+    document.head.append(meta2);
```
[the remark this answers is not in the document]
adriangaro: Yeah, I don't like it either, but I don't think they expose any game objects/api which could be used. Probably most of the game logic is included in the react app, which would be unavailable due to webpack. It seems that they store the state globally for one reason or another. If I find an API I will probably change the implementation.

src/extension/background.js
```diff
@@ -4,6 +4,7 @@
 
 var settings = getDefaultSettings()
 var fvtt_tabs = []
+var avtt_tabs = []
```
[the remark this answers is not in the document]
adriangaro: Thanks for the heads up, will change it similarly to roll20.

src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
+async function postChatMessage({message, color, icon, title, whisper}) {
+    const user = getUser().uid;
+    const room = getRoom();
+    const token = await getAccessToken();
+    const character = getCharacter();
```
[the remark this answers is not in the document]
adriangaro: I made it this way since I have no idea what is the expiry time of an firebase access token and how often it is refreshed. This will at least make sure the token is not expired.

src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
...
+}
+
+function damagesToRolls(damages, damage_types, crits, crit_types) {
+    return [
+        ...damages.map((value, index) => [damage_types[index], subDamageRolls(value)]),
+        ...(settings['critical-homebrew'] == CriticalRules.HOMEBREW_DOUBLE ? 
+            [] : 
+            crits.map((value, index) => ["Crit " + crit_types[index], subDamageRolls(value)])
+        )
+    ];
```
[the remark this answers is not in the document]
adriangaro: Yeah, I have started removing the `CriticalRules.HOMEBREW_DOUBLE` as they initially were requested by my DM if I could implement them. It should be more readable now (well when I push the first draft of changes).

src/avtt/content-script.js
```diff
@@ -0,0 +1,580 @@
...
+    const replaceCB = (dice, modifier) => {
+        return dice == "" ? modifier : `${dice}${modifier} (!!(${dice}${modifier}))`;
+    }
+    return subRolls(description, replaceCB);
+}
+
+function subDamageRolls(text) {
+    return subRolls(text, (dice, modifier) => {
+        const prefix = settings["auto-roll-damage"] ? "!(" : "!!(";
+        modifier = settings['critical-homebrew'] == CriticalRules.HOMEBREW_DOUBLE ? `)${modifier}` : `${modifier})`;
```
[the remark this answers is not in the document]
adriangaro: Hmm right, this was specifically made like this due to the fact that I can't really check for the state of a crit or not in Astral, so I simply separated the modifier from the roll to easily double the damage. To be fair, I wasn't happy either with the hackish implementation. Now with it removed, I hope the logic is better.

src/avtt/content-script.js
adriangaro: The spread here I think may be a leftover from usually expecting after the advantage roll to add damage or other rows. Changed it to the recomandation.

src/avtt/content-script.js
adriangaro: From what I read in their roll reference, it seems so by prefixing the roll with an `i`. From how we used it till now, it seems to work for the GM.

src/avtt/content-script.js
adriangaro: It should technically light up the damage if the dice match for an easier indication if the attack bounces again. Looks like I put it on chromatic org instead in chaos bolt. Either way tested it a bit and no highlighting appears, so I'll remove it.

adriangaro:
Hey, I have implemented most of the requested changes. If there are any more changes to be done, just comment on which and I'll get on them.

Regarding the spread operator abuse, may be sourced to my job as we work with React and we are encouraged to always create new objects or lists, for the purpose of not mutating state. This is not applicable for this codebase/use case and I am sorry 😜 .

Regarding some of the missing features, I didn't have time to implement the rendered rolls, but I plan to work on them the next few days. In the case of queries, I don't know if they are able to be implemented in Astral as I haven't seen anything related to this in their references, so it might not be possible at the moment.

A couple of caveats to the refactor. I analyzed the `__NEXT_DATA__` variable in astral and it seems to be used for the initial render of the app. As such it is not updated when the area is changed or when characters are added or removed. There is also no easy way of seeing if characters have been added to the board, without either websocket or request interception, which I think is not desirable as users might not agree to that. So the only way of getting all available characters in a campaign is to query the game endpoint and extract the characters from there. This also means that every time we need to update a character or send a message/roll from a character we need to do the request again for the character list to check if any character has the corresponding name.

Lastly, for the purpose of not querying each time the access token from indexed, a cache was implemented which will automatically refresh the firebase access token if possible. To be able to do this, jwt_decode was added as a dependency for Astral. Hopefully, I inserted the license correctly for it, as to not cause problems for you in the future.

Haev a nice day, and thanks for your time for the review!

kakaroto:
Hi!
Thanks for implementing the changes so quickly!
I won't have time to review this for the next few days, but I just wanted to let you know that while reviewing it last week, I thought I'd check if we can do the same thing for Astral as we do for Roll20. I.e: fill in the chat textarea and click the "send" button to send the message without having to mess with firebase, tokens, and API endpoints, etc..
Turns out I discovered two things : 
1 - The chat area is within an iframe, so having to actually hook Beyond20 to `/widget` instead of `/app` was necessary
2 - while I can set the textarea content, there's no send button and trying to trigger keydown/keypress events didn't work, apparently due to React actually looking for some fake `React.KeyboardEvent<>` events instead of regular keypress Events.

I contacted the Astral developer and not only did they give official approval for Beyond20 to support Astral, but they also were open to making changes to help that task happen. I was just told that they've already made a change by adding a hidden button that can be clicked to send the text, I couldn't find it though, so I poked them about that, waiting for response.
I've found this code on stackoverflow which allows setting the textarea's content in a React compatible way :
```js
function setNativeValue(element, value) {
  const valueSetter = Object.getOwnPropertyDescriptor(element, 'value').set;
  const prototype = Object.getPrototypeOf(element);
  const prototypeValueSetter = Object.getOwnPropertyDescriptor(prototype, 'value').set;

  if (valueSetter && valueSetter !== prototypeValueSetter) {
    prototypeValueSetter.call(element, value);
  } else {
    valueSetter.call(element, value);
  }
  element.dispatchEvent(new Event('input', { bubbles: true }));
}
```
But I still couldn't get it to recognize the Enter keydown event. Hopefully with Astral team on board to provide us with an API, that can make things much simpler. I've contacted them to say that I can't find the new hidden button (maybe it's not yet on production servers).
Can you come up with a list of APIs that we might also need? If there's no way to query a user from a roll formula, then it can be done with the existing roll renderer alertify code, I think thought hat most queries will be done on the D&D Beyond side in the future (planning on moving queries from roll20 to dndb too).
I'm thinking the APIs we need are : 
- send chat message
- list of all actors (or get actor by name)
- set actor HP

I don't know if all tokens must be associated with an actor in Astral, but might need a "list all tokens in scene" API as well. I can push that to the devs to provide us with that.
In the meantime (depending on how fast they implement APIs to help Beyond20 and how fast you can implement those changes), I think I'm ok with just moving forward with the current implementation that uses the API endpoints with the auth token as a temporary thing.
Thanks!

adriangaro:
From what I have seen regarding the API endpoints that they have, there is `/game/<game-id>/area/<area-id>/character/` which returns all tokens/characters in an area. Sadly I couldn't use it due to the `area-id` not being exposed anywhere, and no events to which I could hook to.

Maybe if they were able to provide an API similarly to roll20 by providing a global object which may hold the game data, which will be kept up to date by the React app, we could easily discard all API endpoint calls and not need all the complication with auth tokens as you said.

I haven't checked how their initiative tracker works, but I suspect we may need to do something similar to how we handle others by doing requests (although this might as well be handled with websockets, in which case tough luck) for the purpose of syncing the initiative tracked with encounter tracker in beyond.

kakaroto:
Yeah, when I said "they provide an API", I meant as a global object with public methods (the API) we can use directly without using the HTTP API endpoints.
I just gave it a quick test by the way, I really like the message output format with the color/icon.
- I tried to post a Spellcasting chat, it silently errored, the API says error 400 `{"code":400,"error":"This message was longer than the maximum allowed length","message":"This message was longer than the maximum allowed length","title":"Message too long"}`
- I tried "Rage" and noticed that each paragraph is shown as a bullet point (same with pretty much every other feature that has more than one paragraph).
- I realized that the crits are always sent, I guess I should have expected it, I think you said so at some point.
- Setting roll type to "always ask" made it crash instead of defaulting to normal roll.
- The biggest issue was that nothing worked at first until I realized I had to disable digital dice (due to rendered rolls not being handled).

My current proposition would be : 
- Add rendered roll support (because most people will have digital dice enabled)
- Remove HP syncing entirely (as it's not particularly accurate on the actor selection and will cause problems with DMs having multiple windows open)
- Do more testing of most corner cases to improve stability and make sure it works correctly

By next week hopefully, we'll have a way to send chat messages via the chat widget directly, and we can drop the whole NEXT_DATA meta and token/fetch/etc..
For the next release, the implementation can be improved to make use of the (hopefully) upcoming API object for getting the actors and do HP syncing and all the other stuff. I expect that part to take longer for Astral to implement, which is why I'm putting it for 2.3 release.

I'm hoping to get a Beyond 20 release out this week. I might delay until next week if the chat message thing is coming soon (I was told updates are dropped on Fridays), or I might just get submerged with work again and it will delay itself on its own.
Regardless, what do you think ?

kakaroto:
Btw, I was just told that the domain for beta users is beta.astraltabletop.com instead of app.astraltabletop.com so some users in the Astral discord weren't able to test your repo as they were enrolled in beta. So I think your initial use of `*.astraltabletop.com` as the matching URL was the correct one to use (in the manifests and constants.js).

adriangaro:
I'll try to fix the observed issues either today or tomorrow, and in the weekend I'll check the chat iframe variant to see if the hidden button was added. 

Rendered rolls will be the main focus since they caused some issues with my players too since they tried rolling the digital dice and got nothing in astral. 
For HP syncing, any input on how are they inconsistent I tried a test campaign with multiple characters as GM and it worked for me.
I'll check the trait formatting since maybe some of the initial assumption of formatting are not valid anymore and may need to be fixed/removed entirely.
The silent chat error in this case is expected, looks like Astral has a 4096 or so character limit, in which case I don't know how it should be handled. Maybe splitting the message into multiple ones if possible? From what I know I encountered this on spell cards and items.

adriangaro:
Probably the first thing to do would be to release a version which allows the beta version to use it, so we may get more feedback.

kakaroto:
The chat button is now in the production app! 🥳 
And I just tested, and it worked.
```js
function setReactElementValue(element, value) {
  const valueSetter = Object.getOwnPropertyDescriptor(element, 'value').set;
  const prototype = Object.getPrototypeOf(element);
  const prototypeValueSetter = Object.getOwnPropertyDescriptor(prototype, 'value').set;

  if (valueSetter && valueSetter !== prototypeValueSetter) {
    prototypeValueSetter.call(element, value);
  } else {
    valueSetter.call(element, value);
  }
  element.dispatchEvent(new Event('input', { bubbles: true }));
}
const chatIframe = document.querySelector("div[data-id=chat] iframe");
const textarea = chatIframe.contentDocument.querySelector("textarea")
const button= textarea.closest("form").querySelector("button[type=submit]")
setReactElementValue(textarea, "Beyond20 is awesome!")
button.click();
```

I'll answer the rest of your questions later 👋

kakaroto:
So, a user who tested the extension reported this error to me : 
`Uncaught (in promise) TypeError: Cannot read property 'accessToken' of undefined`
It looks like the `_getTokenFromDb()` was returning `undefined` for them, which is why it wasn't working.
I think switching to using the chat button directly is definitely a safer approach.


> For HP syncing, any input on how are they inconsistent I tried a test campaign with multiple characters as GM and it worked for me.

I hadn't noticed that you changed the code for setting the HP to get the character by name. I thought you said you couldn't do it without it being complicated, but I guess it was just that it would be heavy on doing needless requests. In that case, it's fine, just make sure to add checks again errors like above, where there's no accessToken or it fails to do things properly.
Also, don't forget to send the data via an event to the content script instead of placing it in a meta tag.

> The silent chat error in this case is expected, looks like Astral has a 4096 or so character limit, in which case I don't know how it should be handled. Maybe splitting the message into multiple ones if possible? From what I know I encountered this on spell cards and items.

If the limit is known and is 4k, then it should just crop at 4093 characters and add `...` to it. I think that makes the most sense to me. Using the chat area directly though, you probably won't need to do that, but it would be good to test how it reacts to huge data being set in the textarea.

adriangaro:
I'll take a look over the beta too, if possible is it accessible to the public?

We might use the chat area, but for now idk how worth it is considering we can't speak as a user/character without using the chat endpoint. Similarly, I don't think we can add title, color, and icon without calling the endpoint.

kakaroto:
I don't know about beta, might be able to request it to the devs from their discord. If the manifest says `*.` instead of `app.` for the domain to match, that should be all that's needed.

Oh, I hadn't thought of the color/icon/etc.. and speak as... I thought that was stuff available from the chat interface... yeah, if that's not doable... 😬 
How about this : 
- Use API if possible, if it fails for whatever reason, fallback on the chat button

Also thought of this: for the HP syncing, keep a cache of the characters so it's not one API request every time. If the set fails (the token/actor was deleted), or it doesn't exist (might be new?) then update the cache and try again.

adriangaro:
I have implemented rendered rolls. They work, although not as good as in roll20. Astral has no way currently to create pre-rendered rolls or create tooltips using markup. As such only the total is displayed and if it is a crit or fumble. The formula is not visible, neither are the parts. We could show them outside of the roll, but I think it would clutter the messages.

manifest.json
```diff
@@ -36,8 +37,10 @@
 	"web_accessible_resources": [
         "images/*",
         "libs/alertify.min.js",
+        "libs/jquery-3.4.1.min.js",
```
kakaroto, line 40: humm.. I never noticed I had a mix of tabs and spaces here, I'll need to fix that!

manifest_ff.json
```diff
@@ -43,6 +44,7 @@
         "libs/alertify.min.js",
```
kakaroto, line 44: Forgot to add jquery as web accessible for FF

src/astral/data-utils.js
```diff
@@ -0,0 +1,86 @@
...
+
+const getAccessToken = async () => {
+    if (!tokenCache.expires) {
+        const tokenData = await _getTokenFromDb();
+
+        tokenCache.access = tokenData.accessToken;
+        tokenCache.apiKey = tokenData.apiKey;
+        tokenCache.expires = tokenData.expirationTime;
+        tokenCache.refresh = tokenData.refreshToken;
+    } else if (tokenCache.expires < Date.now()) {
```
kakaroto, line 35: Consider it expired 1 minute before real expiration. Otherwise, if it will expire in 100ms, but it takes more than 100ms to send the request to the server (either based on latency or on clock skew between local and remote server clocks), then the request will fail.

src/astral/data-utils.js
```diff
@@ -0,0 +1,86 @@
...
+        const db = req.result;
+        const trans = db.transaction("firebaseLocalStorage", IDBTransaction.READ)
+        const objStore = trans.objectStore("firebaseLocalStorage");
+
+        const cursorReq = objStore.openCursor();
+
+        cursorReq.onsuccess = (event) => {
+            console.log("Create cursor req", cursorReq.result)
+            resolve(cursorReq.result.value.value.stsTokenManager);
+        };
```
kakaroto, line 21: Is there a onerror for cursorReq too ? and can you check if the `result.value.value.stsTokenManager` does exist ?

src/astral/data-utils.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    req.onerror = reject;
+});
+
+const getAccessToken = async () => {
+    if (!tokenCache.expires) {
+        const tokenData = await _getTokenFromDb();
```
kakaroto, line 29: Don't assume it will not return an error or a `null` value

src/astral/data-utils.js
```diff
@@ -0,0 +1,86 @@
...
+    } else if (tokenCache.expires < Date.now()) {
+        const tokenData = await (await fetch('https://securetoken.googleapis.com/v1/token?key=' + tokenCache.apiKey, {
+            method: 'POST',
+            headers: {'content-type': 'application/x-www-form-urlencoded'},
+            body: 'grant_type=refresh_token&refresh_token=' + tokenCache.refresh,
+        })).json();
+        
+        tokenCache.access = tokenData.access_token;
+        tokenCache.refresh = tokenData.refresh_token;
+        const tokenInfo = jwt_decode(tokenCache.access);
```
kakaroto, line 44: If this is all you need, you shouldn't have to depend on jwt_decode.
You can get the info like this `JSON.parse(atob(tokenCache.access.split(".")[1]))` in theory.. which is better/shorter than importing that an external dependency just to call `atob` and `JSON.parse`

src/astral/page-script.js
```diff
@@ -0,0 +1,200 @@
...
+        chatIframe = $("div[data-id=chat] iframe").contents();
+        
+        // Added this since previews links will not work and they open a new tab
+        chatIframe.on('click', `a[href*='#beyond20-rendered-roll-button-']`, (ev) => {
+            ev.preventDefault();
+        })
+    } else {
+        setTimeout(trySetDOMListeners, 1000);
+    }
+}
```
kakaroto, line 198: Move the global var and the function to the top of the file since `chatIframe` is used by the functions at the very top but doesn't get defined until the end, which was confusing during review. (the call to `trySetDOMListeners` should remain at the end of the file though)

src/astral/page-script.js
```diff
@@ -0,0 +1,200 @@
...
+                updateAt: Date.now(),
+                attributeBarMax: maxHp + tempHp,
+                attributeBarCurrent: hp + tempHp,
+
+            }
+        }),
+        headers: {
+            'x-authorization': `Bearer ${token}`, 'content-type': 'application/json'
+        }
+    });
```
kakaroto, line 171: Not sure if it would be required to check in case of error, clear the character cache and try again (i.e: the character was deleted). But I'm assuming here that characters are not tokens, and tokens are always links to characters on Astral? (sorry don't know how it works).
In which case, I'd say this is fine and clearing character cache would be overengineering it.
Up to you if you think it's a use case that would happen often (i.e: a character with a matching name getting deleted and recreated with a different id).

src/common/utils.js
```diff
@@ -55,7 +55,7 @@ function injectPageScript(url) {
     const s = document.createElement('script');
     s.src = url;
     s.charset = "UTF-8";
-    s.onload = () => s.remove();
+    // s.onload = () => s.remove();
```
kakaroto, line 58: Any reason to keep the tag? Honestly, I just copied that from the stackoverflow, it never caused any issues but I don't see a reason for the tag to be removed either.

kakaroto (approved):
Did a quick review again, it generally looked good, so I'll be merging it this weekend. If there's any fixes you want to get in from the comments I made, go ahead, otherwise I'll be testing it and doing any fixes myself at the same time as the merge.
Thanks again for doing this and being so responsive to all my comments!

It was really hard for me not to do an in-depth review of every single line and comment about anything I would have done differently, I'm trying to be less nitpicky in my reviews, but your code looks good in general anyway and I trust that you're doing it right! 👍 If there is some bug, or something I missed in review, that can simply be fixed for the 2.3 release!
Next step: Updating the website/docs to include Astral as a supported VTT! 🥳 
Thanks! :)

src/common/utils.js
adriangaro: Bot specifically, I think I just removed to check something and format to undo the change, will fix it tomorrow morning.

src/astral/data-utils.js
adriangaro: I'll take a look, to be fair I have seen the logic in jwt_decode to have a couple of fail safes and assumed they were needed, if in our case, that's all we need, I'll change it.

adriangaro:
I just got access to the Beta so I'll try to see tomorrow morning if it works with it and if there are any fixes needed for it.

src/astral/page-script.js
adriangaro: From what I observed tokens are always linked to characters on Astral. Either way the `getCharacters` method tries to look into the cache and if not found tries to refresh the cache from server. An edge case I see here may be if a character is deleted, but still available in cache, then trying to send a message as that character or updating it may error.

adriangaro:
I have added some basic support for the combat tracker. It works but I need to do a lot of requests to sync it up. For syncing, I need to clean the combat and recreate it due to not being able to set the turn of a specific id. But all in all, works pretty goog-ish.

kakaroto (commented):
Every time I review your changes, it feels like you've changed a whole lot of stuff and I don't recognize the code, lol. I think the PR's title of "basic astral support" is not so true anymore, especially with the combat tracker stuff added.
I failed miserably at working on B20 and making the release over the weekend. I'll try and do that over the next couple of days though. I don't have other pressing things, but I'm exhausted, so let's hope I manage it 👍 
Thanks!

kakaroto:
I fixed the two small issues I mentioned and did the merge 🥳

PR #499, opened 2020-08-24
Fix double character names in roll20 condition display

macmaxbh:
Right now, when a character condition changes for a player in a Roll20 campaign their character's name is displayed twice in the chat log. Was going to log this as an issue; but then decided to just track it down and fix it. Thanks for the great plugin; LMK if you'd prefer a different process / want me to log an issue as well. 

As per [Roll20 Text Chat](https://roll20.zendesk.com/hc/en-us/articles/360039675093-Text-Chat), /em does not require a character name, only /emas for GMs wants a character name (surrounded in quotes). Update the roll20 javascript to match this (and also clean up some spacing/grammar since I happened to be there). 

Tested this as both a player and GM in Roll20 Chrome; looks good to me. 

Also, because the build was updating the built dndbeyond_item.js based on one of the earlier commits (a4d09d0), check in that update as well.

### DIFF
--- a/dist/dndbeyond_item.js
+++ b/dist/dndbeyond_item.js
@@ -3280,7 +3280,7 @@ function addDisplayButton() {
     const item_name = $(".page-title").text().trim();
     const item_type = descriptionToString(".item-details .item-info .details, .details-container-equipment .details-container-content-description > div > .details-container-content-description-text");
     const description = descriptionToString(".item-details .more-info-content, .details-container-equipment .marginBottom20 + .details-container-content-description-text");
-    const item_tags = $(".details-container-content-footer .tags").find(".tag").toArray().map(elem => elem.textContent);
+    const item_tags = $(".details-container-content-footer .tags .tag").toArray().map(elem => elem.textContent);
     $(".page-heading__content").after(button);
 
     $(".ct-beyond20-roll").css({
--- a/dist/roll20.js
+++ b/dist/roll20.js
@@ -3509,12 +3509,12 @@ function handleMessage(request, sender, sendResponse) {
 
             // We can't use window.is_gm because it's  !available to the content script;
             const is_gm = $("#textchat .message.system").text().includes("The player link for this campaign is");
-            const em_command = is_gm ? "/emas " : "/em ";
+            const em_command = is_gm ? "/emas \"" + character_name + "\" " : "/em ";
             let message = "";
             if (conditions.length == 0) {
-                message = em_command + character_name + " has no active condition";
+                message = em_command + "has no active conditions";
             } else {
-                message = em_command + character_name + " is : " + conditions.join(", ");
+                message = em_command + "is: " + conditions.join(", ");
             }
             postChatMessage(message, character_name);
         }
--- a/dist/roll20.js
+++ b/dist/roll20.js
@@ -3509,12 +3509,12 @@ function handleMessage(request, sender, sendResponse) {
 
             // We can't use window.is_gm because it's  !available to the content script;
             const is_gm = $("#textchat .message.system").text().includes("The player link for this campaign is");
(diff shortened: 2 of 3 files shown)


Aeristoka:
You should only be editing code in the files in /src/, then building to change the Dist files.

macmaxbh:
> You should only be editing code in the files in /src/, then building to change the Dist files.

I figured that out after the first attempt ;). I included an unrelated dist change from a previous src change that didn't get into the dist copy (a4d09d0) so it didn't appear on each build for everyone else going forward.

Aeristoka:
If you're building, and committing properly, we should have your src/ changes, and the resulting dist/ changes as well in your PR.

macmaxbh:
> If you're building, and committing properly, we should have your src/ changes, and the resulting dist/ changes as well in your PR.

You do? Change includes three files...the dist change for a4d09d0 (dist/dndbeyond_item.js), the dist change for my change (dist/roll20.js), and the src change for my change (src/roll20/content-script.js)? Am I missing something?

Aeristoka:
https://github.com/kakaroto/Beyond20/pull/499/commits/e0bcb52bdba85280fd807f43f693d86635a81799
https://github.com/kakaroto/Beyond20/pull/499/commits/a4bf548a7992e36539fa114bbe45124496a1c87e
https://github.com/kakaroto/Beyond20/pull/499/commits/cc4d548e20b1927328a4e01d53ebcf0cecab3bf4
https://github.com/kakaroto/Beyond20/pull/499/commits/df35674e723e6d1528b6c49fb753f06c1dffe567
You ONLY changed a dist file, this is a no-no, and won't work anyway (you edit only the source files, and build, that will generate the dist files)

macmaxbh:
The source change is in cc4d548.

> [e0bcb52](https://github.com/kakaroto/Beyond20/commit/e0bcb52bdba85280fd807f43f693d86635a81799)
> [a4bf548](https://github.com/kakaroto/Beyond20/commit/a4bf548a7992e36539fa114bbe45124496a1c87e)
> [cc4d548](https://github.com/kakaroto/Beyond20/commit/cc4d548e20b1927328a4e01d53ebcf0cecab3bf4)
> [df35674](https://github.com/kakaroto/Beyond20/commit/df35674e723e6d1528b6c49fb753f06c1dffe567)
> You ONLY changed a dist file, this is a no-no, and won't work anyway (you edit only the source files, and build, that will generate the dist files)

The source change is in cc4d548.

Aeristoka:
Right, so the other changes to dist should have been abandoned, then the actual src/ with the accompanying dist/ change from building should be the only things included in the PR.

macmaxbh:
So I assumed you would squash merge so you didn't care about the intermediate changes since one undid the other so there was no net effect. I can recreate the change without those two changes if you would prefer. 

To confirm before I resubmit, are you also saying you do _not_ want the dist change for a4d09d0 in this PR even though it is automatically generated by the build system and was due to a previous change being incomplete?

Aeristoka:
No, the ONLY Dist change you should have included is the one that is a result of the build off the Source change.

And the later commits did not revert all of the earlier changes, as the "changed files" are still showing other Dist changes.

macmaxbh:
Only the changes for a4d09d0; which I deliberately included because I thought bringing the dist copy back in sync with the source would be the right thing to do; but happy to just cherry-pick my single core commit. Closing this, re-opening the tighter-scoped PR.

kakaroto:
> So I assumed you would squash merge so you didn't care about the intermediate changes since one undid the other so there was no net effect.

Yep, I would have squash merged, but there were other changes in dist that weren't in any src file either, so the squashed commits would still have had content that wouldn't make sense to have within that change and which would have been overridden on the next build.

I'll check/review #500, but FYI you didn't need to close/re-open, you could have pushed new commits they would have appeared here, or if you squashed manually, a `git push -f` would have worked as well.

PR #500, opened 2020-08-24
Fix double character names in roll20 condition display

macmaxbh:
Right now, when a character condition changes for a player in a Roll20 campaign their character's name is displayed twice in the chat log. Was going to log this as an issue; but then decided to just track it down and fix it. Thanks for the great plugin; LMK if you'd prefer a different process / want me to log an issue as well.

As per Roll20 Text Chat, /em does not require a character name, only /emas for GMs wants a character name (surrounded in quotes). Update the roll20 javascript to match this (and also clean up some spacing/grammar since I happened to be there).

Tested this as both a player and GM in Roll20 Chrome; looks good to me.

As per the discussion in #499; _not_ including the other dist change due to a past src change that did not include a matching dist update to scope this PR just to my change.

### DIFF
--- a/dist/roll20.js
+++ b/dist/roll20.js
@@ -3509,12 +3509,12 @@ function handleMessage(request, sender, sendResponse) {
 
             // We can't use window.is_gm because it's  !available to the content script;
             const is_gm = $("#textchat .message.system").text().includes("The player link for this campaign is");
-            const em_command = is_gm ? "/emas " : "/em ";
+            const em_command = is_gm ? "/emas \"" + character_name + "\" " : "/em ";
             let message = "";
             if (conditions.length == 0) {
-                message = em_command + character_name + " has no active condition";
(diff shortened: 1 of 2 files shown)


Aeristoka:
There we go, nice and clean 👍

kakaroto:
Thanks @macmaxbh for the PR. The name was kept on purpose when using /em because it's duplicated only if the character name matches the selected name. I think you're right though, if it's the same name, it makes it duplicate which can be a problem.
I'd need to check if the character name is available in the "Send message as" combobox (and does that affect `/em`?) and only add the name if it's not available.
I should have some time this week to work on Beyond20 so I'll check this more in depth and test. At first glance though, other than the above comment about using a character sheet that isn't our own, the commit looks good 👍

kakaroto:
Thanks, I've merged it and add the missing character name in case the character doesn't match one of the characters the player can speak as.

PR #527, opened 2020-10-02
Fix for DnDBeyond Skill Badge Changes

Aeristoka:
Simplifies and improves the Skill Badge Override Logic


kakaroto:
Such poor code, I should report you for spamming!

PR #548, opened 2020-10-07
Text Clarication for Features & Traits

Aeristoka:
Just a thought, might help the several people having issues with where to click for this.

Doesn't need to happen necessarily.


kakaroto:
Yeah agreed, was thinking the same, but you were faster! 👍

PR #551, opened 2020-10-08
Do not query for Colossus Slayer Critical Hit

### LINKED ISSUE #546: Colossus Slayer Edge Case Bug
Ranger: Hunter w/ Colossus Slayer
Critical hit rule: Homebrew: Perfect rolls

When asked on the pop-up if you should use Colossus Slayer, select "No" (the default)
image

The Perfect Crit for Colossus slayer is applied, though the base damage is not:
image

Aeristoka:
Fixes #546
Properly removes Critical Hit Damage and Critical Hit Damage Type (in addition to the already removed Damage/Damage Type).
This is necessary because damagesToCrits has already been called, so the Colossus Slayer Critical Hit Damage and Damage Type are already ready to go.

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -548,14 +548,26 @@ class Beyond20RollRenderer {
             }
 
             // Ranger Ability Support;
+            let dmg_dice = "0";
             for (let [dmgIndex, dmgType] of damage_types.entries()) {
                 if (dmgType == "Colossus Slayer") {
                     const dmg = damages[dmgIndex].toString();
                     if (dmg) {
-                        const dmg_dice = await this.queryGeneric(request.name, `Add ${dmgType} damage ?`, { "0": "No", [dmg]: "Yes" }, "dmg_dice", ["0", dmg]);
+                        dmg_dice = await this.queryGeneric(request.name, `Add ${dmgType} damage ?`, { "0": "No", [dmg]: "Yes" }, "dmg_dice", ["0", dmg]);
                         if (dmg_dice == "0") {
                             damages.splice(dmgIndex, 1);
                             damage_types.splice(dmgIndex, 1);
+                        }
+                    }
+                }
+            }
+
+            for (let [dmgIndex, dmgType] of critical_damage_types.entries()) {
+                if (dmgType == "Colossus Slayer") {
+                    const dmg = critical_damages[dmgIndex].toString();
+                    if (dmg) {
+                        // dmg_dice = await this.queryGeneric(request.name, `Add ${dmgType} damage ?`, { "0": "No", [dmg]: "Yes" }, "dmg_dice", ["0", dmg]);
+                        if (dmg_dice == "0") {
                             critical_damages.splice(dmgIndex, 1);
                             critical_damage_types.splice(dmgIndex, 1);
                         }
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -556,6 +556,8 @@ class Beyond20RollRenderer {
                         if (dmg_dice == "0") {
                             damages.splice(dmgIndex, 1);
                             damage_types.splice(dmgIndex, 1);
+                            critical_damages.splice(dmgIndex, 1);
+                            critical_damage_types.splice(dmgIndex, 1);
                         }
                     }
                 }
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -547,33 +547,6 @@ class Beyond20RollRenderer {
                 damages[0] = damages[0].replace("d8", ttd_dice);
             }
 
-            // Ranger Ability Support;
-            // let dmg_dice = "0";
-            // for (let [dmgIndex, dmgType] of damage_types.entries()) {
-            //     if (dmgType == "Colossus Slayer") {
-            //         const dmg = damages[dmgIndex].toString();
-            //         if (dmg) {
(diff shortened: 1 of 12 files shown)


kakaroto:
There's no guarantee that colossus slayer would be in the same index on the critical array (other than brutal, I think there are other cases where it adds damage only to crit), so would need to find it in the crit array and delete the correct index instead of reusing the same index.

Aeristoka:
> There's no guarantee that colossus slayer would be in the same index on the critical array (other than brutal, I think there are other cases where it adds damage only to crit), so would need to find it in the crit array and delete the correct index instead of reusing the same index.

That's technically correct, but under what circumstances are we pushing a damage or a damage type without the other?

kakaroto:
Like I said, brutal... imagine this : 
- damages: []
- crit_damages: []

Push piercing damage
- damages: [piercing]
- crit_damages: [piercing]

Push brutal critical
- damages: [piercing]
- crit_damages: [piercing, brutal]

Push colossus thing
- damages: [piercing, colossus]
- crit_damages: [piercing, brutal, colossus]

Indexes do not match anymore. I don't *know* if that's the exact use case, or if we're doing colossus before brutal or not, but I'd rather not be hit by this bug one year from now if that ever changes in some other unforeseen situation.

Aeristoka:
@kakaroto Ready for review again (sorry, forgot I needed to fix this one up)

src/common/roll_renderer.js
```diff
@@ -528,6 +529,18 @@ class Beyond20RollRenderer {
...
+                    const dmg = critical_damages[dmgIndex].toString();
+                    if (dmg) {
+                        if (dmg_dice == "0") {
+                            critical_damages.splice(dmgIndex, 1);
+                            critical_damage_types.splice(dmgIndex, 1);
+                        }
+                    }
+                }
+            }
+
```
kakaroto, line 543: The `if (dmg_dice =="0")` should be before the for loop, otherwise you're doing the for loop for nothing. Also, I don't know why you're not just doing the check/change inside the original `if (dmg_dice == "0")`, why have a separate loop for crits...
Also, I'm realizing.. why do we loop over the whole list to find Colossus Slayer instead of doing this : 
```js
const collossusSlayerIdx = damage_types.findIndex(d => d === "Colossus Slayer");
if (colossusSlayerIdx !== -1) {
  const dmg = damages[collossusSlayerIdx].toString();
  if (dmg) {
    const dmg_dice = await this.queryGeneric(request.name, `Add ${dmgType} damage ?`, { "0": "No", [dmg]: "Yes" }, "dmg_dice", ["0", dmg]);
   if (dmg_dice == "0") {
      const critIdx = critical_damage_types.findIndex(d => d === "Colossus Slayer");
      damages.splice(dmgIndex, 1);
      damage_types.splice(dmgIndex, 1);
      if (critIdx !== -1) {
        damages.splice(dmgIndex, 1);
        damage_types.splice(dmgIndex, 1);
     }
    }
  }
}

**Comment by kakaroto** [comment]
Humm.. I thought we said we'd move the queries to the dndb side, but don't see that in this conversation... 
Anyway, see comment above which rewrites the whole section in a cleaner/simpler solution. Please apply it, test it, confirm and push it so I can merge, taking the credit for that change :P

**Comment by Aeristoka** [comment]
> Humm.. I thought we said we'd move the queries to the dndb side, but don't see that in this conversation...
> Anyway, see comment above which rewrites the whole section in a cleaner/simpler solution. Please apply it, test it, confirm and push it so I can merge, taking the credit for that change :P

Moved the query to the DnDB side, let me know what you think about it now. If you like it this way, I'll scrub the commented code out.

**Comment by kakaroto** [comment]
Yeah, it's so much simpler. 
I preferred the `Is the enemy you're targeting below its Hit Point Maximum?` question, and I thought we should just go ahead an use the roll renderer query instead of `confirm. I don't like the discrepancy. And yes, I know it makes it async, but I think query advantage and whisper are already making it async so... 🤷‍♂️ Just do that and create an issue about checking for async issues  so I can look at it later and make sure it's consistent and not doing anything wrong

**Comment by Aeristoka** [comment]
Changes made as requested, Issue #572 opened for you to verify modifying "rollItem" to "async" @kakaroto

**Comment by kakaroto** [comment]
the diff looks quite wrong, it just removes some code.

**Comment by Aeristoka** [comment]
> the diff looks quite wrong, it just removes some code.

That's a result of the last commit:
-----------------------------------------------------------------------------------------
Change to Colossus Slayer Behavior

Brings it in-line with things like Sneak Attack
Ie, DM/Player needs to know when to apply, if they don't, their problem
-----------------------------------------------------------------------------------------

It essentially strips down all the crazy query code, so it's like Sneak Attack now.

File: src/dndbeyond/content-scripts/character.js [hunk 2]
```diff
@@ -204,7 +204,7 @@ function rollHitDie(multiclass, index) {
     });
 }
 
-function rollItem(force_display = false, force_to_hit_only = false, force_damages_only = false, spell_group = null) {
+async function rollItem(force_display = false, force_to_hit_only = false, force_damages_only = false) {
```
kakaroto, line 207: your fix to the merge conflict here is wrong, removing the `spell_group` argument.
Also, not sure why changing the function to async is needed anymore

kakaroto:
> It essentially strips down all the crazy query code, so it's like Sneak Attack now. It's doing what I asked it to.

Ah ok, so it's not "bugfix" but rather "don't query anymore for CS and apply it all the time"

PR #567, opened 2020-10-19
Implement Fighting Initiate - GWF Feat && Champion Fighter "Additional Fighting Style"

Aeristoka:
Behaves the same as Great Weapon Fighting Style

### DIFF
--- a/src/dndbeyond/base/character.js
+++ b/src/dndbeyond/base/character.js
@@ -343,17 +343,21 @@ class Character extends CharacterBase {
         }
     }
 
-    hasClassFeature(name) {
-        return this._class_features.includes(name);
+    hasClassFeature(name, substring=false) {
+        if (substring) return this._class_features.some(f => f.includes(name));
+        else return this._class_features.includes(name);
     }
-    hasRacialTrait(name) {
-        return this._racial_traits.includes(name);
+    hasRacialTrait(name, substring=false) {
+        if (substring) return this._racial_traits.some(f => f.includes(name));
+        else return this._racial_traits.includes(name);
     }
-    hasFeat(name) {
-        return this._feats.includes(name);
+    hasFeat(name, substring=false) {
+        if (substring) return this._feats.some(f => f.includes(name));
+        else return this._feats.includes(name);
     }
-    hasAction(name) {
-        return this._actions.includes(name);
+    hasAction(name, substring=false) {
+        if (substring) return this._actions.some(f => f.includes(name));
+        else return this._actions.includes(name);
     }
     getClassLevel(name) {
         return this._classes[name] || 0;
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -194,7 +194,7 @@ function buildAttackRoll(character, attack_source, name, description, properties
                 if (highest_dice != 0) {
                     let brutal_dmg = `${brutal}d${highest_dice}`
                     // Apply great weapon fighting to brutal damage dice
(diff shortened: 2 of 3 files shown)


kakaroto:
Wondering if the 'hasClassFeature' shouldn't just look for the string (an 'includes' instead of '==') and use "Great Weapon Fighting" instead of "Fighting styel.." so it matches all 3.
I've been thinking of something like that as well for the polearm claws/natural weapons thing as well (for that one, I'd do a 'isNaturalWeapon' probably) as, a way to simplify the long if statements.
What do you think ?

Aeristoka:
> Wondering if the 'hasClassFeature' shouldn't just look for the string (an 'includes' instead of '==') and use "Great Weapon Fighting" instead of "Fighting styel.." so it matches all 3.
> I've been thinking of something like that as well for the polearm claws/natural weapons thing as well (for that one, I'd do a 'isNaturalWeapon' probably) as, a way to simplify the long if statements.
> What do you think ?

Based on how they keep adding things in like this, yeah I think that would be better.

Aeristoka:
> Wondering if the 'hasClassFeature' shouldn't just look for the string (an 'includes' instead of '==') and use "Great Weapon Fighting" instead of "Fighting styel.." so it matches all 3.
> I've been thinking of something like that as well for the polearm claws/natural weapons thing as well (for that one, I'd do a 'isNaturalWeapon' probably) as, a way to simplify the long if statements.
> What do you think ?

Should also be done for the 'hasFeat' I think, to simplify things out. In this example, I bet this drops the ' (UA)' at the end, and as it is now, code has to be changed when that changes. If it's using an 'includes' the code can just sit.

kakaroto:
Is the UA still required here or did this change with Tasha's release ?

Aeristoka:
> Is the UA still required here or did this change with Tasha's release ?

It looks like DnDBeyond has this feat screwed up now... You can select the feat, but CAN'T choose the Fighting Style...

Aeristoka:
> Is the UA still required here or did this change with Tasha's release ?

This is fixed now.

Aeristoka:
> Wondering if the 'hasClassFeature' shouldn't just look for the string (an 'includes' instead of '==') and use "Great Weapon Fighting" instead of "Fighting styel.." so it matches all 3.
> I've been thinking of something like that as well for the polearm claws/natural weapons thing as well (for that one, I'd do a 'isNaturalWeapon' probably) as, a way to simplify the long if statements.
> What do you think ?

`hasClassFeature(name) {
        return this._class_features.includes(name);
    }
    hasRacialTrait(name) {
        return this._racial_traits.includes(name);
    }
    hasFeat(name) {
        return this._feats.includes(name);`

Looks like the relevant code for this, not sure how to get a substring of an array.includes() call...

src/dndbeyond/base/character.js
```diff
@@ -343,17 +343,21 @@ class Character extends CharacterBase {
...
     }
-    hasFeat(name) {
-        return this._feats.includes(name);
+    hasFeat(name, substring=false) {
+        if (substring) return this._feats.some(f => f.includes(name));
+        else return this._feats.includes(name);
     }
     hasAction(name) {
-        return this._actions.includes(name);
+        if (substring) return this.actions.some(f => f.includes(name));
```
kakaroto, line 359: really, dude? really?
Can't even use the right variable name.
Much disappointed.

Aeristoka: Fixed, and squashed and everything.

PR #570, opened 2020-10-24
Notify about features & traits instead of blocking the user with a modal dialog.

Kvalyr:
Modal dialogs create a frustrating UX and should only be used to control what a user *must* do, not what they *should* do. 

This modal dialog interrupts players from just using DNDBeyond regardless of whether or not they're actively playing with a VTT. A user just viewing a character sheet for a recently created/copied/leveled character doesn't necessarily care that Beyond20 can't parse all its information; and shouldn't be interrupted for that.

Caveat: I'm not a JS coder, I just took a cursory look at the alertify docs to see what else it can do and offered this change for the sake of putting my money where my mouth is, so to speak.

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -1840,7 +1840,7 @@ function documentModified(mutations, observer) {
     activateQuickRolls();
     if (character._features_needs_refresh && !character._features_refresh_warning_displayed) {
         character._features_refresh_warning_displayed = true;
-        alertify.alert("This is a new or recently leveled-up character sheet and Beyond20 needs to parse its information. <br/>Please select the <strong>'Features &amp; Traits'</strong> panel on your DnDBeyond Character Sheet for Beyond20 to parse this character's features and populate the character-specific options.");
+        alertify.notify("This is a new or recently leveled-up character sheet and Beyond20 needs to parse its information. <br/>Please select the <strong>'Features &amp; Traits'</strong> panel on your DnDBeyond Character Sheet for Beyond20 to parse this character's features and populate the character-specific options.");
     }
 
     const pane = $(".ct-sidebar__pane-content > div");


Aeristoka:
Please make sure to run the proper build command on your changes, and do a new commit.

kakaroto:
I hesitated to make it a notification instead of a dialog because a lot of the time, users don't see the notification as it will tend to auto hide after some time and you sometimes open a page and go do something else while it loads.
The alert only happens once though (in theory) or when you level up, so it's not like it's particularly disruptive.
What do you think @Aeristoka ?

Aeristoka:
I'm not adverse to it being a modal, as it is absolutely vital to click to that area to have features & traits parsed at all. Even when it pops up on a new character you're just perusing, it goes away with one click, and then doesn't pop up again without a level change.

Kvalyr:
As an end-user, my stance on it is this:

Nothing in Beyond20 needs to be a modal dialog - Beyond20 is an enhancement to my use of D&DBeyond - It should never get in the way of me just using D&DBeyond.

This might not seem like a big deal; but some of us create many characters (or copies of characters) to experiment with builds or to design NPCs, and this modal gets in the way *far* too often - every new/copied character, every level change. It's an annoying interruption and unpleasant UX. Even if it's just a _short_ interruption, it's one that has me disabling Beyond20 ordinarily and only enabling it specifically when I'm about to play with a VTT.

kakaroto:
@Kvalyr I understand your concern and your use case, but I'm not convinced unfortunately. I think you're more of the exception rather than the rule, and I prefer that it be an annoyance to some people vs a problem to the majority. The alert doesn't prevent the page from loading, it can be dismissed by clicking anywhere (i.e: you don't have to move your mouse to click on the "OK" button), it only happens for new sheets or after a level up which should technically be a rare occurence. With nearly 250k users, I have to think about whether it's a positive or a negative change for the majority of these users, and I'm not convinced that making it a simple notification that disappears after 5 seconds is the right move in this case.

I might make an option to make that behavior disable-able, but I try to avoid adding unnecessary options as it's already way too full of options and I try to keep it to the strict minimum.

> I said this stuff originally in a review, but I opened this PR so that I could try to be more than just a complaining end-user. 😂

Yeah, I saw the review and made the link to the PR ;) Thanks for wanting to contribute with code as well as comments :)

kakaroto:
I'm closing this PR. Sorry @Kvalyr but I think having it as a notification is a step in the wrong direction. Remember that if you click the 'features and traits' once (per character), you're done the alert won't appear anymore, and that you can dismiss it by clicking anywhere on the screen, not just by clicking on the ok button, so I think the annoying factor is very low compared to how often we have to respond to people asking about why their options aren't what they expect them to be.

PR #573, opened 2020-10-28
Paladin Sacred Weapon

Aeristoka:
https://www.dndbeyond.com/classes/paladin#OathofDevotion

`As an action, you can imbue one weapon that you are holding with positive energy, using your Channel Divinity. For 1 minute, you add your Charisma modifier to attack rolls made with that weapon (with a minimum bonus of +1).`


kakaroto:
Looking good, thx!

PR #576, opened 2020-10-30
allow hiding the results dialog on ddb on whisper

rispig:


src/common/roll_renderer.js
```diff
@@ -334,11 +334,12 @@ class Beyond20RollRenderer {
             if (discord_error != undefined)
                 this._displayer.displayError("Beyond20 Discord Integration: " + discord_error);
         });
+        const hideResults = request.whisper === WhisperType.YES && this._settings['hide-results-on-whisper'];
 
         if (request.sendMessage && this._displayer.sendMessage)
             this._displayer.sendMessage(request, title, html, character, request.whisper, play_sound, source, attributes, description, attack_rolls, roll_info, damage_rolls, total_damages, open)
         else
-            this._displayer.postHTML(request, title, html, character, request.whisper, play_sound, source, attributes, description, attack_rolls, roll_info, damage_rolls, total_damages, open);
+            !hideResults && this._displayer.postHTML(request, title, html, character, request.whisper, play_sound, source, attributes, description, attack_rolls, roll_info, damage_rolls, total_damages, open);
```
kakaroto, line 342: this looks like minified code, if should be `else if (!hideResults)` instead

src/common/settings.js
```diff
@@ -68,6 +68,13 @@ const options_list = {
         }
     },
 
+    "hide-results-on-whisper": {
+        "title": "Hide roll results on D&D Beyond",
+        "description": "Don't show the roll results on D&D Beyond when using whisper",
+        "type": "bool",
+        "default": false
+    },
+
```
kakaroto, line 77: I disagree with this option because those who aren't using a VTT, will simply not see anything. This should be conditional on sending to Discord only, and the option's text should reflect that too.

kakaroto (changes requested):
Thanks for the PR. I've made a couple of comments, my issue is that I'm realizing that this is convoluted and confusing... I should probably add comments to clarify what this does. 

`sendError` is used when using digital dice, it's so the roll is not posted but sent as a `rendered-roll` message to the VTT. If it fails to find a VTT, the rolls gets rendered locall with `Whisper` type being disabled, and `postHTML` is called directly, so this shouldn't affect you.
Your code though looks like if the option is enabled, it would never post the HTML when whisper is true... but that's not the case, since `postHTML` can get called in other places. I find that to be confusion and annoying and it shows just how much my own code lacks documentation/comments and a clearer design rather than it evolving into the mess it currently is.

I don't have an issue with the feature, other than it needs to be specific to rolling to Discord. As for the implementation, if you have digital dice enabled, it won't work (see `dndbeyondDiceRoller.handleRollError` at the end of `src/base/dice.js`) and if you have it set to send rolls to "All VTTs" but no VTT is open, then it also won't work, for the same reason (as it will force disable whisper in order to render it locally).

Finally, the biggest issue with this patch is that since the roll renderer is used by the Foundry VTT implementation, disabling the postHTML when whisper is enabled means that it will never send any whispers within Foundry itself since it will skip it too.

rispig:
> Thanks for the PR. I've made a couple of comments, my issue is that I'm realizing that this is convoluted and confusing... I should probably add comments to clarify what this does.
> 
> `sendError` is used when using digital dice, it's so the roll is not posted but sent as a `rendered-roll` message to the VTT. If it fails to find a VTT, the rolls gets rendered locall with `Whisper` type being disabled, and `postHTML` is called directly, so this shouldn't affect you.
> Your code though looks like if the option is enabled, it would never post the HTML when whisper is true... but that's not the case, since `postHTML` can get called in other places. I find that to be confusion and annoying and it shows just how much my own code lacks documentation/comments and a clearer design rather than it evolving into the mess it currently is.
> 
> I don't have an issue with the feature, other than it needs to be specific to rolling to Discord. As for the implementation, if you have digital dice enabled, it won't work (see `dndbeyondDiceRoller.handleRollError` at the end of `src/base/dice.js`) and if you have it set to send rolls to "All VTTs" but no VTT is open, then it also won't work, for the same reason (as it will force disable whisper in order to render it locally).
> 
> Finally, the biggest issue with this patch is that since the roll renderer is used by the Foundry VTT implementation, disabling the postHTML when whisper is enabled means that it will never send any whispers within Foundry itself since it will skip it too.

Wow, I really appreciate the effort of explaining all of this, it's not trivial, so thanks for that.
I'll get your issues fixed :)

src/common/roll_renderer.js
rispig: I'm very into code golfing, hence my style :)

rispig:
@kakaroto bump

kakaroto:
No need to bump, it's still on my radar, but I only have 16 hours per day of work and there's a limit to the amount of stuff I can get to in that amount. I expect late next week it will free up. 
Thanks for your patience 👍

kakaroto:
Thanks for your patience. I've reviewed it now and it looks good so I'm merging this as is, but I think the option will potentially change in the future, because it would be better to have "Send to : Discord Only" as an option where it wouldn't show it in D&D Beyond and wouldn't be dependent on the whisper setting either.

rispig:
> Thanks for your patience. I've reviewed it now and it looks good so I'm merging this as is, but I think the option will potentially change in the future, because it would be better to have "Send to : Discord Only" as an option where it wouldn't show it in D&D Beyond and wouldn't be dependent on the whisper setting either.

Sounds like the natural course for this feature

PR #588, opened 2020-11-21
Tashas cauldron of fixes

Aeristoka:
I think this is a pretty fair number at least (not sure I caught all of them) of the new Tasha's stuff.

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -560,6 +560,12 @@ const character_settings = {
         "description": "Align your conciousness to the calculations of Mechanus. You enter a heightened state.",
         "type": "bool",
         "default": false
+    },
+    "druid-symbiotic-entity": {
+        "title": "Circle of Spores: Symbiotic Entity",
+        "description": "Your symbiotic entity lends its power to your melee weapon strikes.",
+        "type": "bool",
+        "default": false
     }
 }
 
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -559,7 +559,7 @@ const character_settings = {
         "title": "Sorcerer: Trance of Order",
         "description": "Align your conciousness to the calculations of Mechanus. You enter a heightened state.",
         "type": "bool",
-        "default": true
+        "default": false
     }
 }
 
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -459,7 +459,7 @@ const character_settings = {
         "type": "bool",
         "default": false
     },
-    "fighter-giants-might": {
+    "fighter-giant-might": {
         "title": "Fighter: Giant’s Might",
         "description": "Activate Giant’s Might to get advantage on Strength checks and saving throws and deal 1d6 extra damage",
         "type": "bool",
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -548,6 +548,12 @@ const character_settings = {
         "description": "Your charisma and deity guide your attacks",
         "type": "bool",
         "default": false
+    },
+    "cleric-blessed-strikes": {
+        "title": "Cleric: Blessed Strikes",
+        "description": "Deal an extra 1d8 damage on damaging cantrips and weapon attacks",
+        "type": "bool",
+        "default": true
     }
 }
 
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -459,9 +459,9 @@ const character_settings = {
         "type": "bool",
         "default": false
     },
-    "fighter-giant-might": {
-        "title": "Fighter: Giant Might",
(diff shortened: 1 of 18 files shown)


src/common/settings.js
```diff
@@ -459,9 +459,9 @@ const character_settings = {
         "type": "bool",
         "default": false
     },
-    "fighter-giant-might": {
-        "title": "Fighter: Giant Might",
-        "description": "Activate Giant Might to get advantage on Strength checks and saving throws and deal 1d6 extra damage",
+    "fighter-giants-might": {
```
kakaroto, line 462: no reason to change the name of the option, that would invalidate whatever setting the user had before

src/dndbeyond/content-scripts/character.js
```diff
@@ -432,9 +453,16 @@ function rollItem(force_display = false, force_to_hit_only = false, force_damage
...
+        if (character.hasClassFeature("Blessed Strikes") &&
+            character.getSetting("cleric-blessed-strikes", false)) {
+            damages.push("1d8");
+            damage_types.push("Blessed Strikes");
+        }
         // Bard's Psychic blades;
         if (character.hasClassFeature("Psychic Blades") &&
-            character.getSetting("bard-psychic-blades", false)) {
+            character.getSetting("bard-psychic-blades", false) &&
+            character.hasClass("Bard")) {
```
kakaroto, line 465: Can you have a class feature of Psychic Blades without being a Bard?  Is that some new feature in Tasha's ?

src/dndbeyond/content-scripts/character.js
```diff
@@ -726,6 +774,13 @@ function rollAction(paneClass, force_to_hit_only = false, force_damages_only = f
                 damages.push("1d8");
                 damage_types.push("Radiant");
             }
+
+            // Cleric Blessed Strikes
+            if (character.hasClassFeature("Blessed Strikes") &&
+            character.getSetting("cleric-blessed-strikes", false)) {
```
kakaroto, line 780: some indentation issues here

kakaroto: Blessed Strikes applies to "weapon attacks and cantrips", so it should apply on rollItem and rollSpell. I'm unsure about rollAction. Wouldn't actions be non-weapon attacks ?

kakaroto (changes requested):
Looking good, just small tweaks.

src/dndbeyond/content-scripts/character.js
Aeristoka: This is based on a conversation we had some time back about us deciding weapon attacks included natural weapons (which would then include fists/claws/tails/etc.)

src/dndbeyond/content-scripts/character.js
Aeristoka: Yeah, thank WotC for this crap:
Rogue Psychic Blades: https://www.dndbeyond.com/sources/tcoe/rogue#PsychicBlades
Bard College of Whispers Psychic Blades: https://www.dndbeyond.com/classes/bard#CollegeofWhispers

Just, ya know, named the same dang stupid exact same name...

src/common/settings.js
Aeristoka: Good point, fixed

src/dndbeyond/content-scripts/character.js
kakaroto: oh cool, so.. who cares about multiclassers!

Aeristoka: They'll still be good, as it catches only for Bard here, and they'd have that if they multi-classed bards & rogues.

Unless you're mouthing off at WotC, if so I AGREE!

kakaroto: yeah, I was mouthing off 😁 
But if someone has both rogue psychic blades and bard psychic blades then they're SOL.

Aeristoka: Shouldn't be, the part for the Rogue is elsewhere in the code

kakaroto: eh, right, I'm stupid, I was thinking of rolling "Psychic Blades" not of rolling something while also having the feat.
How about you're a Bard, but not yet 3rd level, or not of the right college, and you have rogue psychic blades.
(note: I'm just thinking of issues that I have no intentions of fixing. There's an option for that feat anyway, just annoyed at WotC crap)

Aeristoka: yeah, that could be an issue... they'd have to have it toggled on...

I say screw it as well, the DM and Player should know what the things be

PR #596, opened 2020-12-03
Fix RegEx for monster attack when attack default damage is missing

flangelier:
This is related to the bug seen on discord, a homebrew monster with an attack text without the default damage was giving a bad roll. Eg: 
`...Hit: 2d4 + 2 ...` would be interpret as `d4+2` 
because the regex took the first number as the default damage.

I also change the [0-9] to \d

### DIFF
--- a/dist/dndbeyond_monster.js
+++ b/dist/dndbeyond_monster.js
@@ -3835,7 +3835,7 @@ class Monster extends CharacterBase {
         if (hit_idx > 0)
             hit = description.slice(hit_idx);
         // Using match with global modifier then map to regular match because RegExp.matchAll isn't available on every browser
-        const damage_regexp = new RegExp(/([\w]* )(?:([0-9]+))?(?: *\(?([0-9]*d[0-9]+(?:\s*[-+]\s*[0-9]+)?(?: plus [^\)]+)?)\)?)? ([\w ]+?) damage/)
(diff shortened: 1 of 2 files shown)


Aeristoka:
Looks like you didn't actually build the changes here, see: https://github.com/kakaroto/Beyond20#build, and add the built files as part of the commit.

flangelier:
@Aeristoka didn't know that! Thx!

Aeristoka:
> @Aeristoka didn't know that! Thx!

Welcome (soon) to the contributor cabal!

kakaroto:
Thanks @flangelier! That's a good catch, thanks for looking into it and finding the regexp bug. The fix looks good 👍 

I prefer however to keep `[0-9]` on purpose (you're not the first to make the comment btw) because while `\d` is convenient, I think it makes it more readable when it's not used. As an example, what is more readable `[0-9]*d[0-9]+` or `\d*d\d+` Especially with how complex the regexp is, it makes it easier to parse it and see where the values are, whereas if it's all full of the abbreviations, it's hard for the eyes to see where we should split it mentally. Especially with the `d` from the dice notations making it possibly confusing sometimes.

Please revert the `\d` change and add the `[^d]` fix and I'll merge it, thanks!

flangelier:
I made a cleaner PR #599

PR #599, opened 2020-12-04
Fix RegEx for monster attack when attack default damage is missing

flangelier:
I hate adding commit with no values in the log, so I made the changes of PR #596
I will close the other PR.

This is related to the bug seen on discord, a homebrew monster with an attack text without the default damage was giving a bad roll. Eg:
`...Hit: 2d4 + 2 ...` would be interpret as `d4+2`
because the regex took the first number as the default damage.


kakaroto:
hehe, it's fine, I would have squashed the commits in that case, otherwise, in other situation, it's good to just append, so I could see which of my comments were handled, etc..
Thanks for the quick fix! 👍

PR #600, opened 2020-12-04
Bugfix for Artificer: Alchemical Savant

### LINKED ISSUE #591: Alchemist Artificer: Alchemical Savant should not apply to `Aid`
Thanks for the awesome plugin! It makes playing with d&d beyond so much easier :-)

**Describe the bug**

When Casting the `Aid` spell as an Alchemist Artificer the Alchemical Savant feature is applied / displayed while according to the rules it shouldn't apply because there is no _roll_ involved.

**Expected behavior**

The Alchemical Savant feature should not apply when there is no roll involved. When there are no dice in the roll20 "roll" it shouldn't apply.

> **Alchemical Savant**
> *5th-level Alchemist feature*
> 
> You've developed masterful command of magical chemicals, enhancing the healing and damage you create through them. Whenever you cast a spell using your alchemist’s supplies as the spellcasting focus, you gain a bonus **to one roll** of the spell. That roll must restore hit points or be a damage roll that deals acid, fire, necrotic, or poison damage, and the bonus equals your Intelligence modifier (minimum of +1).

> **Aid**
> Your spell bolsters your allies with toughness and resolve. Choose up to three creatures within range. Each target's hit point maximum and current hit points increase by 5 for the duration.

**Screenshots**

image

Aeristoka:
Fixes #591

### DIFF
--- a/dist/dndbeyond_character.js
+++ b/dist/dndbeyond_character.js
@@ -5391,10 +5391,11 @@ function rollSpell(force_display = false, force_to_hit_only = false, force_damag
 
         if (character.hasClassFeature("Alchemical Savant") &&
             character.getSetting("artificer-alchemical-savant", false) &&
-            damages.length) {
+            damages.length > 0) {
+            const alchemical_savant_regex = /[0-9]+d[0-9]+/g;
             for (let i = 0; i < damages.length; i++){
-                if (damage_types[i] === "Acid" || damage_types[i] === "Fire" ||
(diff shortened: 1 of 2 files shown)


Aeristoka:
@kakaroto let me know if that regex looks ok for making sure "this damage is a die roll, instead of a flat amount". As #591 points out, Alchemical Savant should only be applying on Die Rolls.

kakaroto:
Looks good to me!

PR #601, opened 2020-12-05
Implement feat: Piercer

Aeristoka:
From Tasha's Cauldron of Everything
https://www.dndbeyond.com/feats/piercer

### DIFF
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -155,13 +155,11 @@ function buildAttackRoll(character, attack_source, name, description, properties
                 }
             }
             if (character.hasFeat("Piercer")) {
-                const piercer_feat_regex = /[0-9]+/g;
                 for (i = 0; i < damage_types.length; i++) {
                     if (damage_types[i].includes("Piercing")){
-                        let piercer_damage_array = [damages[i]];
-                        let piercer_damage = damagesToCrits(character, piercer_damage_array);
-                        if (piercer_damage[0].length > 0) {    
-                            piercer_damage[0].replace(piercer_feat_regex,1);
+                        let piercer_damage = damagesToCrits(character, [damages[i]]);
+                        if (piercer_damage.length > 0 && piercer_damage[0] != "") {    
+                            piercer_damage[0] = piercer_damage[0].replace(/([0-9]+)d([0-9]+)/, '1d$2');
                             crit_damages.push(piercer_damage[0]);
                             crit_damage_types.push("Piercer Feat");
                         }
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -154,6 +154,20 @@ function buildAttackRoll(character, attack_source, name, description, properties
                     crit_damage_types.push(damage_types[i]);
                 }
             }
+            if (character.hasFeat("Piercer")) {
+                const piercer_feat_regex = /[0-9]+/g;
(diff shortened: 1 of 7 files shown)


src/dndbeyond/base/utils.js
```diff
@@ -154,6 +154,20 @@ function buildAttackRoll(character, attack_source, name, description, properties
...
                 }
             }
+            if (character.hasFeat("Piercer")) {
+                const piercer_feat_regex = /[0-9]+/g;
+                for (i = 0; i < damage_types.length; i++) {
+                    if (damage_types[i].includes("Piercing")){
+                        let piercer_damage_array = [damages[i]];
+                        let piercer_damage = damagesToCrits(character, piercer_damage_array);
+                        if (piercer_damage[0].length > 0) {    
+                            piercer_damage[0].replace(piercer_feat_regex,1);
```
kakaroto, line 164: 1 - you don't need the `piercer_damage_array`, so it's best to just pass that directly to the `damagesToCrits(character, [damages[i]])` call
2 - you need to check `if (piercer_damage.length > 0)` not `if (piercer_damage[0].length > 0)`
3 - You're replacing with the regexp, any number into 1, so a flat amount (homebrew crit) would be replaced with 1. It's better to replace : `piercer_damage[0].replace(/([0-9]+)d([0-9]+)/, '1d$2')` to only replace the dice amount
4 - You're calling replace and not storing its result. The replace return value should be the one pushed into `crit_damages`

Aeristoka: 1. From my testing, if I didn't have that array, it flipped out damagesToCrits, just sent me back a bunch of quote pairs
2. Yep
3. From my testing that was only catching the first match, so it didn't matter, but that does seem cleaner
4. Woop, I think this is a result of me failing to find ANY dang weapon that has a 2dX that does piercing damage, so the replace never actually going off never actually mattered...

kakaroto: 1. yeah, still pass it as an array, you just don't need to extract it as its own variable :)
3. you used `/g` flag in your regexp, so it would have caught them all (i.e: `2d8` becoming `1d1`), the issue though is that a flat number would have also been replaced (i.e a `8` becoming `1`)
4. explained why 3) above worked in your testing 🤣

Aeristoka: Rebased and fixed, lemme know.

src/dndbeyond/base/utils.js
```diff
@@ -154,6 +154,18 @@ function buildAttackRoll(character, attack_source, name, description, properties
...
+                for (i = 0; i < damage_types.length; i++) {
+                    if (damage_types[i].includes("Piercing")){
+                        let piercer_damage = damagesToCrits(character, [damages[i]]);
+                        if (piercer_damage.length > 0 && piercer_damage[0] != "") {    
+                            piercer_damage[0] = piercer_damage[0].replace(/([0-9]+)d([0-9]+)/, '1d$2');
+                            crit_damages.push(piercer_damage[0]);
+                            crit_damage_types.push("Piercer Feat");
+                        }
+                    }
+                }
```
kakaroto, line 167: missing a `break;` in the for loop since it says you can reroll one of the attack's damage dice.
Will fix post merge

Aeristoka: Ah yeah, good catch.

Real fun trying to find a weapon thing with multiple piercing damages...

PR #615, opened 2020-12-09
Character subclass settings name changes + ordering

pandanielxd:
Edited the character settings to always show what class they come from + subclass and ordered them from a-z. Non-class specific settings are at the bottom.

### DIFF
--- a/dist/astral.js
+++ b/dist/astral.js
@@ -564,23 +564,185 @@ const character_settings = {
         "type": "string",
         "default": ""
     },
-    "rogue-sneak-attack": {
-        "title": "Rogue: Sneak Attack",
-        "description": "Send Sneak Attack damage with each attack roll",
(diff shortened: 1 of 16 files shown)


kakaroto:
Yeay for your first PR! 🥳 
So, issue : You'd need to make that change in the `src/common/settings.js` then follow the instructions in the README for how to build (`npm install` and `npm run build`). That will generate the files in the `dist` folder. Otherwise your changes will be overwritten, and also the settings list appears in more than just one file.
Fix that, make a new commit and push it again, and it should update this PR automatically 👍

Aeristoka:
What I like to do in the Search Settings of Visual Studio Code (my preferred IDE) is to tell the search to ignore 'dist' so that I never see those files in search.

And yeah, always work in src/

pandanielxd:
hopefully did it correctly now 👌

PR #621, opened 2020-12-09
Ordering the shortcut actions in the settings screen

flangelier:
I think the title is self explanatory :)


kakaroto:
Thanks for the PR but I'm going to reject it for now. While it does the job, I do prefer having the ability to decide on the order, and the ordering has already been changed to be sorted by PR #615. I might revisit this if I find it's problematic having to place new options in the right position in the array, but I think it's going to be ok.
Thanks

PR #627, opened 2020-12-12
Implement escaping commas in damage bonus

Cube-o:
Implemented the option to escape commas in damage dice formula bonus field using `\`.

Wasn't sure if it is wanted there, but I included mentions of the possibility in option and on the FAQ page.

### DIFF
--- a/dist/astral.js
+++ b/dist/astral.js
@@ -554,7 +554,8 @@ const character_settings = {
     },
     "custom-damage-dice": {
         "title": "Custom Damage dice formula bonus",
-        "description": "Add custom dice to damage rolls (Magic Weapon, Elemental Weapon, Green-flame Blade, etc..). Use a comma to separate multiple independent rolls.",
+        "description": "Add custom dice to damage rolls (Magic Weapon, Elemental Weapon, Green-flame Blade, etc..). Use a comma to separate multiple " +
+            "independent rolls.\nFor macros including commas inside the formula, preceed each comma with \\",
         "type": "string",
         "default": ""
     },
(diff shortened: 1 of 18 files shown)


kakaroto:
Thanks! Will review some time later.
I think escaping is a great idea, and would like to have it, but here's a thought : 
What if escaping wasn't necessary if the formula starts with `?` ? This way, if it's a Roll20 query, it detects it, and therefore just parses it normally without splitting it as separate damages on the comma.

Cube-o:
Could work, although probably if it's done on the whole group of `?` plus both of the curly brackets. Because otherwise a question mark inside the query (let's say they want the dropdown to be to a question) would throw that off.
And also the question mark doesn't have to be at the start, as there can be the damage type for that damage.

Now that you got me thinking about it, how about just having the split ignore commas that are inside a pair of curly brackets?
The result would be the same, but there wouldn't be any new instruction the users would have to understand.

kakaroto:
> Now that you got me thinking about it, how about just having the split ignore commas that are inside a pair of curly brackets?
> The result would be the same, but there wouldn't be any new instruction the users would have to understand.

Yeah, I like that.. you open a parenthesis or a bracket, then you need to close it, otherwise the comma is ignored.. so you could even do `Dmg1: +1d4[foo,bar], Dmg2: ?{query|1d4,1d8}` or whatever

I like this solution quite a bit, especially since it's the least painful for users having to understand how to do things.

kakaroto:
Thanks for your suggestion. I've implemented it differently and I think it works great now with no need of manually escaping anything. Feel free to review the commit 5a3dae7 and let me know if you have any feedback about it.

PR #643, opened 2020-12-27
Add toggle for Colossus Slayer

IvanGirderboot:
- Added a settings toggle for the next role to add the ranger's Colossus Slayer damage die.
- Also fixed a bug where multiple settings changed during a damage role were not being applied properly.
- Fixed some typos.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -1,3 +1,8 @@
+Unreleased
+====
+* **Feature**: Add a toggle option for the Ranger's Colossus Slayer ability, rather than always rolling it
+* **Bugfix**: Fixed an issue where only one setting could be changed during damage rolls, even if multiple settings should have been changed
+
 v2.3.0
 ===
 
@@ -11,7 +16,7 @@ v2.3.0
 * **Feature**: Allow queuing up of rolls when the Digital Dice are enabled so all rolls are executed
 * **Feature**: Add support for capturing and transferring digital dice rolled manually through D&D Beyond's interface
 * **Feature**: Add support for parsing the item customization options (Hex Weapon, Pact Weapon)
-* **Feature**: Display a class feature's choices (such as selected profiency or the Monastic Tradition) when displaying a feature to the VTT
+* **Feature**: Display a class feature's choices (such as selected proficiency or the Monastic Tradition) when displaying a feature to the VTT
 * **Feature**: Add option to hide from the player the results of a whispered roll that is sent to Discord (by [@rispig](https://github.com/rispig))
 * **Feature**: Add a notification when the class features are parsed so the user gets visual feedback when it's done
 * **Feature**: Wizard Bladesong: Support concentration constitution saves and adds the intelligence modifier (by [@Aeristoka](https://github.com/Aeristoka))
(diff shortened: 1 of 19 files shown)


src/extension/popup.js
```diff
@@ -145,6 +145,11 @@ function populateCharacter(response) {
             e = createHTMLOption("ranger-planar-warrior", false, character_settings);
             options.append(e);
         }
+        if (response["class-features"].includes("Hunter’s Prey: Colossus Slayer")) {
+            console.log('CS Class feature detected.');
```
kakaroto, line 149: `console.log` for debugging should be removed from final commit

src/dndbeyond/content-scripts/character.js
```diff
@@ -240,6 +240,7 @@ function rollItem(force_display = false, force_to_hit_only = false, force_damage
     if (!force_display && Object.keys(properties).includes("Damage")) {
         const item_full_name = $(".ct-item-pane .ct-sidebar__heading .ct-item-name,.ct-item-pane .ct-sidebar__heading .ddbc-item-name").text();
         let to_hit = properties["To Hit"] !== undefined && properties["To Hit"] !== "--" ? properties["To Hit"] : null;
+        let settings_to_change = {}
```
kakaroto, line 243: the variable should be const, since it doesn't change (its content is modified, but not the variable itself)

src/dndbeyond/content-scripts/character.js
```diff
@@ -419,11 +420,13 @@ function rollItem(force_display = false, force_to_hit_only = false, force_damage
...
                 damage_types.push("Ambush");
-                character.mergeCharacterSettings({ "ranger-dread-ambusher": false });
+                settings_to_change["ranger-dread-ambusher"] = false;
             }
-            if (character.hasClassFeature("Hunter’s Prey: Colossus Slayer")) {
+            if (character.hasClassFeature("Hunter’s Prey: Colossus Slayer") &&
+                character.getSetting("ranger-colossus-slayer", false)) {
                 damages.push("1d8");
                 damage_types.push("Colossus Slayer");
+                settings_to_change["ranger-colossus-slayer"] = false; // Reset to off since CS only applies once per turn
```
kakaroto, line 429: CS shouldn't be resetting every time, it's supposed to be considered as Sneak Attack, where if you enable it, it would appear, but the player can decide to count it or not. Most likely, player will have one attack per turn, or only one of their two attacks will succeed. It's easier to ignore a number when doing the math, than to go into settings and re-enable it before each click.

docs/Changelog.md
```diff
@@ -1,3 +1,8 @@
+Unreleased
```
kakaroto, line 1: Oh I like this idea of keeping track of the changelog under an unreleased tag.

kakaroto (approved):
Thank you for the PR! There's a couple of small tweaks, which I will take care of locally before merging. My main issue is that these should have been 3 separate commits instead of 1 commit that includes all 3 changes. Having commits "as small as possible" is a good practice and makes for a cleaner history, git bisect and git blame results, or cherry-picking.
I like the idea of having that "unreleased" changelog, though for me, what I usually do is to translate the commit log into the changelog when it's time to release, so that's also helpful when you have one commit per feature/bugfix.

Thanks!

kakaroto:
Merged manually in b503e14e4fc9cd51846145e10a099980c89fbc6d, 4c94e9cd83e8029b7f982f2cfe981a69f3ac04a4 and 2a212de4f726f1d283b73218b36501f8adbae7c5
Thanks!

PR #647, opened 2021-01-07
Implement Half-Elf Variant Racial Features

Aeristoka:
Mark of Detection Half-Elf: Deductive Intuition
Mark of Storm Half-Elf: Windwright's Intuition

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -113,6 +113,10 @@ async function rollSkillCheck(paneClass) {
             roll_properties.d20 = `1d20min${min}`
         }
     }
+
+    if (character.hasRacialTrait("Deductive Intuition") && (skill_name == "Investigation" || skill_name == "Insight")){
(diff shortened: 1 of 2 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -113,6 +113,14 @@ async function rollSkillCheck(paneClass) {
             roll_properties.d20 = `1d20min${min}`
         }
     }
+
+    if (character.hasRacialTrait("Deductive Intuition") && (skill_name == "Investigation" || skill_name == "Insight")){
```
kakaroto, line 117: For the future, I'd do `&& ["Investigation", "Insight"].includes(skill_name)` as it's a nicer way to check for it and scales more easily.

kakaroto:
Thanks, merged!

src/dndbeyond/content-scripts/character.js
Aeristoka: ah yeah, definitely does

PR #657, opened 2021-01-17
Fix for DNDBRoll -> FVTTRoll transform

### LINKED ISSUE #655: Chaos Bolt Damage not working
Chaos Bolt: If the d8 damage rolls are equal, dialog box never shows up and damage roll stalls and will not be produced.  Using Foundry VTT.

Aeristoka:
Fixes #655


kakaroto:
oh wow, what a nice fix! Good job on finding how to fix the issue with such an easy to review patch! 👍

Issue #671, opened 2021-02-01
Add +5 damage for Charger Feat as a pull down toggle


fireboy92k:
From our Discord discussion:
that Feat could be implemented as a toggle like other similar things
so, if you knew you were fulfilling the requirements for the +damage, toggle it on, then roll, it would add, and toggle it back off

From me:
Suppress the 1d20 + 5 roll from the Features UI if possible as well to eliminate confusion.

Aeristoka:
Pull request is in!

PR #680, opened 2021-02-03
Implement Charger Feat

Aeristoka:
Fixes Add +5 damage for Charger Feat as a pull down toggle #671

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -599,6 +599,12 @@ const character_settings = {
         "description": "Unleash your divine soul to deal extra radiant damage equal to your level.",
         "type": "bool",
         "default": false
+    },
+    "charger-feat": {
+        "title": "Charger Feat - Extra Damage",
+        "description": "You charge into battle, lending weight to your blow!",
+        "type": "bool",
+        "default": false
     }
 }
 
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -601,7 +601,7 @@ const character_settings = {
         "default": false
     },
     "charger-feat": {
-        "title": "Charger Feat - Extra Damage",
+        "title": "Charger Feat - Extra Damage (Apply to next roll only)",
         "description": "You charge into battle, lending weight to your blow!",
         "type": "bool",
         "default": false
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -833,7 +833,6 @@ function rollAction(paneClass, force_to_hit_only = false, force_damages_only = f
         }
 
         // Charger Feat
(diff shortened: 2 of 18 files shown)


src/common/settings.js
```diff
@@ -599,6 +599,12 @@ const character_settings = {
         "description": "Unleash your divine soul to deal extra radiant damage equal to your level.",
         "type": "bool",
         "default": false
+    },
+    "charger-feat": {
+        "title": "Charger Feat - Extra Damage",
+        "description": "You charge into battle, lending weight to your blow!",
+        "type": "bool",
+        "default": false
```
kakaroto, line 607: Seems to disable itself when used (makes sense as per the rules), I'd suggest adding a "(Apply to next roll)" like we have with sharpshooter.

src/dndbeyond/content-scripts/character.js
```diff
@@ -817,6 +826,14 @@ function rollAction(paneClass, force_to_hit_only = false, force_damages_only = f
                 damages.push("1d8");
                 damage_types.push("Blessed Strikes");
             }
+
+            // Charger Feat
+            // I don't love the action_name isn't "Lightning Launcher" here, but it removes code duplication of the containing if statement above, while keeping it Melee-only
+            if (character.getSetting("charger-feat") && !action_name.includes("Lightning Launcher")) {
```
kakaroto, line 832: oh boy, yeah, that was confusing. Here's my suggestion, all the way up in line 744, I'd do : 
```js
const isMeleeAttack = action_name.includes("Polearm Master - Bonus Attack") || action_name.includes("Unarmed Strike") || action_name.includes("Tavern Brawler Strike")
            || action_name.includes("Psychic Blade") || action_name.includes("Bite") || action_name.includes("Claws") || action_name.includes("Tail")
            || action_name.includes("Ram") || action_name.includes("Horns") || action_name.includes("Hooves") || action_name.includes("Talons")
            || action_name.includes("Thunder Gauntlets");

if (isMeleeAttack || action_name.includes("Lightning Launcher")) {
....
```

And here, you can do `&& isMeleeAttack`, it would look nicer and easier to maintain long term than a negative check that excludes a previous if.

kakaroto (changes requested):
Some small improvements needed, thx!

src/common/settings.js
Aeristoka: Done

src/dndbeyond/content-scripts/character.js
Aeristoka: I wish DnDBeyond just had them specified with the Melee or Ranged info still... But yeah, I'll do this

PR #684, opened 2021-02-04
Added support for new resource bars in Astral

adriangaro:
This PR will handle the changes which happened some time ago to how the hp bar was handled and now allows multiple resource bars. The changes in this PR will include:

- updating the hp, max hp, and temp hp attributes in the character sheet. The default ones used by astral for dnd5e sheet are used.
- updating the resource bars to use said attributes above and force update them to include new values.
- the temp hp bar would be hidden if no temp hp
- added setting to be able to configure if you want beyond20 to mess with the resource bars in Astral (might not be needed, but someone might just want astral to mess with stats and that's it.


src/astral/page-script.js
```diff
@@ -205,19 +205,29 @@ async function updateHpBar({characterName, hp, maxHp, tempHp}) {
...
+            },
             body: JSON.stringify({
                 character: {
                     updateAt: Date.now(),
-                    attributeBarMax: maxHp + tempHp,
-                    attributeBarCurrent: hp + tempHp,
-
+                    customAttributes: characterData.customAttributes,
+                    ...(settings['astral-control-hp-bar'] ? {
+                        resourceBars: characterData.resourceBars
```
kakaroto, line 226: 😱 at that construct
For readability I'd define it before the fetch : 
```js
const resourceBarsUpdate = {};
if (settings['astral-control-hp-bar']) 
  resourceBarsUpdate['resourceBars'] = characterData.resourceBars;
return fetch(/* [.....] */
  customAttributes : characterData.customAttributes,
  ...resourceBarsUpdate
}
```

src/astral/page-script.js
```diff
@@ -205,19 +205,29 @@ async function updateHpBar({characterName, hp, maxHp, tempHp}) {
             console.warn(`Couldn't update the character hp due to the following reason: No character found with name ${characterName}`)
             return
         }
+        const characterData = await getCharacterData(character);
```
kakaroto, line 208: That function doesn't do any checks, you'd need to catch it in case it errors out, and handle the fail case so it doesn't prevent the rest from happening.

src/astral/page-script.js
```diff
@@ -205,19 +205,29 @@ async function updateHpBar({characterName, hp, maxHp, tempHp}) {
...
+        updateCustomAttribute(characterData, "HP_Max", maxHp);
+        updateCustomAttribute(characterData, "HP_Current", hp);
+        updateCustomAttribute(characterData, "HP_Temp", tempHp);
+        if (settings['astral-control-hp-bar']) {
+            updateResourceBar(characterData, "hp", "HP_Current", "HP_Max", "48bb78", hp, maxHp, true);
+            updateResourceBar(characterData, "temphp", "HP_Temp", "HP_Temp", "48b3bb", tempHp, tempHp, tempHp != 0);
```
kakaroto, line 214: what are these "48bb78" and "48b3bb" magic values? if it's colors, then add a comment above it to explain that it's RGB values, and not a hash/key.

src/astral/data-utils.js
```diff
@@ -95,6 +95,51 @@ const getCharacter = async (name) => {
...
+            value: value.toString()
+        }
+        characterData.customAttributes.push(attr);
+        return;
+    }
+    attr.value = value.toString();
+}
+
+const updateResourceBar = async (characterData, barName, value, maxValue, color = null, actualValue = null, actualMaxValue = null, display = true) => {
+    let bar = characterData.resourceBars.find(attr => attr.label == barName);
```
kakaroto, line 120: any possibility the character data doesn't have a `resourceBars` field? same for `customAttributes` above? Try a blank new character and make sure those two fields are defined and of type array.

src/astral/data-utils.js
```diff
@@ -95,6 +95,51 @@ const getCharacter = async (name) => {
...
+        leftEquation: value.toString(),
+        rightEquation: maxValue.toString(),
+        display,
+        ...(actualValue != null ? {
+            leftValue: actualValue
+        } : {}),
+        ...(actualMaxValue  != null ? {
+            rightValue: actualMaxValue
+        } : {}),
+        ...(color && !bar.color ? { color } : {})
```
kakaroto, line 139: Same comment as above (or below?) with the fetch. The construct works, but makes it hard for readability/maintainability. Better to spread it out a bit in my opinion.

src/common/settings.js
```diff
@@ -283,6 +283,13 @@ const options_list = {
         // callbacks will be added after the functions are defined
     },
 
+    "astral-control-hp-bar": {
+        "title": "Control and sync Astral hp resource bars",
+        "description": "Beyond20 will control hp and temp hp resource bars and sync hp, it will use the first bar, the default one for hp, and the second one for temp hp.",
+        "type": "bool",
+        "default": true
+    },
+
```
kakaroto, line 292: Why is this necessary? There's already the `update-hp` setting, so why not use that directly? IS there a reason to have two settings about updating the hp ? Or is this just for setting the bars, in which case is it necessary (i.e: can the user define the bars to match hp, and if they're not, then don't do anything, or if hp update is enabled, then always sync the hp bars too)?

kakaroto (commented):
A couple of questions, a readability issue in the constructs you used, the only real issue is in case the api request fails, it wouldn't be handled properly.

src/astral/page-script.js
adriangaro: Yeah will do, they are colors.

src/astral/data-utils.js
adriangaro: From what I know they should all have those attributes since even new tokens have hp and a resourceBar, but will add some sanity checks.

src/astral/page-script.js
adriangaro: Well without character data it would be pointless to continue either way since we don't have the data structures and objects to update the char sheet. So I would say it would be better to crash here and the error would be caught in the catch below and logged.

src/common/settings.js
adriangaro: Added it at first to check some things out that they didn't break. Wanted to make it such that you could allow beyond20 to manage the resource bars or not and configure them manually in astral to reflect the updated stats. But sadly resource bars don't update their values after updating the hp stats through a post request, so we need to force update the bars too. I guess at this point it is vestigial from the initial purpose it had.

src/astral/data-utils.js
adriangaro: On second thought if they were undefined it would be better to simply crash and let the catch below throw the log. You can't do partial updates to `resourceBars` and `customAttributes` at least from what I have tried without breaking the actor/character. You need to send all existing bars and attributes. So if we were to populate them if undefined for whatever reason, it might cause problems down the line by maybe removing attributes created in the stat sheet by the user.

src/astral/page-script.js
kakaroto: shouldn't it fallback on what you had before which didn't use the character data? 
```
                    attributeBarMax: maxHp + tempHp,
                    attributeBarCurrent: hp + tempHp,
```

kakaroto:
Thanks for the changes. I suggested a fallback if character data can't be fetched or the data is wrong, so I would do something like : 
```js
const updateData = {};
try {
  const character = getCharacterData()
  // set all the stuff
  Object.assign(updateData, {..theStuff});
} catch (err) {
  updateData.attributeBarMax = maxHp + tempHp,
  updateData.attributeBarCurrent = hp + tempHp,
}
return fetch(..., body: updateData);
```

src/astral/page-script.js
adriangaro: The one previous isn't use anymore, I think those attributes are deprecated.

kakaroto: ah ok, thanks!

kakaroto:
Thanks for the PR!

PR #686, opened 2021-02-08
Fix for Unarmed Fighting not processing properly

Aeristoka:
In the same vein as Unarmed Strike, but ya know, named differently...


kakaroto:
I'd like to say thanks, but I'll instead say "sorry" because 🤦

PR #712, opened 2021-03-18
Bugfix for Ranger: Gathered Swarm

Aeristoka:
Now applies to rollAction and rollSpell as well, which is the more proper implementation RAW and RAI.

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -542,7 +542,7 @@ const character_settings = {
     },
     "ranger-gathered-swarm": {
         "title": "Ranger: Swarmkeeper: Gathered Swarm",
-        "description": "Use your Gathered Swarm ability to add extra Force damage to your weapon attacks",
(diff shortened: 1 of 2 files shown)


kakaroto:
Who approved the previous implementation who did 2d6 instead of 1d8?! That person should get fired!
Thanks for this!

Aeristoka:
> Who approved the previous implementation who did 2d6 instead of 1d8?! That person should get fired!
> Thanks for this!

You did...

Aeristoka:
> Who approved the previous implementation who did 2d6 instead of 1d8?! That person should get fired!
> Thanks for this!

More seriously, I believe it was originally implemented when it was UA, and I think this was a holdover from the UA version, then never got fixed.

kakaroto:
Ah, that makes sense then! Thanks for this fix and catching the damage change :)

PR #719, opened 2021-03-20
Apply Symbiotic Entity to rollAction Melee Attacks

Aeristoka:


kakaroto:
That whole block is inside ` if ( isMeleeAttack || action_name.includes("Lightning Launcher")) {`
So shouldn't this check for symbiotic Entity be `if (isMeleeAttack && symbiotic....)` ? Or should it also apply on Lightning Launcher action?

Aeristoka:
> That whole block is inside ` if ( isMeleeAttack || action_name.includes("Lightning Launcher")) {`
> So shouldn't this check for symbiotic Entity be `if (isMeleeAttack && symbiotic....)` ? Or should it also apply on Lightning Launcher action?

Good catch, fixed now.

src/dndbeyond/content-scripts/character.js
```diff
@@ -898,6 +898,12 @@ function rollAction(paneClass, force_to_hit_only = false, force_damages_only = f
             settings_to_change["charger-feat"] = false;
         }
 
+        // Druid: Circle of Spores - Symbiotic Entity
+        if (isMeleeAttac && character.hasClassFeature("Symbiotic Entity") && character.getSetting("druid-symbiotic-entity", false)) {
```
kakaroto, line 902: You clearly didn't test this 🙄  (type in variable name 😝)

Aeristoka: Not sure what you're talking about 😇

PR #720, opened 2021-03-20
Implement Crit Damage for Jim's Magic Missile

### LINKED ISSUE #139: Jim's magic missle from Acquisitions Inc doesn't calculate damage correctly
On a crit, it doubles the dice.  it should go from 2d4 to 5d4.

Aeristoka:
Fixes #139


kakaroto:
Nice!

PR #723, opened 2021-03-23
Implement Minus Sign fix for NPC Sheets

### LINKED ISSUE #628: Support minus sign character in monster sheets
Some Monster statblocks use the minus sign instead of a dash (`−` instead of `-`) causing parsing of dice formulas to fail.

https://www.dndbeyond.com/monsters/giant-fire-beetle
> Cistern
Just confirmed that it doesn't work for the awakened shrub or baboon attacks either, although rolled hit points with a minus do work (like for the hawk and owl).  You can graphically see that their is a difference in lengths of the minus used in hit points for creatures as opposed to the minus used for attacks as noted above.

The best solution would be a `.replace` instead of adding it to the regexp because the minus sign would not work in `parseInt` and other areas where the math happens.

Aeristoka:
Fixes #628

### DIFF
--- a/src/dndbeyond/base/monster.js
+++ b/src/dndbeyond/base/monster.js
@@ -381,7 +381,8 @@ class Monster extends CharacterBase {
                 action_name = action_name.slice(0, -1);
             //console.log("Action name: ", action_name);
             if (add_dice) {
-                const description = descriptionToString(action);
+                let description = descriptionToString(action);
+                description = description.replace(/−/g, "-");
                 const roll_properties = this.buildAttackRoll(action_name, description);
                 if (roll_properties) {
                     const id = addRollButton(this, () => {


src/dndbeyond/base/monster.js
```diff
@@ -283,6 +283,7 @@ class Monster extends CharacterBase {
         if (hit_idx > 0)
             hit = description.slice(hit_idx);
         // Using match with global modifier then map to regular match because RegExp.matchAll isn't available on every browser
+        hit = hit.replace("−", "-");
```
kakaroto, line 286: use `.replace(/−/g, "-")` so it replaces it all, not just the first one it encounters. Also, should be done on the description itself, not just the to-hit, as it could affect damage value too.

PR #724, opened 2021-03-23
Removal of Cleric: Disciple of Life Handling

Aeristoka:
DnDBeyond now natively supports Disciple of Life

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -553,7 +553,7 @@ class Beyond20RollRenderer {
                 all_rolls.push(roll);
                 const dmg_type = damage_types[i];
                 let damage_flags = DAMAGE_FLAGS.REGULAR;
-                if (["Healing", "Disciple of Life", "Temp HP", "Alchemical Savant Healing", "Enhanced Bond Healing"].includes(dmg_type)) {
+                if (["Healing", "Temp HP", "Alchemical Savant Healing", "Enhanced Bond Healing"].includes(dmg_type)) {
                     damage_flags = DAMAGE_FLAGS.HEALING;
                 } else if (i == 0) {
                     damage_flags = DAMAGE_FLAGS.REGULAR;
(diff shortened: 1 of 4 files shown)


kakaroto:
Good bye friend, we had a good run, you shall be missed.

jkomg:
It's honestly too bad - they're just lumping Disciple of Life into the main healing of the spell instead of showing it as separate in the same roll the way Beyond20 does - which is much more elegant a solution. 

This still doesn't show Blessed Healer and makes you have to think about it, too, because you can't just take the Beyond20 showing of Disciple of Life and add that to yourself, you gotta go "Okay, 2 + spell level to me..." which is clearly a small problem or nonissue but...

Sometimes the "fix" seems lazy compared to the workaround.

TL;DR: we appreciate you @kakaroto

Aeristoka:
> It's honestly too bad - they're just lumping Disciple of Life into the main healing of the spell instead of showing it as separate in the same roll the way Beyond20 does - which is much more elegant a solution.
> 
> This still doesn't show Blessed Healer and makes you have to think about it, too, because you can't just take the Beyond20 showing of Disciple of Life and add that to yourself, you gotta go "Okay, 2 + spell level to me..." which is clearly a small problem or nonissue but...
> 
> Sometimes the "fix" seems lazy compared to the workaround.
> 
> TL;DR: we appreciate you @kakaroto

Yeah, when I was verifying my code stripping for this I had that same feeling, "this doesn't show anywhere near the info as we are in Beyond20...", sucks.

jkomg:
and I'm sorry I didn't give you credit for this code, @Aeristoka!

Aeristoka:
> and I'm sorry I didn't give you credit for this code, @Aeristoka!

No worries 😄 , I didn't write the original code, just cleaned it out when it became redundant 😭 .

kakaroto:
yeah, I know! I had the same feelings about it as well. Ideally, someone should poke ddb to say "put it as a separate healing like you can have a spell with multiple damages".

kakaroto:
It's definitely an improvement for users who aren't beyond20 users, but I also bet some people might still think they need to add it manually and not realize it's already included.

PR #729, opened 2021-03-26
Refactor of Attack Processing Extras

Aeristoka:


src/dndbeyond/content-scripts/character.js
```diff
@@ -1181,46 +1223,7 @@ function rollSpell(force_display = false, force_to_hit_only = false, force_damag
...
-            damage_types.push("Favored Foe");
-        }
-
-        // Warlock: Genie Patron - Genie's Wrath
-        if (character.hasClassFeature("Genie’s Vessel") && character.getSetting("genies-vessel", false) && to_hit != null) {
-            damages.push(character._proficiency);
-            damage_types.push("Genie's Wrath");
-        }
-    
+        handleSpecialSpells(spell_name, damages, damage_types, {spell_level: level, spell_source, castas}, to_hit);
```
kakaroto, line 1226: You removed the things below here (like hexblade's curse) but you don't call the `handleGenericAttacks` to get them added back, so they are technically gone now.
You might need to add a call to `handleGenericAttacks` but would need to make sure that the things in there would all apply for the use case here (i.e: spells), so that all the conditions are necessarily false apart from the 5 features below that you removed (i.e: sneak attack wouldn't apply because it's not a finesse weapon.. but would it still get triggerred by a "Ranged" spell? because it shouldn't, etc...).

kakaroto (commented):
incomplete review but hopefully enough to get you started again on a new round of fixes

src/dndbeyond/content-scripts/character.js
```diff
@@ -401,62 +401,36 @@ function rollItem(force_display = false, force_to_hit_only = false, force_damage
...
+
+        if (properties["Attack Type"] == "Ranged") {
+            handleSpecialRangedAttacks(damages, damage_types, properties, settings_to_change, to_hit);
         }
+
+        handleSpecialGeneralAttacks(damages, damage_types, properties, settings_to_change, to_hit);
+
+        // Should remain here, as they require a weapon you're holding
+        // -----------------------------------------------------------
+        // Paladin: Sacred Weapon
```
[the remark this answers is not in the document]
Aeristoka: OH SNAP!!!

That's a good idea.

src/dndbeyond/content-scripts/character.js
```diff
@@ -997,58 +758,115 @@ function rollAction(paneClass, force_to_hit_only = false, force_damages_only = f
...
-        character.getSetting("artificer-alchemical-savant", false) &&
-        damages.length > 0) {
-        const alchemical_savant_regex = /[0-9]+d[0-9]+/g;
-        for (let i = 0; i < damages.length; i++){
-            if ((damage_types[i] === "Acid" || damage_types[i] === "Fire" || damage_types[i] === "Necrotic" || damage_types[i] === "Poison") &&
-                alchemical_savant_regex.test(damages[i])) {
-                damages.push(`${character.getAbility("INT").mod < 2 ? 1 : character.getAbility("INT").mod}`);
-                damage_types.push("Alchemical Savant");
-                break;
+function handleSpecialSpells(spell_name, damages=[], damage_types=[], {spell_source="", spell_level="Cantrip", castas}={}, to_hit="") {
```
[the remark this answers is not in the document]
Aeristoka: I love learning new Javascript/Programming things... That is neat.

src/dndbeyond/content-scripts/character.js
```diff
@@ -873,84 +703,15 @@ function rollAction(paneClass, force_to_hit_only = false, force_damages_only = f
                     damage_types.push("Crimson Rite");
                 }
             }
```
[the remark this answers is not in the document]
Aeristoka: It'll go into the `handleSpecialItemAttacks` as you suggested above. It's one where the wording explicitly requires an item-type weapon.

src/dndbeyond/content-scripts/character.js
Aeristoka: Yeah, need to add those things back. It got to the point `handleGenericAttacks` become too bad of a case to handle, so I started to put them into `handleSpecialSpells` instead, but missed those.

Aeristoka: The only one I forgot was "Hexblade's Curse" 😄

src/dndbeyond/content-scripts/character.js
```diff
@@ -1136,6 +945,239 @@ function handleSpecialHealingSpells(spell_name, damages=[], damage_types=[], {sp
     }
 }
 
+function handleSpecialMeleeAttacks(damages=[], damage_types=[], properties, settings_to_change={}, to_hit, action_name="") {
```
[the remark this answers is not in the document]
Aeristoka: This was actually deliberate, as rollItem doesn't have `action_name`.

the rollItem call:
`handleSpecialMeleeAttacks(damages, damage_types, properties, settings_to_change, to_hit);`

the rollAction call:
`handleSpecialMeleeAttacks(damages, damage_types, properties, settings_to_change, to_hit, action_name);`

So, less consistent, but tailored to the situation.

src/extension/popup.js
```diff
@@ -220,6 +220,7 @@ function populateCharacter(response) {
         }
         if (response["actions"].includes("Channel Divinity: Sacred Weapon")) {
             e = createHTMLOption("paladin-sacred-weapon", false, character_settings);
+            options.append(e);
```
kakaroto, line 223: oups :D

src/dndbeyond/content-scripts/character.js
```diff
@@ -997,58 +998,124 @@ function rollAction(paneClass, force_to_hit_only = false, force_damages_only = f
...
-            }
+    if (character.hasClass("Warlock")) {
+        // Warlock: The Hexblade: Hexblade's Curse
+        if (damages.length > 0 &&
+            character.getSetting("warlock-hexblade-curse", false) &&
+            character.hasClassFeature("Hexblade’s Curse") &&
+            character._proficiency !== null) {
+            damages.push(character._proficiency);
+            damage_types.push("Hexblade's Curse");
+        }
```
kakaroto, line 1082: You can't have hexblade's curse here. It's already in the generic handler, so a warlock doing a melee attack with green flame blade means it gets added twice.

src/dndbeyond/content-scripts/character.js
```diff
@@ -997,58 +998,124 @@ function rollAction(paneClass, force_to_hit_only = false, force_damages_only = f
...
+            damage_types.push("Empowered Evocation");
         }
     }
 
+    // Protector Aasimar: Radiant Soul Damage
+    if (character.hasRacialTrait("Radiant Soul") &&
+        character.getSetting("protector-aasimar-radiant-soul", false)) {
+        damages.push(character._level);
+        damage_types.push("Radiant Soul");
+    }
```
kakaroto, line 1107: Same for radiant soul. Basically you should just be calling `handleSpecialGeneralAttack` in the roll spell attack function so radiant soul, and hexblade's curse and the other stuff gets handled automatically. You should never have an ability handled in two of the functions.

src/extension/popup.js
Aeristoka: Yeah, it was the best, I was like "Y u no test properly!" And then I found it...

PR #735, opened 2021-03-29
Move extra part of Planar Warrior into function

Aeristoka:


kakaroto:
this won't affect the "roll both' for versatile weapons... do we care?

Aeristoka:
> this won't affect the "roll both' for versatile weapons... do we care?

No, good point. There's a few things like that where the code just has to kind of "sticky" into certain places for reasons like that.

PR #738, opened 2021-03-30
Add Bard Jack of All Trades to Initiative

Aeristoka:
https://www.sageadvice.eu/2014/08/23/jack-of-all-trades-bard/


Aeristoka:
This is natively handled by DnDBeyond, so nevermind 😄

PR #742, opened 2021-03-30
Match Hunter's Mark behavior with Absorb Elements

Aeristoka:
Also moves `Absorb Elements` and `Hunter's Mark` to the top of `handleSpecialSpells` to make sure (more with `Hunter's Mark` that they operate ONLY on their initial block.

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -1002,6 +1002,25 @@ function rollAction(paneClass, force_to_hit_only = false, force_damages_only = f
 }
 
 function handleSpecialSpells(spell_name, damages=[], damage_types=[], {spell_source="", spell_level="Cantrip", castas}={}) {
+    // Handle special spells;
+    // Absorb Elements
+    if (spell_name == "Absorb Elements") {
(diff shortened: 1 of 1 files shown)


kakaroto:
I see. Makes sense for Absorb elements. But why is hunter's mark setting the damages array to 0? Also, you seem to have added hunter's mark, not moved it.

Aeristoka:
> I see. Makes sense for Absorb elements. But why is hunter's mark setting the damages array to 0? Also, you seem to have added hunter's mark, not moved it.

Blanking it out, same as Absorb Elements, then setting it to the generic instead.

Only Absorb Elements got moved.

kakaroto:
ha ha haaaaaaaai seee!
ok then. thanks!

Aeristoka:
> ha ha haaaaaaaai seee!
> ok then. thanks!

Context of conversation about this from Discord:
"Absorb Elements has like, 8 damages, right?
your handling for that clarifies it down to 1
because it's only the triggering type that matters
Hunter's Mark is a similar thing
it has 3 "possible" damage types
only 1 matters
and they're the same dice count and dice size
so why not handle it the same?"

PR #743, opened 2021-03-30
Arms of the Astral Self can make Unarmed Strikes

Aeristoka:


kakaroto:
such a beautifully simple change now! 😍

PR #744, opened 2021-03-30
Stop Hunter's Mark from acquiring extra damage

Aeristoka:
Example: Favored Foe
Shouldn't apply specifically to Hunter's Mark, because Hunter's Mark is already "extra damage" to an attack roll.


Aeristoka:
@kakaroto I'm figuring this might have more additions in the future, thus the commentary on it in Comments

PR #748, opened 2021-03-31
Add Utensils check in finding if Item is a Tool

Aeristoka:
Covers particularly "Cook's Utensils"

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -274,7 +274,8 @@ function isItemATool(item_name, source) {
         item_name.includes(" Kit") ||
         item_name.includes(" Supplies") || 
         item_name.includes(" Set") || 
-        item_name === "Wagon";
+        item_name === "Wagon" ||
+        item_name.includes(" Utensils");
 }
 function isItemAnInstruction(item_name, item_tags) {
     // Rhythm-Maker’s Drum, +1, +2, +3 don't have the tag


kakaroto:
I thought that was handled! 🤦‍♂️ guess I missed it!

Artemisthemp:
> I thought that was handled! 🤦‍♂️ guess I missed it!

I get a use tool when opening Cooking Utensils,

Aeristoka:
> > I thought that was handled! 🤦‍♂️ guess I missed it!
> 
> I get a use tool when opening Cooking Utensils,

Screenshot it then.

Artemisthemp:
> > > I thought that was handled! 🤦‍♂️ guess I missed it!
> > 
> > 
> > I get a use tool when opening Cooking Utensils,
> 
> Screenshot it then.

This is weird, I could have sworn there was a button earlier today and yet neither my Desktop or Laptop display it. 

So either the issue fix it self after a long time or I click Carpenter tools (most likely)

Aeristoka:
> > > > I thought that was handled! 🤦‍♂️ guess I missed it!
> > > 
> > > 
> > > I get a use tool when opening Cooking Utensils,
> > 
> > 
> > Screenshot it then.
> 
> This is weird, I could have sworn there was a button earlier today and yet neither my Desktop or Laptop display it.
> 
> So either the issue fix it self after a long time or I click Carpenter tools (most likely)

I'll agree with most likely the mis-click onto another tool. There was literally 0 handling for something name "x Utensil". My PR put some in.

kakaroto:
I would have switched the two lines so Wagon stays at the end and the 'includes' stay together, but it's a personal choice, not a bug and not a blocker. Thanks for the PR @Aeristoka.

PR #751, opened 2021-03-31
Special Handling for Spirit Guardians

Aeristoka:
Matches Absorb Elements + Hunter's Mark Handling (pare down to 1 Damage).

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -1021,6 +1021,15 @@ function handleSpecialSpells(spell_name, damages=[], damage_types=[], {spell_sou
         damage_types.push("Weapon Type");
     }
 
+    // Handle Spirit Guardians
+    if (spell_name == "Spirit Guardians") {
+        const dmg = damages[0];
+        damages.length = 0;
+        damage_types.length = 0;
+        damages.push(dmg);
+        damage_types.push("Radiant or Necrotic");
+    }
+
     // Artificer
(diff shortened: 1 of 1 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -1021,6 +1021,15 @@ function handleSpecialSpells(spell_name, damages=[], damage_types=[], {spell_sou
         damage_types.push("Weapon Type");
     }
 
+    // Handle Spirit Guardians
+    if (spell_name == "Spirit Guardians") {
+        const dmg = damages[0];
+        damages.length = 0;
+        damage_types.length = 0;
+        damages.push(dmg);
+        damage_types.push("Radiant or Necrotic");
```
kakaroto, line 1030: I'm slightly uneasy about these changes, and wondering if it shouldn't first check that `damages.length` is a specific length and with specific types, so we're not destroying some potential other damages that we wouldn't want to remove?   I could imagine spells being handled from within an item (like booming blade) or a custom spell with the same name or something.
I'm tempted to do `if (damages.length === 2 && damage_types[0] === "Radiant" && damage_types[1] === "Necrotic")` otherwise don't replace it all.
Same would apply to Absorb Elements and Hunter's Mark.
Thoughts?

Aeristoka: So just some nice extra code to cover those 3? That makes good sense, yeah.

PR #757, opened 2021-04-02
Parsing functions bugfixes

Aeristoka:
`handleSpecialGeneralAttacks` now checks, where appropriate for `to_hit != null`
Removed redundant checks in `handleSpecialMeleeAttacks` and `handleSpecialRangedAttacks` for `to_hit != null`, as that's handled in the function calling logic
Re-ordered some parts of `handleSpecialSpells` and `handleSpecialHealingSpells` so that features which work on ALL damage types now have their chance to do so.

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -326,8 +326,7 @@ function handleSpecialMeleeAttacks(damages=[], damage_types=[], properties, sett
 
     // Feats
     // Great Weapon Master Feat
-    if (to_hit !== null && 
-        character.getSetting("great-weapon-master", false) &&
+    if (character.getSetting("great-weapon-master", false) &&
         character.hasFeat("Great Weapon Master") &&
         (properties["Properties"] && properties["Properties"].includes("Heavy") ||
         action_name.includes("Polearm Master - Bonus Attack")) &&
@@ -352,8 +351,7 @@ function handleSpecialMeleeAttacks(damages=[], damage_types=[], properties, sett
 function handleSpecialRangedAttacks(damages=[], damage_types=[], properties, settings_to_change={}, {to_hit, action_name=""}={}) {
     // Feats
     // Sharpshooter Feat
-    if (to_hit !== null && 
-        character.getSetting("sharpshooter", false) &&
+    if (character.getSetting("sharpshooter", false) &&
         character.hasFeat("Sharpshooter") &&
         properties["Proficient"] == "Yes") {
         to_hit += " - 5";
@@ -377,7 +375,7 @@ function handleSpecialGeneralAttacks(damages=[], damage_types=[], properties, se
     // Class Specific
     if (character.hasClass("Cleric")) {
         // Cleric: Blessed Strikes
-        if ((item_name || action_name || (spell_name && spell_level.includes("Cantrip"))) &&
+        if ((((item_name || action_name) && to_hit != null) || (spell_name && spell_level.includes("Cantrip"))) &&
             character.hasClassFeature("Blessed Strikes") &&
             character.getSetting("cleric-blessed-strikes", false)) {
             damages.push("1d8");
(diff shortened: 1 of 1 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -326,8 +326,7 @@ function handleSpecialMeleeAttacks(damages=[], damage_types=[], properties, sett
 
     // Feats
     // Great Weapon Master Feat
-    if (to_hit !== null && 
-        character.getSetting("great-weapon-master", false) &&
+    if (character.getSetting("great-weapon-master", false) &&
```
kakaroto, line 329: why did you remove the check for to_hit in here?

src/dndbeyond/content-scripts/character.js
```diff
@@ -513,8 +514,7 @@ function handleSpecialWeaponAttacks(damages=[], damage_types=[], properties, set
 
     if (character.hasClass("Paladin")) {
         // Paladin: Sacred Weapon
-        if (to_hit !== null && 
-            character.getSetting("paladin-sacred-weapon", false)) {
+        if (character.getSetting("paladin-sacred-weapon", false)) {
```
kakaroto, line 517: same question about removing check for to_hit

kakaroto (commented):
brain = dead. will leave this comments and retry to review some other day.

src/dndbeyond/content-scripts/character.js
Aeristoka: There is no condition under which there isn't a `to_hit` on this now in the function call

src/dndbeyond/content-scripts/character.js
Aeristoka: There is no condition under which there isn't a `to_hit` on this now in the function call

src/dndbeyond/content-scripts/character.js
```diff
@@ -1131,36 +1129,55 @@ function handleSpecialSpells(spell_name, damages=[], damage_types=[], {spell_sou
...
-        for (let i = 0; i < damages.length; i++){
-            if (damage_types[i] === "Healing") {
-                damages.push("1d8");
-                damage_types.push("Enhanced Bond Healing");
-                break;
+    // Supreme Healing must ALWAYS be at the end, at it maxes all healing dice
+    if (character.hasClass("Cleric")) {
+        if (character.hasClassFeature("Supreme Healing")) {
+            for (let i = 0; i < damages.length; i++) {
+                if (damage_types[i] !== "Healing") continue;
```
kakaroto, line 1177: Shouldn't this check for `.includes("Healing")` because of things like "Enhanced Bond Healing" or "Alchemical Savant Healing" ?

src/dndbeyond/content-scripts/character.js
```diff
@@ -1249,6 +1258,11 @@ function rollSpell(force_display = false, force_to_hit_only = false, force_damag
         if (healing_modifiers.length > 0) {
             handleSpecialHealingSpells(spell_name, damages, damage_types, {spell_level: level, spell_source, castas});
         }
+        else {
+            to_hit = handleSpecialGeneralAttacks(damages, damage_types, properties, settings_to_change, {to_hit, spell_name, spell_level: level});
+        
+            handleSpecialSpells(spell_name, damages, damage_types, {spell_level: level, spell_source, castas});
+        }
```
kakaroto, line 1265: Why is `handleSpecialSpells` inside the `else` so it's only for non healing spells ? I think it should be outside the else (even if there is nothing that would apply for it in non-healing spells)
Also did the condition that made you call `handleSpecialGeneralAttacks` for non-hunters-mark not apply anymore ?

src/dndbeyond/content-scripts/character.js
```diff
@@ -1248,6 +1257,11 @@ function rollSpell(force_display = false, force_to_hit_only = false, force_damag
         if (healing_modifiers.length > 0) {
             handleSpecialHealingSpells(spell_name, damages, damage_types, {spell_level: level, spell_source, castas});
         }
+        else {
```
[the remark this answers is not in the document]
kakaroto: Yep, it's fine after you explained to me on Discord the reason for using the `else`. I just needed to understand it.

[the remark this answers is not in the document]
Aeristoka: But I already changed it 😭

src/dndbeyond/content-scripts/character.js
Aeristoka: Done

kakaroto:
oh yeah, much cool, much nice, much thanks!

PR #758, opened 2021-04-02
Add check for spell source for Empowered Evocation feature.

cashoes:
Check spell source for Empowered Evocation feature

Class: Wizard
Subclass: Evocation
Feature: Empowered Evocation (Lvl.10)

Currently checks that character has feature and spell is the proper school (Evocation), but not source of said spell. 

Feature reads:

Empowered Evocation
Beginning at 10th level, you can add your Intelligence modifier to one damage roll of any **wizard** evocation spell you cast.
 
Resolution:
Additionally check that spell_source includes "Wizard".


Aeristoka (approved):
Good catch! Thank you.

PR #762, opened 2021-04-04
GetHitCritCard Add render and content

captify-ochekhovskyi:
Little feature for your awesome plugin. 
That little script just add possibility to chose new home-brew with getting cards with effects when character get crit/fail .

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -63,7 +63,8 @@ const SRC_FILES = {
     roll20: [
         ...ROLL_RENDERER_DEPS,
         "src/roll20/renderer.js",
-        "src/roll20/content-script.js"
+        "src/roll20/content-script.js",
+        "src/common/d52cards.js"
     ],
     roll20_script: [
         "src/common/sandbox-header.js",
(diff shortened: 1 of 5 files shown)


Aeristoka:
Can you post Screenshots of what the results are from what you've added here?

captify-ochekhovskyi:
image

Screenshot from 2021-04-05 02-13-40
Screenshot from 2021-04-05 02-12-41

captify-ochekhovskyi:
function just parse results, get random chaise  and getting content based on types of damage. After that render result on Roll20 page.

kakaroto:
Hi @captify-ochekhovskyi,
Thank you for the PR, however, I cannot accept this change as it adds something very specific and opinionated. I cannot have a guarantee that every one of those 52 quotes are part of the public domain and aren't infringing on anybody's copyright, and their contents might not be suitable for every group. A more generic solution where users can define their own "roll table" of results would make more sense, but I would still not accept it as that would be a quite significant feature that I don't believe fits within the purpose of Beyond20, which remains to simplify rolling from D&D Beyond. 
Thanks for understanding.

captify-ochekhovskyi:
Ok. thank you for your time.

PR #774, opened 2021-04-17
Fix for Monster Display Feature on Non-Features

### LINKED ISSUE #773: Display NPC Features catches on Spell Levels as well
image

Aeristoka:
Fixes #773 
Might be a cleaner way to do this though

### DIFF
--- a/src/dndbeyond/base/monster.js
+++ b/src/dndbeyond/base/monster.js
@@ -401,34 +401,36 @@ class Monster extends CharacterBase {
             if (add_dice) {
                 let description = descriptionToString(action);
                 description = description.replace(/−/g, "-");
-                const roll_properties = this.buildAttackRoll(action_name, description);
-                if (roll_properties) {
-                    const id = addRollButton(this, () => {
-                        // Need to recreate roll properties, in case settings (whisper, custom dmg, etc..) have changed since button was added
-                        const roll_properties = this.buildAttackRoll(action_name, description);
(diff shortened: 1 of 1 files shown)


kakaroto:
Good catch, yeah. But this would be the wrong fix. There will be other situations where it can affect the NPC. Example: https://www.dndbeyond.com/monsters/annis-hag ("3/day each"), or I *would* have expected to see the same issue on each of the ray descriptions for https://www.dndbeyond.com/monsters/beholder but it's not the case, but I would expect some other monster having similar multi-paragraph descriptions of a single thing to be affected.
I think the best fix would be to only do it when the paragraph starts with a bold/italic element.
Also noticed the beholder "Eye Ray" legendary is missing the name because it's bold but not italic.

Aeristoka:
> Good catch, yeah. But this would be the wrong fix. There will be other situations where it can affect the NPC. Example: https://www.dndbeyond.com/monsters/annis-hag ("3/day each"), or I _would_ have expected to see the same issue on each of the ray descriptions for https://www.dndbeyond.com/monsters/beholder but it's not the case, but I would expect some other monster having similar multi-paragraph descriptions of a single thing to be affected.
> I think the best fix would be to only do it when the paragraph starts with a bold/italic element.
> Also noticed the beholder "Eye Ray" legendary is missing the name because it's bold but not italic.

So super duper glad they were consistent...

Yeah figured it might not be the whole thingy. You'll get that though I'm sure 😄

Issue #638, opened 2020-12-22
Maneuvers: Parry still not quite right in 2.3.0


aryairani:
I see two different ways to roll a Fighter's Parry maneuver, and I think each is wrong in a different way.

## Way 1
If I roll from the Reactions list, a `1d20 + DEX modifier` is rolled.  At Level 7, my superiority die is a d8, so this should be `1d8 + DEX modifier`.
![image](https://user-images.githubusercontent.com/538571/102942594-cadfc800-4483-11eb-8715-c0f750373daa.png)
![image](https://user-images.githubusercontent.com/538571/102942791-4c375a80-4484-11eb-856d-41f92b882f12.png)

## Way 2
If I roll from the "Maneuvers: Parry" sidebar, a `1d8` is rolled, with no DEX modifier:
![image](https://user-images.githubusercontent.com/538571/102942752-28741480-4484-11eb-82f4-7e3295afbcb2.png)
![image](https://user-images.githubusercontent.com/538571/102942804-58231c80-4484-11eb-8d9b-094dfe9b5c17.png)

kakaroto:
Parry isn't something that Beyond20 does on its own. It either : 
1 - Sees a `-1` in that description, independent of its context and rolls a `1d20 - 1` (just like you would see in a statblock `To hit: +3` and it rolls a `1d20 + 3`)
2 - Sees it's a Maneuver and rolls a superiority die, which is the 1d8.

There was an issue that was filed and fixed (#288) which adds special handling for Parry and Rally so it added the right modifiers, but it looks for the "Parry" action and in the screenshot you sent, it seems that D&D Beyond renamed it to "Maneuvers: Parry"? That would explain why it doesn't add the modifiers anymore.

tyler-macinnis:
This issue should be closable now, correct?

kakaroto:
Yep! Hadn't noticed it didn't auto-close.
Thanks!

PR #778, opened 2021-04-19
Fix for issue 638 - Fighter Maneuvers

tyler-macinnis:
Small fix for the Parry and Rally Maneuvers for Fighter
#638


kakaroto:
Thank you!

PR #779, opened 2021-04-20
Savage / Brutal Critical Fix

### LINKED ISSUE #305: Savage/Brutal Critical should differentiate between weapon damages
Weapon damage dice are the only ones that need to be rerolled. Do not take the highest dice among things like divine smite, sneak attack, etc...

tyler-macinnis:
Weapon dice are the first ones passed into Roll20 and Foundry so this gets that first dice size and uses it for the calculation.

Tested it with a 14 Barbarian / Champion Fighter 4 / Rogue 2 Build with Great Weapon Fighting Style.

The first critical damage from a dagger was:
- 1d4(2)+5=7 for the dagger damage
- 1d4(3)=3 for the dagger crit damage
- 1d6(3)=3 for the sneak attack damage
- 1d6(4)=4 for the sneak attack crit damage
- 2d4(1+2)=3 for the level 14 brutal critical damage from the dagger

It only rolls 2 extra d4s for the brutal critical like it should

The second critical damage from a greatsword was:
- 2d6(1(dropped and rerolled as 5)+2(dropped and rerolled as 2))+5=12 for the greatsword damage
- 2d6(2(dropped and rerolled as 2)+6) = 8 for the greatsword crit damage
- 2d6(2(dropped and rerolled as 5)+1(dropped and rerolled as 3)) = 8 for the level 14 brutal critical damage from the greatsword

It only roll 2 extra d6s for the brutal critical like it should

Brutal Crticial Savage Attacks Fix

I think this should fix #305 by making brutal critical just roll the damage dice for the weapon. Wasn't as easy to fix as the maneuvers, especially when I'm still learning javascript.

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -499,6 +499,8 @@ class Beyond20RollRenderer {
             const damage_types = request["damage-types"];
             const critical_damages = request["critical-damages"];
             const critical_damage_types = request["critical-damage-types"];
+            const brutal_critical_and_savage_attack_damages = request["brutal-and-savage-damages"];
+            const brutal_critical_and_savage_attack_damage_types = request["brutal-and-savage-damage-types"];
             if (request.name === "Chromatic Orb") {
                 const damage_choices = {}
                 const critical_damage_choices = {}
@@ -609,6 +611,11 @@ class Beyond20RollRenderer {
             }
             if (is_critical) {
                 const critical_damage_rolls = []
+                if (request["brutal-and-savage-damage-types"].length > 0){
+                    const WeaponDiceType = await this.queryDamageType(request.name, brutal_critical_and_savage_attack_damage_types);
+                    critical_damages.push(brutal_critical_and_savage_attack_damages[WeaponDiceType]);
+                    critical_damage_types.push(brutal_critical_and_savage_attack_damage_types[WeaponDiceType]);
+                }
                 for (let i = 0; i < (critical_damages.length); i++) {
                     const roll = this._roller.roll(critical_damages[i]);
                     critical_damage_rolls.push(roll);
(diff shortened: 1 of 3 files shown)


Aeristoka:
This is at least incomplete because it doesn't cover the Homebrew Max, as that's still getting the highest damage of all the crit_damages
image

I'll leave the rest to @kakaroto

kakaroto:
Thank you for submitting this.
Unfortunately, weapon damage would include things like fire damage from a flametongue sword (see https://rpg.stackexchange.com/questions/127429/can-a-barbarian-choose-which-weapon-die-to-add-for-his-brutal-critical) 
So the fix here would not be to use the first damage, but rather to separate which is damage from the weapon and which are damages from added abilities. That could probably be done by the caller, by saving the length of the damages array before calling the `handleSpecialAttacks` function (and similar) and passing that along to the `buildAttackRoll` so it knows how many items in the damages array it should look at before breaking. I think that would be my solution for it here.

tyler-macinnis:
I think you mean that it would include the **option** of the fire damage. I get what you are saying for a weapon like the Wand of Orcus. But in the end, the player gets to chose which additional weapon damage die (or dice) they want to roll. If that's the case, then one solution might to expand upon the _Brutal Critical/Savage Attacks: Roll extra die_ option when someone has that feature and allow them to select in it what weapon dice size they want rerolled.

tyler-macinnis:
Or, I could maybe have a popup like what happens when the player specifies advantage or disadvantage that asks them which dice they want to reroll.

tyler-macinnis:
I believe that I got it passed into buildAttackRoll, but I'm sure how to make it so the user can specify which damage dice they will want to use.

tyler-macinnis:
I got it now so that it only applies Brutal Critical/Savage Attacks to dice which are from the weapon; however, I'm not sure how to get it to have the user chose which dice they want to add and make it so that only the one they chose is added to the total. I've been testing with a Sun Blade Longsword simply because it has both Versatile and Finesse.
Screenshot 2021-04-24 161555

tyler-macinnis:
I think what needs to happen now, is before the total damage is returned, determine how many critical damages are from savage attacks and/or brutal, store those values and their damage types by moving them to another array, ask the user which damage type they want, and then only add that one to the total. @kakaroto What are your thoughts on this?

tyler-macinnis:
It should now correctly determine which weapon dice/damage type can be used for brutal critical/savage attacks and, when a critical hit has been rolled, ask the user which weapon dice/damage type they would like to add to their damage roll.

Screenshot 2021-04-26 183857
Screenshot 2021-04-26 183934

PR #780, opened 2021-04-22
Fix for Rogue Sneak Attack checks

atomicpeach:
The check for Psychic Blade was outside of the OR block and thus was not
checked against the sneak attack setting; therefore resulting in always
rolling sneak attack damage when using Psychic Blade. This fix shifts
the Psychic Blade check into the main OR block.


kakaroto:
Nice catch, thank you!

PR #790, opened 2021-05-06
Fix Brutal Critical for Perfect Rolls Hombrew Crit

### LINKED ISSUE #357: Brutal Critical doesn't work with Perfect Rolls Crit option

Aeristoka:
Fixes #357

### DIFF
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -101,7 +101,7 @@ function damagesToCrits(character, damages) {
 
 function buildAttackRoll(character, attack_source, name, description, properties,
                          damages = [], damage_types = [], to_hit = null,
-                         brutal = 0, force_to_hit_only = false, force_damages_only = false) {
+                         brutal = 0, force_to_hit_only = false, force_damages_only = false, {weapon_damage_length=0}={}) {
     const roll_properties = {
         "name": name,
         "attack-source": attack_source,
@@ -174,13 +174,20 @@ function buildAttackRoll(character, attack_source, name, description, properties
             if (brutal > 0) {
                 const rule = parseInt(character.getGlobalSetting("critical-homebrew", CriticalRules.PHB));
                 let highest_dice = 0;
+                let weapon_damage_counter = 0;
                 for (let dmg of damages) {
-                    const match = dmg.match(/[0-9]*d([0-9]+)/);
-                    if (match) {
-                        const sides = parseInt(match[1]);
-                        if (sides > highest_dice)
-                            highest_dice = sides;
+                    if(weapon_damage_counter < weapon_damage_length){
+                        const match = dmg.match(/[0-9]*d([0-9]+)/);
+                        if (match) {
+                            const sides = parseInt(match[1]);
+                            if (sides > highest_dice)
+                                highest_dice = sides;
+                        }
                     }
+                    else {
+                        break;
+                    }
+                    weapon_damage_counter++;
                 }
                 const isBrutal = character.hasClassFeature("Brutal Critical");
                 const isSavage = character.hasRacialTrait("Savage Attacks");
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -174,39 +174,29 @@ function buildAttackRoll(character, attack_source, name, description, properties
             if (brutal > 0) {
                 const rule = parseInt(character.getGlobalSetting("critical-homebrew", CriticalRules.PHB));
                 let highest_dice = 0;
-                let homebrew_max_damage = 0;
-                if (rule == CriticalRules.HOMEBREW_MAX) {
-                    let highest_damage = 0;
-                    for (let dmg of crit_damages) {
-                        if (dmg > highest_damage){
-                            highest_damage = dmg;
-                        }
-                    }
-                    homebrew_max_damage = brutal * highest_damage;
-                } else {
-                    for (let dmg of crit_damages) {
-                        const match = dmg.match(/[0-9]*d([0-9]+)/);
-                        if (match) {
-                            const sides = parseInt(match[1]);
-                            if (sides > highest_dice)
-                                highest_dice = sides;
-                        }
+                for (let dmg of damages) {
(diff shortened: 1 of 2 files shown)


Aeristoka:
@kakaroto I think this is slightly more elegant than the code I first submitted for this, but this does fix the issues. I'm not super happy with the if/else if logic, so if you can think of a prettier way to do that, please let me know.

tyler-macinnis:
Well this is definitely gonna have some conflicts with my fix so that non-homebrew brutal critical rolls gets the correct dice and lets the user pick which they want to use. #779

Aeristoka:
> Well this is definitely gonna have some conflicts with my fix so that non-homebrew brutal critical rolls gets the correct dice and lets the user pick which they want to use. #779

I still don't understand the desire to put in a pick for that. Why? In what situation would the user NOT want to have the highest possible die?

tyler-macinnis:
> > Well this is definitely gonna have some conflicts with my fix so that non-homebrew brutal critical rolls gets the correct dice and lets the user pick which they want to use. #779
> 
> I still don't understand the desire to put in a pick for that. Why? In what situation would the user NOT want to have the highest possible die?

Let's say you have a wand of Orcus. There's a lot of creatures immune to necrotic damage. I'd rather use the lower damage dice so that I could actually deal damage if I determine the creature is immune to necrotic.

It also accounts for the scenario where the highest dice can be one of two damage types. Let's say you have a Frost Brand Shortsword. It might be important to specify if the damage is gonna be cold or magical piercing depending on the enemy. Same thing for a Serpent's Fang Longsword. Need to know if you are dealing poison damage or magical slashing.

Aeristoka:
> > > Well this is definitely gonna have some conflicts with my fix so that non-homebrew brutal critical rolls gets the correct dice and lets the user pick which they want to use. #779
> > 
> > 
> > I still don't understand the desire to put in a pick for that. Why? In what situation would the user NOT want to have the highest possible die?
> 
> Let's say you have a wand of Orcus. There's a lot of creatures immune to necrotic damage. I'd rather use the lower damage dice so that I could actually deal damage if I determine the creature is immune to necrotic.
> 
> It also accounts for the scenario where the highest dice can be one of two damage types. Let's say you have a Frost Brand Shortsword. It might be important to specify if the damage is gonna be cold or magical piercing depending on the enemy. Same thing for a Serpent's Fang Longsword. Need to know if you are dealing poison damage or magical slashing.

That has strayed far too far into the "Automate everything" attitude, and diverges from Beyond20's goal of simplifying things for the users. YOU as the Player + your GM need to work that stuff out, not have Beyond20 do it for you.

tyler-macinnis:
I'm confused as to why then there is code in place to ask for the damage type for Chromatic Orb and Dragon's Breath as well as the damage dice for Toll the Dead.

Aeristoka:
> I'm confused as to why then there is code in place to ask for the damage type for Chromatic Orb and Dragon's Breath as well as the damage dice for Toll the Dead.

Because those explicitly ONLY have the damage type that you select there, and have a large number of possible damage types. There is nothing established when you roll it originally.

Aeristoka:
Now only considers weapon damages for Brutal Critical/Savage Attacks.

src/dndbeyond/base/utils.js
```diff
@@ -174,39 +174,36 @@ function buildAttackRoll(character, attack_source, name, description, properties
...
-                        if (dmg > highest_damage){
-                            highest_damage = dmg;
-                        }
-                    }
-                    homebrew_max_damage = brutal * highest_damage;
-                } else {
-                    for (let dmg of crit_damages) {
+                let weapon_damage_counter = 0;
+                for (let dmg of damages) {
+                    if(weapon_damage_counter < weapon_damage_length){
```
kakaroto, line 179: It's better to do `if (weapon_damage_counter >= weapon_damage_length) break;`
Also missing a space after the `if` (coding style)
also, I would check for weapon_damage_length since you init it at 0, so it should be `if (weapon_damage_length && weapon_damage_counter >= weapon_damage_length) break;`

src/dndbeyond/base/utils.js
```diff
@@ -174,39 +174,36 @@ function buildAttackRoll(character, attack_source, name, description, properties
...
                 const isSavage = character.hasRacialTrait("Savage Attacks");
                 if (highest_dice != 0) {
                     let brutal_dmg = `${brutal}d${highest_dice}`
-                    // Apply great weapon fighting to brutal damage dice
-                    if ((character.hasClassFeature("Great Weapon Fighting", true) || character.hasFeat("Great Weapon Fighting", true)) &&
-                        properties["Attack Type"] == "Melee" &&
-                        (properties["Properties"].includes("Versatile") || properties["Properties"].includes("Two-Handed"))) {
-                        brutal_dmg += "ro<=2"
+                    if (rule == CriticalRules.HOMEBREW_MAX) {
+                        crit_damages.push(damagesToCrits(character, [brutal_dmg])[0]);
```
kakaroto, line 197: noice

kakaroto (commented):
Thanks for your patience!
It looks good, just some small style issues, I'll fix locally and push the fixed changes.
Thanks also @tyler-macinnis for the initial implementation. I do agree with @Aeristoka on the fact that adding a prompt should really be a last resort, and choosing the highest dice is the most logical thing to do. The very rare use case that you described, while valid, would likely have to be something determined at the table between the DM and the player rather than trying to automate it in the extension.
Thanks you two!

PR #792, opened 2021-05-07
Add support for Ranger: Natural Explorer

### LINKED ISSUE #399: support for Natural Explorer
At level 1 Rangers get the following feature:

> Natural Explorer:
> You are particularly familiar with one type of natural environment and are adept at traveling and surviving in such regions. Choose one type of favored terrain: arctic, coast, desert, forest, grassland, mountain, swamp, or the Underdark. **When you make an Intelligence or Wisdom check related to your favored terrain, your proficiency bonus is doubled if you are using a skill that you’re proficient in.**

yadayadayada some extra fun stuff that don't really have to do with Beyond20. 

I was wondering if there could be a switch like for example Hexblade's Curse (https://i.imgur.com/viOTCQ8.png) to enable you being in your favored terrain, and thus doubling your proficiency bonus when making an Intelligence or Wisdom check (related to your favored terrain, which is probably most of them I do imagine).

Thanks in advance :)

Aeristoka:
Fixes #399

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -547,6 +547,12 @@ const character_settings = {
         "type": "bool",
         "default": false
     },
+    "ranger-natural-explorer": {
+        "title": "Ranger: Natural Explorer",
+        "description": "Your familiarity with certain regions aids in your exploration",
+        "type": "bool",
+        "default": false
+    },
     "ranger-gathered-swarm": {
         "title": "Ranger: Swarmkeeper: Gathered Swarm",
         "description": "Use your Gathered Swarm ability to add extra Force damage to your attacks",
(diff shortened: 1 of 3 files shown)


src/common/settings.js
```diff
@@ -547,6 +547,12 @@ const character_settings = {
         "type": "bool",
         "default": false
     },
+    "ranger-natural-explorer": {
+        "title": "Ranger: Natural Explorer",
+        "description": "Your familiarity with certain regions aids in your exploration",
```
kakaroto, line 552: I would change this into "Your familiarity with the current region" so it's clear that it's meant to boost the user rather than just say "I have this feature" (which is a mistake some users will inevitably make I think).
Not a blocker though, thx!

PR #795, opened 2021-05-11
Adjust padding to prevent jumping text in extension options on Firefox browsers

### LINKED ISSUE #760: Settings flicker caused by line break in firefox settings menu
**Describe the bug**
Opening the settings inside Firefox (about:addons) causes the Whisper rolls to DM section to flicker. It tries to create a line break but then instantly resets.

**To Reproduce**
1. Go to 'Beyond 20 settings in about:addons'
2. Click on 'open settings tab'
3. Scroll down to '...'

**Expected behavior**
Create a proper line break.

**Screenshots**
Flickering between the two states shown on the screenshots.
https://cdn.discordapp.com/attachments/746985780546633872/828250912904183838/unknown.png
https://cdn.discordapp.com/attachments/746985780546633872/828250938858537060/unknown.png

 - Browser Firefox
 - Version 87.0 64-Bit
 
Not sure if relevant but just in case.
 - Win 10
 - Desktop resolution 1920x1080

ljrafols:
Fixes #760 

Text no longer jumps around on Firefox 88.0.1 or Firefox Dev 89.0b10. Nothing changed as far as I can see on Chrome 90.0.4430.212, aside from the expected shifting of option description text. Tested on all browsers using a 1920x1080 monitor. 

There is a little bit of horizontally-scrollable area, but I am not sure if that existed prior to this issue or not.

### DIFF
--- a/options.css
+++ b/options.css
@@ -22,7 +22,7 @@
 	.list-group > .list-group-item > .list-content {
 		display: block;
 		padding: 10px 15px;
-		padding-right: 130px;
+		padding-right: 150px;
 		position: relative;
 	}
 	.list-group > .list-group-item > .list-content h4 {


kakaroto:
Thanks. I've tried this but I still see the flickering of the scroll bar unfortunately, even with this change. I'm testing on Firefox 88.0.1 though.
I'll poke at it and see if I can make a dent into this bug.

PR #824, opened 2021-06-11
FVTT CSS Styling Changes

LorduFreeman:
Updated the CSS styling to be close to the regular Foundry style.

**A few minor problems remain:**  (see image attached)
1. On a crit, two result rows for total damage are generated: Crit and Regular damage. It's not possible to style both using CSS without affecting other row displays.

 Suggested solution: If another result row displaying combined Crit and Regular damage would be added, the problem would be solved - and it would save some calculation time on the players ends. :) Alternatively, a custom id for applying CSS to the total damage result lines works as well.
2. The Beyond20 die icon needed to be resized to fit the smaller text style, becoming less recognizable. It needs to be recreated and maybe even simplified.

image


Aeristoka:
@LorduFreeman Noticed something important here:

You changed the CSS styling ONLY for the FVTT Module, you need to change it in the Beyond20 Chrome Extension as well so it matches up.

Beyond20 Chrome Extension has CSS for FoundryVTT in it itself, the FVTT Module styles things for people who don't have the Chrome extension to look nice, so both places need to be consistent to be considered as a good pull request.

Thanks!

LorduFreeman:
I see, I knew about the module for people not having the extension but  didn't know about all the spots I need to edit the CSS in. Thanks, I'll go change the Extension CSS as well!

kakaroto:
Thanks for submitting this. It looks much nicer this way I believe.
I'm a bit unclear on the critical damage issue you're talking about though. It sounds like you're saying that the last line itself gets increased size? Would it help if we also added a class to those "total damage" divs so you can target that class in the css instead of targetting the last-child (which I assume is what you did) ? 
I wanted to see what happens if you have an attack that deals multiple damages + extra damages from special feats like this : 
image

Or the case of a versatile weapon which would show like this : 
image
Or with crit : 
image

I haven't had a chance to actually pull in your changes to test, but thankfully @Aeristoka did as we were just discussing it and here's the before/after : 
image

image

I like the styling quite a bit more, though I'm unsure about the smaller font size for the damages, because my eyes are old and I don't like to squint, hehe.

Note: I saw you updated the beyond20.zip file, please revert that, as that's the zip that gets downloaded when someone installs the module, it should only get updated during official released after the module version is incremented.

Thanks!

LorduFreeman:
I was unsure about the zip file so roger that, will revert :)

As for the last lines, you are on point: My css targets the last-child. A class for the total damage rows would solve this formatting issue nicely. Another class for total crit-damage would be nice as well so these can be styled too, giving them some kind of fancy look to highlight the extra crit damage.

As for the font size, I agree. I tried to go for a more compact look but in terms of accessibility it is a bad choice, so I will revert to the regular size. That said, the regular Foundry item description text on chat cards is just 12px and I set the Beyond20 descriptions to the same size to fit the look - should this be reverted as well?

Aeristoka:
> I was unsure about the zip file so roger that, will revert :)
> 
> As for the last lines, you are on point: My css targets the last-child. A class for the total damage rows would solve this formatting issue nicely. Another class for total crit-damage would be nice as well so these can be styled too, giving them some kind of fancy look to highlight the extra crit damage.
> 
> As for the font size, I agree. I tried to go for a more compact look but in terms of accessibility it is a bad choice, so I will revert to the regular size. That said, the regular Foundry item description text on chat cards is just 12px and I set the Beyond20 descriptions to the same size to fit the look - should this be reverted as well?

Like the "Ridiculous Greatsword" text? No, keep that, it looks great.

LorduFreeman:
> Like the "Ridiculous Greatsword" text? No, keep that, it looks great.

No, the text that appears if you open the chat card (click the ▶ button reveals the description text).

kakaroto:
> Like the "Ridiculous Greatsword" text? No, keep that, it looks great.

I think they meant the expanded description of it. It would likely effect spell cards more, like in the first screenshot in the PR. That's why the beyond20 dice icon has to be made smaller.
I don't mind the smaller dice icon, it serves its purpose well enough, but yeah, if the text is 12 pixels, that's small. I thought 16 pixels was the norm. 
I'd need to see it to judge, and not having time at the moment to poke at it in a local env, so will check back on that later and give you an answer.

While I like the idea of being closer to the Foundry style (I don't particularly like it, but the changes here do look nicer from what I had), I don't think we need to be 100% the same, especially if it's something that affects accessibility.

LorduFreeman:
I agree, if adapting the core style means sacrificing accessibility (and 12px is too small) I'm quite fine with not adapting to it in this case, despite the visual differences that creates.

LorduFreeman:
chatcardsb20

Whooping a new preview in here, files should be up soon. + another example of the last-child styling issue, visible with the Eldritch Blast damage roll. As for legibility/font size, this is much easier to read.

Another question, is there an easy way to revert to the old beyond.zip? For now I just grabbed the old one and replaced it in my next commit.

Aeristoka:
Looking better:
image

LorduFreeman:
Anything left to do before this can get added?

Aeristoka:
> Anything left to do before this can get added?

Fair amount of review + Testing, and if I remember right, adding some more classes to let things be prettier.

LorduFreeman:
> 
> Fair amount of review + Testing, and if I remember right, adding some more classes to let things be prettier.

I see, looking forward to those upcoming classes :)

kakaroto:
I just pushed a commit that adds the classes to the html to let you tweak it further.
You can see the extra classes `beyond2-total-damage` and `beyond20-critical-damage` and I also added `beyond20-healing` in case we want to do something special with critical/healing damages at some point : 
image


This is btw how it looks in the DDB page itself. It looks pretty nice with the green tint to it :)
image

Aeristoka:
> I just pushed a commit that adds the classes to the html to let you tweak it further.
> You can see the extra classes `beyond2-total-damage` and `beyond20-critical-damage` and I also added `beyond20-healing` in case we want to do something special with critical/healing damages at some point :
> image
> 
> This is btw how it looks in the DDB page itself. It looks pretty nice with the green tint to it :)
> image

The roll numbers look weird though... like they're a different font... Same with the item name...

LorduFreeman:
Niice. Now the only thing I'm wondering is: Why does the Sneak Attack Damage number get the bold formatting? Same for the Necrotic Damage in the Rediculous Greatsword, looks like I need to check that.

> The roll numbers look weird though... like they're a different font... Same with the item name...

That's because I have changed the font-family for these regions (Modesto Condensed and Signika, the core fonts), while Foundry provides them when running in browser or app, these are non-existent outside of Foundry. So when rolling on DnDBeyond Palatino Linotype or a generic serif are used (the defined fallbacks font).

Aeristoka:
> Niice. Now the only thing I'm wondering is: Why does the Sneak Attack Damage number get the bold formatting? Same for the Necrotic Damage in the Rediculous Greatsword, looks like I need to check that.
> 
> > The roll numbers look weird though... like they're a different font... Same with the item name...
> 
> That's because I have changed the font-family for these regions (Modesto Condensed and Signika, the core fonts), while Foundry provides them when running in browser or app, these are non-existent outside of Foundry. So when rolling on DnDBeyond Palatino Linotype or a generic serif are used (the defined fallbacks font).

So maybe a good idea to let the Beyond20 CSS have a different font? And then the FVTT Module have the nice FoundryVTT Font? Or is there a way we can get Modesta Condensed and Signika loaded up @kakaroto ?

LorduFreeman:
> So maybe a good idea to let the Beyond20 CSS have a different font? And then the FVTT Module have the nice FoundryVTT Font? Or is there a way we can get Modesta Condensed and Signika loaded up @kakaroto ?

The easiest solution would be defining fallback fonts that are present on every system, else you'd have to look into loading fonts with the extension first. The Foundry-Fonts need to be licensed if I remember correctly.

I found a little oversight: The Critical Total Damage number does not get the `.beyond20-roll-total` class.

LorduFreeman:
image

Another issue I noticed: Without last-child styling some other rolls need a class as well: Regular weapon attacks and spells just display several `beyond20-roll-result`.

LorduFreeman:
Thinking about how to restyle the total damage/healing results as well:
image 
image

LorduFreeman:
While I was add it and because styling is partially linked, I changed the popup styles on Beyond when no VTT is found as well. 
damages

Now the only thing still missing is formatting for regular non-total damage. + the missing class for total critical damage numbers.

LorduFreeman:
What's the current status on this? Is it going to get merged or is something still missing?

kakaroto:
> What's the current status on this? Is it going to get merged or is something still missing?

I don't think anything is missing. It is very likely to be merged, I just need to find the time to look at it further, test and merge it.
I generally have very little free time, so I do most Beyond20 stuff all at once rather than being very reactive at all times.
Thanks for being patient :)

kakaroto:
Thanks! I just managed to test this, and it's looking really good, thanks for making this much needed change!
I did have a couple of small issues : 
1 - non consistent indentation in the css files
2 - resizing the custom icon needs to be restricted to that 'details' section of a rendered roll, otherwise, it will also affect the custom dice icons in a monster's statblock or in class feature descriptions, but we want to keep it at that 24x24 size there (or 16x16, I don't remember). `img.ct-beyond20-custom-icon` into `.beyond20-message details img.ct-beyond20-custom-icon`

I think that's all that's needed from what I've seen. I'll be making a release tomorrow, so I'll fix the above myself before merging if you don't get to it first.
Thanks again!

LorduFreeman:
1:fixed indentation (using prettier.io)
2:fixed the die icon resizing using your styling.

I decided to add back `.beyond20-roll-result:last-child` for styling single-damage attacks, to highlight damage:
image

Applying last-child styling fixes the following issue as well:
The total critical damage's tooltip number is missing its `.beyond20-roll-total` tag, resulting in a non-bold number without last-child styling applied:
image

kakaroto:
Awesome, thanks. I'll merge it now, and try to get the release out within the next hour or so.

Aeristoka:
@LorduFreeman One minor gripe here:
image

See the Weapon Name + Roll here:
image

They still need a fallback font.

LorduFreeman:
Indeed, should have given that a little more love. Proposal: Using the Roboto fonts DnDBeyond uses:
image

Aeristoka:
> Indeed, should have given that a little more love. Proposal: Using the Roboto fonts DnDBeyond uses:
> image

I think that looks great!

kakaroto:
Looks great indeed! ♥

LorduFreeman:
I created a new pull request for these minor changes. At some point in the future I'd like to further adjust the look to fit the DDB look more tightly as well.

Aeristoka:
> 1:fixed indentation (using prettier.io)
> 2:fixed the die icon resizing using your styling.
> 
> I decided to add back `.beyond20-roll-result:last-child` for styling single-damage attacks, to highlight damage:
> image
> 
> Applying last-child styling fixes the following issue as well:
> The total critical damage's tooltip number is missing its `.beyond20-roll-total` class, resulting in a non-bold number (without last-child styling applied):
> image

Is this something that @kakaroto still needs to fix? to make this cleaner?

LorduFreeman:
> > The total critical damage's tooltip number is missing its `.beyond20-roll-total` class, resulting in a non-bold number (without last-child styling applied):
> > image
> 
> Is this something that @kakaroto still needs to fix? to make this cleaner?

Yes, the class needs to be added so that the damage number is formatted correctly. I can try do do it myself as well, would be useful so I can test it myself before submitting a pull request update.

PR #830, opened 2021-06-15
Implement Warlock: The Undead: Grave Touched

### LINKED ISSUE #828: Support Warlock: The Undead - Grave Touched feature
Support for the new classes in VAN RICHTEN’S GUIDE TO RAVENLOFT

Not sure this one is doable because it is based on the damage die of your attack, but wanted to put it out here in case.
Character
https://www.dndbeyond.com/profile/echmyr/characters/52375493/builder#/ability-scores/manage

Warlock: The Undead

Grave Touched
6th-level Undead feature

Your patron’s powers have a profound effect on your body and magic. You don’t need to eat, drink, or breathe.

In addition, once during each of your turns, when you hit a creature with an attack roll and roll damage against the creature, you can replace the damage type with necrotic damage. **While you are using your Form of Dread, you can roll one additional damage die when determining the necrotic damage the target takes.**

Aeristoka:
Fixes #828

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -601,6 +601,12 @@ const character_settings = {
         "type": "bool",
         "default": false
     },
+    "warlock-grave-touched": {
+        "title": "Warlock: The Undead: Grave Touched",
+        "description": "Your attacks become necrotic, and apply extra damage",
+        "type": "bool",
+        "default": false
+    },
     "wizard-bladesong": {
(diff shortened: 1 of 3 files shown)


kakaroto:
I'm unsure about the rationale of having it auto disable itself... It feels to me like it's more of a sneak attack option, no? Also, if we decide to go with auto-disable, it should have "(Applies to next roll only)" like with sharpshooter.

PR #832, opened 2021-06-17
Implement Wizard: War Magic: Durable Magic

### LINKED ISSUE #831: Request: War Magic Wizard's Durable Magic
Currently there is no way to toggle the Durable Magic feature, but I think it probably could be considered. The feature has two parts:
Beginning at 10th level, the magic you channel helps ward off harm. While you maintain concentration on a spell, you have a +2 bonus to **AC** and **all saving throws**.
The AC part can't be added, but I believe the saving throw part probably would be (similar to Bladesinging Wizard's Bladesong).

I would like to see a feature toggle for Durable Magic that can be toggled when concentrating.

People can easily add +2, its simple addition, but I do believe the bonus can easily be forgotten. Besides, I don't think this feature is out of line with other nice QoL features Beyond20 has.

**Additional context**
N/A

Aeristoka:
Adds +2 to all saving throws with toggle
Used when concentrating on spells
Fixes #831

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -175,6 +175,7 @@ function rollAbilityOrSavingThrow(paneClass, rollType) {
     if (character.hasClassFeature("Durable Magic") && character.getSetting("wizard-durable-magic", false) &&
         rollType == "saving-throw") {
         modifier = parseInt(modifier) + 2;
+        modifier = modifier >= 0 ? `+${modifier}` : `${modifier}`;
         roll_properties["modifier"] = modifier;
     }
     // Fey Wanderer Ranger - Otherworldly Glamour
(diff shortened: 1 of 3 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -171,6 +171,12 @@ function rollAbilityOrSavingThrow(paneClass, rollType) {
             roll_properties["modifier"] = modifier;
         }
     }
+    // Wizard - War Magic - Saving Throw Bonus
+    if (character.hasClassFeature("Durable Magic") && character.getSetting("wizard-durable-magic", false) &&
+        rollType == "saving-throw") {
+        modifier = parseInt(modifier) + 2;
+        roll_properties["modifier"] = modifier;
```
kakaroto, line 178: The modifier has to be a string and needs to start with `+` since it gets appended to the formula, that's what this line (from other similar places) is for : 
```js
            modifier = modifier >= 0 ? `+${modifier}` : `${modifier}`;
```

kakaroto:
Thanks!

PR #835, opened 2021-06-18
Fix bug in RollRenderer that swapped ro -> r

Aeristoka:
Old Behavior
image

New Behavior
image


Aeristoka:
@kakaroto thanks for the education on what this actually does, closing, as this is in error.

PR #842, opened 2021-07-03
Fix for Spell Damage Types now being LowerCase

### LINKED ISSUE #841: Flames of Phlegethos
Recently after a new update, the Flames of Phlegethos Tiefling Feat no longer automatically rerolls 1's on a fire damage in a spell. I don't know if this bug affects other similar Feats like Great Weapon Master, but it most definitely affects this current Feat.
Problemo

Aeristoka:
Fixes #841
This is needed because DnDBeyond recently changed all Spell Damage Types
Used to be properly Capitalized "Fire", now are all like "fire"

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -646,6 +646,10 @@ function addCustomDamages(damages, damage_types) {
     }
 }
 
+function capitalize(str) {
+    return str.replace(/\b\w/g, (c) => c.toUpperCase());
+}
+
 function rollItem(force_display = false, force_to_hit_only = false, force_damages_only = false, spell_group = null) {
     const prop_list = $(".ct-item-pane .ct-property-list .ct-property-list__property,.ct-item-pane .ddbc-property-list .ddbc-property-list__property");
     const properties = propertyListToDict(prop_list);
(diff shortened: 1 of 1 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -1160,7 +1160,7 @@ function handleSpecialSpells(spell_name, damages=[], damage_types=[], {spell_sou
             break;
         }
     }
-    if (elementalAffinity && damage_types.includes(elementalAffinity)) {
+    if (elementalAffinity && damage_types.includes(elementalAffinity.toLowerCase())) {
```
kakaroto, line 1163: `damage_types` would have the non lower case type, I believe, so checking if the lowercase elemental affinity is included in `damage_types`  (which wouldn't be lower case) might fail.
Solution would have to be `damage_types.some(type => type.toLowerCase() === elementalAffinity.toLowerCase())`

kakaroto:
I think if damage types may be all lower case now in some instances, then it might be worth doing a 'capitalize' on it, so `fire` becomes `Fire` when displayed in the VTT.

kakaroto:
Actually.. if we do the capitalize on the damage types, then we wouldn't need to uglify all the code with `toLoweCase` 🤔

src/dndbeyond/content-scripts/character.js
```diff
@@ -1300,6 +1310,11 @@ function rollSpell(force_display = false, force_to_hit_only = false, force_damag
             damage_types.push(dmgtype);
         }
 
+        addCustomDamages(damages, damage_types);
+
+        // Capitalize all Damage Types to ensure consistency for later processing
+        damage_types = damage_types.map(t => capitalize(t));
+
```
kakaroto, line 1317: you moved the "addCustomDamages" from after special handling to before, which I think was done by mistake and could cause issues.

PR #843, opened 2021-07-03
Clean Up of Bogus Damage Types (as in "--") in rollItem

### LINKED ISSUE #233: Arrow of Slaying rolls incorrectly
An Arrow of Slaying (rolled only from a PC Character sheet) rolls incorrectly

image

Perhaps improperly parsing the attack part of the block for the Arrow of Slaying? Specifically happens when clicking the "Beyond 20" button.

image

Aeristoka:
Fixes #233

@kakaroto let me know what you think

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -695,7 +695,7 @@ function rollItem(force_display = false, force_to_hit_only = false, force_damage
                     character.getSetting("ranger-planar-warrior", false))
                     damage_type = "Force";
 
-                if (versatile_damage != "") {
+                if (versatile_damage != "" && damage_type != "--") {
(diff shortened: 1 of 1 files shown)


kakaroto:
Humm.. I'm thinking it's better to be generic about it, and simply not add any damage with value `--`, which we could do when parsing the `Damages` list itself.
I'm sure someone will have a custom weapon where they didn't homebrew it correctly and it has a `--`  in it too.
What do you think ?

Aeristoka:
> Humm.. I'm thinking it's better to be generic about it, and simply not add any damage with value `--`, which we could do when parsing the `Damages` list itself.
> I'm sure someone will have a custom weapon where they didn't homebrew it correctly and it has a `--` in it too.
> What do you think ?

Seems like a wise approach, yes.

kakaroto:
👨‍🍳

PR #850, opened 2021-07-19
Remove TempHPMax Syncing to FoundryVTT

Aeristoka:
This is due to "tempmax" being poorly documented in FoundryVTT.

Prior belief: "tempmax" = "The Maximum Temporary HP for a character"
Corrected: "tempmax" = "A temporary addition to the Maximum HP for a character"

Evidence:
https://gitlab.com/foundrynet/dnd5e/-/issues/828
Aid (inside FoundryVTT, with no DnDBeyond/Beyond20 help) should be added to TempMax
(Beyond20 already syncs a higher max HP via the "HP Max", so we shouldn't touch this)
image

Visual Appearance of the bars:
https://gitlab.com/foundrynet/dnd5e/-/issues/306
We want it to look like the second character in the screenshot, Temp HP overlaying the Current HP, not the 5th character in the screenshot.

### DIFF
--- a/src/fvtt/page-script.js
+++ b/src/fvtt/page-script.js
@@ -413,7 +413,7 @@ function updateHP(name, current, total, temp) {
 
     const tokens = canvas.tokens.placeables.filter((t) => t.owner && t.name.toLowerCase().trim() == name);
 
-    const dnd5e_data = { "data.attributes.hp.value": current, "data.attributes.hp.temp": temp, "data.attributes.hp.max": total, "data.attributes.hp.tempmax": temp }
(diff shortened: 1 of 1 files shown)


kakaroto:
argg.. *grumble*, *grumble*... that makes sense actually, since there is no such concept as a "maximum temp HP", you just either have temp HP or you don't.
Thanks!

Aeristoka:
> argg.. _grumble_, _grumble_... that makes sense actually, since there is no such concept as a "maximum temp HP", you just either have temp HP or you don't.
> Thanks!

That's why I gave you pretty pictures to look at along with the description text.

kakaroto:
picture were pretty! Also gitlab link was helpful!

PR #855, opened 2021-08-07
Fix bug in Custom Skills that Query for Ability

Aeristoka:
`prof_val` was cast as a String, causing the `mod += prof_val;` later in the code to concatenate a string rather than add Integers


kakaroto:
looks good. thanks! Not sure why I used `""` previously 🤷

Aeristoka:
> looks good. thanks! Not sure why I used `""` previously 🤷

🐟 🧠 ?

PR #861, opened 2021-08-13
Implement Damage Type Query for "Spirit Shroud" and "Destructive Wave" Spells

### LINKED ISSUE #860: Macro to choose type of damage for Spirit Shroud like Chromatic Orb has?
I was wondering if Spirit Shroud could have some macro like Chromatic Orb where it asks what type of damage you want to deal. Currently it displays al the three different types of damages separated, which is still useful, but since Chromatic Orb has this feature, maybe Spirit Shroud which has a similar behavior could have it too.

Aeristoka:
Brings spell into line with Chromatic Orb Handling

Fixes #860

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -551,6 +551,16 @@ class Beyond20RollRenderer {
             } else if (request.name == "Toll the Dead") {
                 const ttd_dice = await this.queryGeneric(request.name, "Is the target missing any of its hit points ?", { "d12": "Yes", "d8": "No" }, "ttd_dice", ["d12", "d8"]);
                 damages[0] = damages[0].replace("d8", ttd_dice);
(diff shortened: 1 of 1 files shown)


kakaroto:
Nice, thanks.
Starting to seem repetitive, that function could likely be extracted into a function that does it... the only difference in the code is just the array of the damage types it's replacing/querying for, right? 
It's not a necessity though so approving this one! thx

Aeristoka:
> Nice, thanks.
> Starting to seem repetitive, that function could likely be extracted into a function that does it... the only difference in the code is just the array of the damage types it's replacing/querying for, right?
> It's not a necessity though so approving this one! thx

To an extent, yes, but Destructive Wave in example only has two choices, out of the 3 damage types (1 damage type always, choice of the second between too).

Probably could be a function, don't have the brain energies to work it right now, but I'll think on it.

PR #870, opened 2021-09-23
Adds support for Making Human - Artisan's Intuition

Ustice:
Custom skills are a great way to represent tool proficiencies. This PR adds support for the Mark of Making Human - Artisan's Intuition for any tool proficiency.

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -164,7 +164,7 @@ async function rollSkillCheck(paneClass) {
     }
 
     // Mark of Making Human - Artisan's Intuition
-    if (character.hasRacialTrait("Artisan’s Intuition") && skill_name == "Arcana") {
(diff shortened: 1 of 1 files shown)


Aeristoka:
I'll leave the final say to @kakaroto , but I'm very much of the opinion that this is what the hotkeys are for. Beyond20 isn't meant to do all of everything for you, but to make easily automatable things easier, and especially not to cater to very specific custom things.

kakaroto:
Thanks for the PR @Ustice!
While I agree that Beyond20 isn't meant to do everything for you, I don't agree that this falls outside the scope of what Beyond20 does. Considering we already support the extra dice for specific skill checks, then adding the check for those instruments at the same time makes sense to me.
I would say however that I would have used `/i` flag for case insensitive regexp instead of using `.toLowerCase()`

Aeristoka:
It doesn't check for instruments though, it checks for very specifically named Custom Skills that users have to make themselves @kakaroto .

kakaroto:
oh, I've misread the patch then. 
Yeah, that's not really in the scope of B20 in that case in my opinion. But I would say that rolling the actual instrument is where it should be, and (now I see) that is already supported in the original code.

Issue #877, opened 2021-10-03
Clean hidden monster and attack name


kiloforce:
**Is your feature request related to a problem? Please describe.**
When using the "Hide monster and attack name", the questions marks clutter up the display.
![image](https://user-images.githubusercontent.com/433058/135737089-143cfb8f-7cb3-4cd3-b5e5-e00d9c1fab97.png)

**Describe the solution you'd like**
Would simply like the question marks removed, to clean up the display

Original Display:
![image](https://user-images.githubusercontent.com/433058/135737097-4a6d281c-34c2-42b9-9a90-cbb34cf98322.png)

Updated Clean Display:
![image](https://user-images.githubusercontent.com/433058/135737103-d5f46ac8-b0ce-42a7-9a34-d66bf06afc74.png)

**Proposed Patch**
```
diff --git a/src/roll20/content-script.js b/src/roll20/content-script.js
index 03ac795..dc641f1 100644
--- a/src/roll20/content-script.js
+++ b/src/roll20/content-script.js
@@ -176,14 +176,14 @@ function template(request, name, properties) {

     if (request.whisper == WhisperType.HIDE_NAMES) {
         if (properties["charname"])
-            properties["charname"] = "???"
+            properties["charname"] = ""
         // Take into account links for disabled auto-roll-damages option
         if (properties["rname"])
             properties["rname"] = properties["rname"].includes("](!") ?
-                properties["rname"].replace(/\[[^\]]*\]\(\!/, "[???](!") : "???";
+                properties["rname"].replace(/\[[^\]]*\]\(\!/, "[???](!") : "";
         if (properties["rnamec"])
             properties["rnamec"] = properties["rnamec"].includes("](!") ?
-                properties["rnamec"].replace(/\[[^\]]*\]\(\!/, "[???](!") : "???";
+                properties["rnamec"].replace(/\[[^\]]*\]\(\!/, "[???](!") : "";
         delete properties["description"];
     }
     for (let key in properties)
```

kakaroto:
Going to make that into an option (like the 'crit prefix' and 'components prefix' as a field to enter the replacement text with `???` as default. And going to put that as an advanced option (dependent on #896). This way, we'll both be happy with the PR and be able to merge it.

PR #878, opened 2021-10-03
Clean hidden monster and attack name

kiloforce:
Would be a solution for Issue #877


kakaroto:
Between these two options...
image
image
I'm still really unsure for which is better. I think it's a personal preference thing 🤔 
I can see the appeal of having a "cleaner look", but I also think the `???` have meaning here by saying "an unknown attack from an unknown attacker" rather than just "dice were rolled"
so.... I still don't know. 
I've just created #896 which I think might be a solution so that everybody can be happy with the results.
Leaving this open for now, but unmerged.

Thanks for the PR!

kiloforce:
+1 for #896

PR #885, opened 2021-10-22
Elemental spell handling

Aeristoka:
@kakaroto Also thinking maybe we should switch Spirit Guardians to this method. Your thoughts?

### DIFF
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -144,6 +144,106 @@ async function buildAttackRoll(character, attack_source, name, description, prop
     if (damages.length > 0) {
         roll_properties["damages"] = damages;
         roll_properties["damage-types"] = damage_types;
+        
+        if (roll_properties.name === "Chromatic Orb") {
+            const damage_choices = {}
+            // const critical_damage_choices = {}
+            for (let dmgtype of ["Acid", "Cold", "Fire", "Lightning", "Poison", "Thunder"]) {
+                let idx = damage_types.findIndex(t => t === dmgtype);
+                damage_choices[roll_properties["damage-types"].splice(idx, 1)[0]] = damages.splice(idx, 1)[0];
+                // idx = roll_properties["critical-damage-types"].findIndex(t => t === dmgtype);
+                // if (idx >= 0)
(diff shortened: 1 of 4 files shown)


kakaroto:
> @kakaroto Also thinking maybe we should switch Spirit Guardians to this method. Your thoughts?

I thought you didn't like prompts? 🙄 
The issue with moving Spirit Guardians is that the roll renderer is only used with digital dice or foundry, not when using Roll20 and rolling natively. On the other hand, we do use it for the other stuff already... 
I like having a single way of doing things, so nice work there. I do think if we can have the query happening on the DDB side, it would be beneficial as we'd not have to worry about a query on roll20/foundry, and it would work regardless of using roll renderer or not.
Also, seeing how the 5 special cases are the exact same code right now, that little bit of code could be extracted in a function that takes `name` and `types_list` and does its thing on the damages/damage_types.

kakaroto:
Any update on this btw? Are you changing it to do the prompt on the DDB side or do you want me to prior to merging?

Aeristoka:
> Any update on this btw? Are you changing it to do the prompt on the DDB side or do you want me to prior to merging?

Feel free to if you'd like, I don't know what my availability to work on this will look like coming up. Otherwise, I'll take a look at it when I'm able to.

Aeristoka:
@kakaroto ready for review now I believe

src/dndbeyond/base/utils.js
```diff
@@ -144,6 +144,91 @@ async function buildAttackRoll(character, attack_source, name, description, prop
     if (damages.length > 0) {
         roll_properties["damages"] = damages;
         roll_properties["damage-types"] = damage_types;
+        
+        if (roll_properties.name === "Chromatic Orb") {
+            const damage_choices = {}
+            for (let dmgtype of ["Acid", "Cold", "Fire", "Lightning", "Poison", "Thunder"]) {
+                let idx = damage_types.findIndex(t => t === dmgtype);
+                damage_choices[roll_properties["damage-types"].splice(idx, 1)[0]] = damages.splice(idx, 1)[0];
```
kakaroto, line 152: can use `damage_types` here directly instead of `roll_properties`

kakaroto (approved):

src/dndbeyond/base/utils.js
Aeristoka: I think that was a holdover from something else...

PR #895, opened 2021-11-22
Implement Warlock: The Celestial: Radiant Soul

### LINKED ISSUE #894: Celestial Warlock Radiant Soul (NOT aasimar radiant soul) doesn't seem to be implemented
After not playing her since before the pandemic, I finally converted my celestial warlock from paper to dndbeyond. Upon testing with sacred flame and wall of fire it appears that my charisma modifier is not being added to damage. This is occurring both on dndbeyond with no vtt open and when I send rolls to roll20.

I'd like this feature to be implemented similarly to how the sorcerer's elemental affinity works.

I could add a damage modifier to the spells, and likely will until this is fixed, but it won't show up separately if I do it that way. Having it separate is only really important for wall of fire, however, since it only applies to the initial damage.

**Additional context**
The feature reads: Starting at 6th level, your link to the Celestial allows you to serve as a conduit for radiant energy. You have resistance to radiant damage, and when you cast a spell that deals radiant or fire damage, you can add your Charisma modifier to one radiant or fire damage roll of that spell against one of its targets.

Aeristoka:
Fixes #894

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -589,6 +589,12 @@ const character_settings = {
         "type": "bool",
         "default": false
     },
+    "warlock-the-celestial-radiant-soul": {
+        "title": "Warlock: The Celestial: Radiant Soul",
+        "description": "The bond with your wildfire spirit enhances your destructive and restorative spells.",
+        "type": "bool",
+        "default": true
+    },
     "genies-vessel": {
         "title": "Warlock: The Genie: Genie's Wrath",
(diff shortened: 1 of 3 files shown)


kakaroto:
All looks good to me! It's weird that you have two features with the same name.. at least one is a racial trait, and the other a class feature.
Thanks!

Aeristoka:
(Generic comment about WotC QA not being thorough and this not being the first time they've done this so far that Beyond20 has had to account for...)

Issue #896, opened 2021-11-26
Add a "Advanced" mode for options


kakaroto:
Being able to toggle an "Advanced" vs "Basic" mode for the Beyond20 options would allow us to cram all sorts of options which so far I've tried to avoid adding to avoid bloating the options interface.
Could solve the #877 issue by allowing #878 to be merged (where the string would be customizable instead), and move the "components prefix" and "critical hit prefix" to the advanced options too, as well as some of the less used options, and adding an option to not prompt skill checks that are conditional, which seems to be bother @Aeristoka.

PR #906, opened 2021-12-28
Initial re-name of Handedness

### LINKED ISSUE #868: Feature Request: Always Display The Number of Hands Used For An Attack Alongside the Damage Roll When Using Versatile Weapons
I'm not sure I'd call it a problem, but I find myself wondering if my players used their versatile weapon with one or two hands.

If a player has changed the default setting for versatile weapons to either _Use weapon one-handed_ or _Use weapon two-handed_, it would be nice to see that reflected in the VTT alongside the damage type, no differently than when it's labeled for when a player has not changed the default setting from _Roll both damages separately_.

The alternative is to hover over the damage amount to see the dice formula and compare it to the player's character sheet to match it with either the one-handed or two-handed damage formula. Another alternative would be to memorize the dice formulas.

**Additional context**
For comparison, here's the screenshot of what the roll looks like in Roll20 when both damages are rolled separately. It indicates which one is one-handed and which one is two-handed.

image

When the option is chosen to **only** roll two-handed damage in Beyond 20, the damage roll in the VTT only indicates the damage type (Slashing, in this case) but doesn't specify that the attack was made with two hands.

image

Aeristoka:
Fixes #868

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -286,6 +286,13 @@ const options_list = {
         "advanced": true
     },
 
+    "weapon-handedness": {
+        "title": "Always Show Weapon Handedness",
+        "description": "On Versatile weapons, always show weapon-handedness, even if 1-handed or 2-handed is selected",
+        "type": "bool",
+        "default": false,
+        "advanced": true
+    },
 
     "roll20-tab": {
         "type": "migrate",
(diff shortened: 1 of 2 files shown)


Aeristoka:
Pending #896 for completeness

DO NOT MERGE

PR #912, opened 2022-01-10
Additional Fixes for Weapon Handedness

### LINKED ISSUE #911: Totals bar sums both single and double handed damage when rolling with a versatile weapon while raging with DnD Beyond's Virtual Dice Enabled.
**Describe the bug**
The totals bar shown in the Roll-20 chat sums both the single and double handed damage when rolling with a versatile weapon while raging with DnD Beyond's Virtual Dice Enabled.

**To Reproduce**
1. Open or create a babarian charecter in D&D Beyond.
2. In Beyond 20's settings, toggle both "Use D&D Beyond's Digital Dice" and "Barbarian: Rage! You are raging, ARRGGHHHHHH" on.
3. Initiate a damage roll for a versatile weapon.
4. In Roll20, see that an attack roll is displayed with both one-handed and two-handed damage, followed by a box with 2 rage damage. But then a blank Totals bar is displayed, followed by a second box with a total of both the single and double handed damage plus the two rage damage.

**Expected behavior**
The totals bar should display the total damage for one and two handed attacks separately, 

**Screenshots**
Output as it exists: 
image

Rendering of possible look for expected behavior: 
image

Firefox 95.0.1 (64-bit) for Ubuntu with Beyond 20 Version 2.6.0

Aeristoka:
Expansion and fix-up of #906
Fixes #911

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -744,6 +744,7 @@ async function rollItem(force_display = false, force_to_hit_only = false, force_
     const is_instrument = isItemAnInstruction(item_name, item_tags);
     const description = descriptionToString(".ct-item-detail__description");
     const quantity = $(".ct-item-pane .ct-simple-quantity .ct-simple-quantity__value .ct-simple-quantity__input").val();
+    let is_versatile = false;
     if (!force_display && Object.keys(properties).includes("Damage")) {
         const item_full_name = $(".ct-item-pane .ct-sidebar__heading .ct-item-name,.ct-item-pane .ct-sidebar__heading .ddbc-item-name").text();
         let to_hit = properties["To Hit"] !== undefined && properties["To Hit"] !== "--" ? properties["To Hit"] : null;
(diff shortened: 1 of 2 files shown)


src/dndbeyond/base/utils.js
```diff
@@ -117,7 +117,7 @@ async function queryDamageTypeFromArray(name, damages, damage_types, possible_ty
 
 async function buildAttackRoll(character, attack_source, name, description, properties,
                          damages = [], damage_types = [], to_hit = null,
-                         brutal = 0, force_to_hit_only = false, force_damages_only = false, {weapon_damage_length=0}={}) {
+                         brutal = 0, force_to_hit_only = false, force_damages_only = false, {weapon_damage_length=0}={}, is_versatile) {
```
kakaroto, line 120: Put the `is_versatile` inside the options object `{weapon_damage_length=0, is_versatile=false}={}` so we don't have to always pass all the arguments in order to set a specific one, which was why we added that options object.

src/dndbeyond/content-scripts/character.js
```diff
@@ -907,7 +909,8 @@ async function rollItem(force_display = false, force_to_hit_only = false, force_
             brutal,
             force_to_hit_only,
             force_damages_only,
-            {weapon_damage_length});
+            {weapon_damage_length},
+            is_versatile);
```
kakaroto, line 913: Actually, it might be better not to pass `is_versatile` to buildAttack, since it's a thing that is specific to weapon attacks, not a generic attack thing, so I'd rather see it added afterwards (like we do for `item-type` below), otherwise it wouldn't make sense for us to see a spell attack with `request.versatile = false`

Aeristoka: That's 110% the right approach I think, code pushed, testing now.

PR #925, opened 2022-01-31
Changes to Force Critical Behavior

Aeristoka:
This seems like a proper change to the Force Critical behavior, as discussed on Discord: https://discord.com/channels/578874022968688640/579077147994488862/937516920469598248 
(there and below)

Essentially:

1. Force Criticals is wanted
2. Even with no to-hit, determine critical_damages
3. Even with no to_hit, roll the critical

I can't think of any reason at the moment this would cause issues specifically, and it fixes rolling things like Sneak Attack (with Force Critical via setting or Key Modifier) from the "Other" tab of "Actions":
image

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -580,9 +580,8 @@ class Beyond20RollRenderer {
             }
 
             // If rolling the attack, check for critical hit, otherwise, use argument.
-            if (request.rollAttack) {
-                if (to_hit.length > 0)
-                    this.processToHitAdvantage(to_hit_advantage, to_hit)
+            if (request.rollAttack && to_hit.length > 0) {
+                this.processToHitAdvantage(to_hit_advantage, to_hit)
                 const critical_limit = request["critical-limit"] || 20;
                 is_critical = this.isCriticalHitD20(to_hit, critical_limit);
             } else {
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -196,69 +196,67 @@ async function buildAttackRoll(character, attack_source, name, description, prop
             await queryDamageTypeFromArray(roll_properties.name, damages, damage_types, ["Radiant", "Necrotic"]);
         }
         
-        if (to_hit) {
-            const crits = damagesToCrits(character, damages, damage_types);
-            const crit_damages = [];
-            const crit_damage_types = [];
-            for (let [i, dmg] of crits.entries()) {
-                if (dmg != "") {
-                    crit_damages.push(dmg);
-                    crit_damage_types.push(damage_types[i]);
-                }
+        const crits = damagesToCrits(character, damages, damage_types);
+        const crit_damages = [];
+        const crit_damage_types = [];
+        for (let [i, dmg] of crits.entries()) {
+            if (dmg != "") {
+                crit_damages.push(dmg);
+                crit_damage_types.push(damage_types[i]);
             }
-            if (character.hasFeat("Piercer")) {
-                for (let i = 0; i < damage_types.length; i++) {
-                    if (damage_types[i].includes("Piercing")){
-                        const piercer_damage = damagesToCrits(character, [damages[i]]);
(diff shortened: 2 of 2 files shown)


src/dndbeyond/base/utils.js
```diff
@@ -196,7 +196,7 @@ async function buildAttackRoll(character, attack_source, name, description, prop
             await queryDamageTypeFromArray(roll_properties.name, damages, damage_types, ["Radiant", "Necrotic"]);
         }
         
-        if (to_hit) {
+        if (to_hit || force_damages_only) {
```
kakaroto, line 199: I think you can do a full roll without `force-damages-only`. Like clicking the beyond20 button on the sneak attack action would count as a normal roll, without "force_damages_only" being set to true.

src/dndbeyond/base/utils.js
```diff
@@ -328,7 +328,7 @@ async function sendRoll(character, rollType, fallback, args) {
     if (req.advantage === RollType.QUERY) {
         req.advantage = await dndbeyondDiceRoller.queryAdvantage(args.name || rollType, req["advantage-query"]);
     }
-    if (req["to-hit"] && character.getGlobalSetting("weapon-force-critical", false) || key_modifiers.force_critical) {
+    if (character.getGlobalSetting("weapon-force-critical", false) || key_modifiers.force_critical) {
         req["critical-limit"] = 1;
```
kakaroto, line 332: I'd wrap the `critical-limit` under an `if (req["to-hit"])`

src/dndbeyond/base/utils.js
Aeristoka: image

This somehow sets `force_damages_only`, and has no to-hit, so it won't run damagesToCrits without this, then fails like this:
image

image

kakaroto: Yeah but clicking here with force crit : 
image
or on that button : 
image

Will not cause `force_damages_only` to be set, so you'd get the same error because `damagesToCrits` won't be able to find the crits.. would need to have it check if to_hit or force crits... which isn't the available info at that time.

Aeristoka: Well crap...

src/common/roll_renderer.js
```diff
@@ -580,9 +580,8 @@ class Beyond20RollRenderer {
             }
 
             // If rolling the attack, check for critical hit, otherwise, use argument.
-            if (request.rollAttack) {
-                if (to_hit.length > 0)
-                    this.processToHitAdvantage(to_hit_advantage, to_hit)
+            if (request.rollAttack && to_hit.length > 0) {
```
kakaroto, line 583: hummm, interesting, I assume you found a bug here we hadn't noticed before? 
Anyways, `to_hit` might be `undefined` if there wasn't one, so the `.length > 0` would fail. doing `&& !!to_hit` would be better I think or `&& to_hit && to_hit.length > 0)`

gedankenexperimenter:
With these changes, forcing a critical hit works for Sneak Attack damage, but only if I click on the "2d6" damage formula in the Sneak Attack row of the Attacks (Other) table, not if I click on the crossed swords action icon in that table (on the left side of the row).  I don't know if that was intended or not.

Aeristoka:
image

Works just fine with Crossed Swords

gedankenexperimenter:
Could this be a Roll20 _vs_ FoundryVTT issue?  It's definitely not working for me with Roll20. ("Use D&D Beyond Digital Dice" is off, if that matters.)

Aeristoka:
> Could this be a Roll20 _vs_ FoundryVTT issue? It's definitely not working for me with Roll20. ("Use D&D Beyond Digital Dice" is off, if that matters.)

what I displayed was just DnDBeyond, not FoundryVTT at all, but this does appear to be a Roll20-specific thing now.

gedankenexperimenter:
Thanks for the clarification; I see that now (the styling of the box looks enough like Foundry to fool me at a glance).  I also see those results when I have no VTT tabs open.

gedankenexperimenter:
Would you like me to report this as a separate issue?

PR #928, opened 2022-02-01
Fix for Brutal Critical/Savage Attacks w/ GWF

Aeristoka:
There was no check on Brutal Critical/Savage Attacks applying Great Weapon Fighting. There may be a better one that this, but this is something that is available already at the point of this function.


kakaroto (approved):
Looks good, makes sense! Good catch 👍

PR #937, opened 2022-02-09
Implement Monk: Diamond Soul

Aeristoka:
Adds proficiency to Death Saves at Monk level 14

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -466,6 +466,8 @@ class Beyond20RollRenderer {
     rollDeathSave(request, custom_roll_dice = "") {
         const data = { "custom_dice": custom_roll_dice }
         if (request.modifier) data.modifier = request.modifier
+        if (character.hasClassFeature("Diamond Soul") && character.getSetting("monk-diamond-soul", false))
+            data.modifier += character._proficiency;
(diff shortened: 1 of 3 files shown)


src/common/roll_renderer.js
```diff
@@ -466,6 +466,8 @@ class Beyond20RollRenderer {
     rollDeathSave(request, custom_roll_dice = "") {
         const data = { "custom_dice": custom_roll_dice }
         if (request.modifier) data.modifier = request.modifier
+        if (character.hasClassFeature("Diamond Soul") && character.getSetting("monk-diamond-soul", false))
+            data.modifier += character._proficiency;
```
kakaroto, line 470: humm... isn't the modifier a string, so the addition works as a concatenation ? 
But I see we do stuff  like that elsewhere, I think it's because the proficiency will always be a "+x" string, so the modifier ends up being "4 +6" for example, which works as a string 🤔 
I assume this has been tested though and works, so 👍

Aeristoka: Tested and works, yes, yeah it's kinda following other code at this point.

kakaroto: Very thank you much.

PR #941, opened 2022-02-15
Fix to make Predatory Strike apply Crimson Rite

Aeristoka:
Also adds Predatory Strike to Melee Attack Processing

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -586,7 +586,9 @@ function handleSpecialWeaponAttacks(damages=[], damage_types=[], properties, set
     if (character.hasClass("Blood Hunter")) {
         // Bloodhunter: Crimson Rite
         if (character.getSetting("bloodhunter-crimson-rite", false) &&
-            character.hasClassFeature("Crimson Rite")) {
+            character.hasClassFeature("Crimson Rite") &&
+            ((properties["Attack Type"] == "Ranged" || properties["Attack Type"] == "Melee") ||
+            action_name.includes("Predatory Strike"))) {
             const bloodhunter_level = character.getClassLevel("Blood Hunter");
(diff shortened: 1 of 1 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -586,7 +586,9 @@ function handleSpecialWeaponAttacks(damages=[], damage_types=[], properties, set
     if (character.hasClass("Blood Hunter")) {
         // Bloodhunter: Crimson Rite
         if (character.getSetting("bloodhunter-crimson-rite", false) &&
-            character.hasClassFeature("Crimson Rite")) {
+            character.hasClassFeature("Crimson Rite") &&
+            ((properties["Attack Type"] == "Ranged" || properties["Attack Type"] == "Melee") ||
+            action_name.includes("Predatory Strike"))) {
```
kakaroto, line 591: Humm... why "Ranged || Melee"... what other type of weapon attacks are there?

Aeristoka: Not sure this is the best solution honestly, but it was a way to say "If it's Ranged or Melee" (rollItem), or the action_name includes "Predatory Strike" (rollAction) then do the thing. Probably a better way, I just couldn't see it at the time.

Aeristoka: I believe this break also occurred due to an update to Bloodhunter. I mostly just don't like Bloodhunter from all the weird broken crap it appears to have (in the example character sheet I was given, Unarmed Strike has proficiency, Predatory strikes, the "buffed" unarmed strike has no proficiency. It's just a mess of a class).

kakaroto: My confusion, I thought this was in `rollItem`, thanks for the explanation.

PR #947, opened 2022-02-26
Implement Custom Roll for Blood Curse of the Eyeless

### LINKED ISSUE #159: Support Blood Hunter hemocraft die for Blood Curses
If you click on a blood curse (only tried with "blood curse of the eyeless" you can roll on roll20 but it just rolls a d20 instead of a "hemocraft die" which changes after leveling up in Blood Hunter class. On level 1 it's only a d4.

Aeristoka:
Fixes #159 
Blood Curse of the Eyeless appears to be the only item that needs a roll like this.

However, it appears there are a ton of Bloodhunter things not implemented. I'm hesitant to go further with it, as it's Unofficial, and there have been no requests for it, but I also like having as complete coverage as possible.

Your thoughts @kakaroto ?

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -1046,6 +1046,14 @@ async function rollAction(paneClass, force_to_hit_only = false, force_damages_on
             "description": description,
             "modifier": inspiration_die
         });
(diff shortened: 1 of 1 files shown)


kakaroto:
My thoughts is that it's worth merging, but also unsure if it's worth pursuing every quirk of the blood hunter. This will do for now, thanks!

PR #948, opened 2022-02-26
Fix for displaying dropped dice indication for advantage/disadvantage in roll20, when using digital dice and D&D 5E By Roll20 template

Cube-o:
**Problem:**
When rolling with advantage/disadvantage into roll20, while using digital dice and the D&D 5E By Roll20 template, both rolls were shown in dark black. Correctly one of them should be lighter to indicate it being dropped due to the advantage/disadvantage.
Currently showing: 
image
Correct version for disadvantage:
image

**Changes summary:**
Originally attack prop `always: "1"` was being added. I added 2 special cases for when the roll is indicated to be with advantage/disadvantage and the number of rolled dice is 2 (because rolls beyond 2 are grouped into the second roll and the advantage indicator in the template is unable to process those)

### DIFF
--- a/src/roll20/content-script.js
+++ b/src/roll20/content-script.js
@@ -679,7 +679,9 @@ async function handleOGLRenderedRoll(request) {
     if (request.attack_rolls.length > 1) {
         atk_props["r1"] = convertRollToText(request.whisper, request.attack_rolls[0]);
         atk_props["r2"] = request.attack_rolls.slice(1).map(roll => convertRollToText(request.whisper, roll)).join(" | ")
-        atk_props["always"] = "1";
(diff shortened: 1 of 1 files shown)


kakaroto:
Huh... so that works using the actual result even if no dice are actually rolled? I thought that didn't work for some reason 🤔

Cube-o:
> Huh... so that works using the actual result even if no dice are actually rolled? I thought that didn't work for some reason 🤔

Was honestly a bit surprised myself, so I was thinking "oh, if this would work..." and then it did.

kakaroto:
Awesome, thanks, will merge it! 👍

Issue #833, opened 2021-06-17
Send messages to any website


halibut:
I understand you don't want to support more VTTs, however this extension is doing a lot of the heavy lifting of tying into dndBeyond to send dice rolls and other messages to supported websites. It would be really great if any website could receive messages through this extension. And then, if the site recognizes the messages, handle display/updates itself. 

**Describe the solution you'd like**
I would like to be able to have any site signal to the extension that it is capable of receiving messages, and have Beyond20 send raw messages to it. 

I would imagine there would need to be an initial detection step where the extension looks for something on the page to let it know that page can receive messages. Maybe it looks for a global JS variable or an element by id. I think this kind of logic is already happening for at least 1 of the VTTs, where the extension detects if there's a specific JS script imported in the html. The site's HTML would explicitly have something that the extension is looking for.

If the detection step succeeds, then that page would become a 1-way target for messages sent from the extension. Messages would be received by the content script on that page and posted to it via window.postMessage.

From that point on, it's up to the site to have an onmessage listener and know what to do with the raw messages.

From a support standpoint it moves most of the responsibility to the individual site to interpret and display the messages correctly. Though I admit it may be tricky to get users to understand that an issue with another VTT website is not a Beyond20 issue. I would live with a disclaimer popup or banner when a non-supported VTT page is connected that tells users to direct all of their support issues/questions to the site operator.

**Describe alternatives you've considered**
I'm not an extension developer (yet), so I don't know if there's another solution for this. Maybe a separate extension that can register with Beyond20 and receive its messages, but that sounds complicated (and would still require changes to the Beyond20 extension to send messages externally). 

Maybe Beyond20 could postMessage on the various dndbeyond pages, and another extension could listen for those and send them to other pages.  (Also requires changes to Beyond20.)

Lastly, if it's decided that no changes will be made to this extension for this feature, then I will try to fork this project and strip out support for specific VTTs and just do this 1 function. But I think it would be a bad user experience for people who might then need to use both extensions.

**Additional context**
I'm currently working on a service that rebroadcasts rolls/events/anything to non-browser-based clients (targeting tabletop simulator first, but anything that can long-poll or receive SSEs via HTTP will work with it). The idea is to have a site that players connect to, which receives messages through Beyond20. The site would then send those messages to the backend which makes them available as Server Sent Events for any listening HTTP client. (All of this is working in a proof of concept, but I now need to connect dndbeyond to it and Beyond20 is by far the best solution I can think of.)

[...]

Issue #179, opened 2020-04-10
Use permissions.request and make permissions less broad on FF


Jonfor:
The permissions for Firefox seem to include "Access your data for all websites" instead of the permissions specified in the Firefox manifest. Chrome displays the specific website permissions correctly. I've attached images of both below.

I'm not sure why this is different, the only difference I see in permissions storage:

Firefox -> "*://*/*game"
Chrome -> "*://*.forgevtt.com/game",
		  "*://*.forge-vtt.com/game",
                  "*://beyond20.kicks-ass.org/roll",

Any idea why Firefox seems to be displaying the permissions requested incorrectly?

![Capture](https://user-images.githubusercontent.com/3246528/78963387-4dc31d00-7ac5-11ea-87a5-88f2e7410da5.PNG)

![Capture](https://user-images.githubusercontent.com/3246528/78963513-b1e5e100-7ac5-11ea-98db-8dbca68d0003.PNG)

kakaroto:
The FF manifest is here : https://github.com/kakaroto/Beyond20/blob/master/manifest_ff.json (which I rename before bundling it for release on FF)
And the permissions, as you can see, are : 
```json
	"permissions": [
		"activeTab",
		"tabs",
		"storage",
		"*://app.roll20.net/editor/",
		"*://*.dndbeyond.com/*",
		"*://*/*game"
	],
```
As you can see, there's a "*" permission in there for the host, which is why it asks for all sites.

The reason for that permission is because Foundry VTT is self hosted, so it can be on any URL, or any IP address, and accessing all sites is necessary to detect and work on Foundry VTT in Firefox.
In Chrome, things are different because Chrome changed things to make it harder to get reviewed when asking for broad permissions like that (multiple weeks before an update gets reviewed and released on the store).
For chrome it's more annoying to FVTT users since players have to click the beyond20 icon in the address bar every time they load/reload their Foundry VTT page in order to 'activate' it, and it's one of the biggest annoyances with the extension.
I've explained it all on the release notes of v0.5 where the change happened : https://beyond20.here-for-more.info/release_notes#v05

As for "what does the extension do on all sites", this is the script that gets loaded : https://github.com/kakaroto/Beyond20/blob/master/src/fvtt_test.js
It does a single thing, if the title of the page is "Foundry Virtual Table Top", then it activates the icon, which will then load the FVTT content script. Otherwise, it does nothing.
[...]

PR #955, opened 2022-03-13
Adding support for Roleplay.tv

ClaudeSutterlin:
We added support for Roleplay.tv!

We made minimal changes to the Beyond20 codebase to allow for dice rolls and health updates to sync to the Roleplay.tv platform. Thanks for creating and supporting this extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Beyond20: D&D Beyond & Roll20/Foundry VTT Integration
 ==
 
-[Beyond20](https://beyond20.here-for-more.info/) is a browser extension that integrates the Character Sheet from D&D Beyond into Roll20, Foundry VTT & Discord.
+[Beyond20](https://beyond20.here-for-more.info/) is a browser extension that integrates the Character Sheet from D&D Beyond into Roll20, Foundry VTT, Discord, and Roleplay.tv.
 
 # Install and Use
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -130,7 +130,12 @@ const SRC_FILES = {
 		"src/dndbeyond/base/message-broker.js",
 		"src/dndbeyond/page-scripts/message-broker.js",
         "src/common/sandbox-footer.js"
-	]
+	],
+    rptv: [
+        "src/common/utils.js",
+        "src/common/settings.js",
+        "src/roleplaytv/content-script.js"
+    ]
 }
 const CSS_FILES = ['src/extension/beyond20.css']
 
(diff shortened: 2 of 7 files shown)


Aeristoka:
https://beyond20.here-for-more.info/faq#can-you-add-support-for-insert-new-vtt-name

I would not anticipate this getting merged. Certainly neat, and you can maintain this on your side for your use, but please read on that link.

Final decision is with @kakaroto, who is currently out on vacation.

ClaudeSutterlin:
That makes sense and I totally get it. We'd love to be maintainers for those portions and for our users, but we can just publish our fork separately if that's what @kakaroto prefers. We sponsored the project at a very basic level and would expand that if we become a burden. Just let me know, thanks.

kakaroto:
> 

Hi @ClaudeSutterlin, thanks for the contribution! As discussed with @Aeristoka, unfortunately, I won't be merging this PR because of the problem explained in the FAQ and we don't want to add more complexity to the extension than needed.

That being said, I can see that your PR doesn't actually add any support for interacting with roleplay.tv but only sends DOM messages to it, and I assume that you made roleplay.tv itself support receiving those messages and acting upon them. That's really great, because it means it should work without needing to add any explicit support for your website. 
If you look at issue #833, I already do have a plan for sending those kinds of events to any website, and you can also see in this commit https://github.com/kakaroto/Beyond20/commit/54037c7065861140c764019a92888b650973e766 that I send the parsed roll requests to the DOM, unfortunately they are only getting sent to the DDB site itself at the moment, and what's remaining in #833 is to basically have a setting where a user could add a list of domains in which Beyond20 would be loaded, in which case, it would be able to prompt the user to give permission to load on that site (like we're doing now with Foundry VTT, see #179).

If you want to work towards that and contribute that feature, that's one that I'd be glad to merge, basically : 
- A setting where a user could enter custom domain names to load the extension on
- when the browser_action icon is clicked, if the custom domain matches (and it's not roll20 or foundry or astral), it would load a small content-script like the one you wrote
- I think it should automatically show the permission request in that case like below : 
image
- the content script would simply forward messages to the DOM
```js
function handleMessage(request, sender, sendResponse) {
  sendRollRequestToDOM(request);
}
```

This is a planned feature, if you want to contribute it, great, I'll happily review and merge it, if not, then I'll probably implement it myself soon(-ish) anyway!
At that point, the idea is that any website could directly add support for Beyond20, or separate extensions could add support for beyond20, so you could release an extension which would work only on roleplay.tv and which would depend on Beyond20 and it would just listen to the DOM events and act on them without needing to duplicate all the D&D Beyond side of things (avoids possible conflicts).

I hope that helps, 
Thanks again!

PR #960, opened 2022-03-19
Add Passive Perception Floor for Perception Roll

Aeristoka:
Advanced Option

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -435,6 +435,17 @@ class Beyond20RollRenderer {
         const {advantage, rolls} = await this.getToHit(request, title, "", data, type)
         await this._roller.resolveRolls(title, rolls, request);
         this.processToHitAdvantage(advantage, rolls);
+        if (request.skill && request.skill == "Perception" && character.getGlobalSetting("passive-perception-min")) {
+            let passive_perception = parseInt(character._passive_perception);
+            if (request.advantage == RollType.ADVANTAGE)
+                passive_perception += 5;
+            if (request.advantage == RollType.DISADVANTAGE)
+                passive_perception -= 5;
+            for (let i = 0; i < rolls.length; i++) {
+                if (rolls[i].total < passive_perception)
+                    rolls[i]._total = passive_perception;
+            }
+        }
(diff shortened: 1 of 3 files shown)


Aeristoka:
Possible to also add Passive Investigation and Passive Insight to this as well, as 2x more advanced options.

src/common/roll_renderer.js
```diff
@@ -435,6 +435,17 @@ class Beyond20RollRenderer {
...
+            let passive_perception = parseInt(character._passive_perception);
+            if (request.advantage == RollType.ADVANTAGE)
+                passive_perception += 5;
+            if (request.advantage == RollType.DISADVANTAGE)
+                passive_perception -= 5;
+            for (let i = 0; i < rolls.length; i++) {
+                if (rolls[i].total < passive_perception)
+                    rolls[i]._total = passive_perception;
+            }
+        }
```
kakaroto, line 448: this would have had to be a `min10` modifier to the roll instead

kakaroto:
Unsure about this, as far as I know, that's not really how it works.. or at least, there's a lot of debate on whether or not the passive is the floor.
Technically, you have a passive perception, passive insight, passive stealth, etc... and it's just 10+modifier. And considering that reliable talent and silver tongue are specific class features that makes your roll be a `min10` for specific skills, that means that class feature gives you your passive as the floor... so unless you have that as an actual class features, it wouldn't make sense that perception rolls would default to your passive as the floor, even as an advanced option, it seems to go against RAW and RAI.

Aeristoka:
> Unsure about this, as far as I know, that's not really how it works.. or at least, there's a lot of debate on whether or not the passive is the floor. Technically, you have a passive perception, passive insight, passive stealth, etc... and it's just 10+modifier. And considering that reliable talent and silver tongue are specific class features that makes your roll be a `min10` for specific skills, that means that class feature gives you your passive as the floor... so unless you have that as an actual class features, it wouldn't make sense that perception rolls would default to your passive as the floor, even as an advanced option, it seems to go against RAW and RAI.

https://rpg.stackexchange.com/questions/48281/does-passive-perception-score-supersede-the-result-of-a-lower-active-perception Is the reference, thought I had included that here.

PR #965, opened 2022-03-26
Potential Fix for FVTT 10 Chat Message Rolls

### LINKED ISSUE #963: Investigate Change of required Roll Data Attachment to Chat Message in FoundryVTT v10
https://foundryvtt.com/releases/10.260

image

https://gitlab.com/foundrynet/foundryvtt/-/issues/6467

Aeristoka:
I think this is all this wants?

### DIFF
--- a/src/fvtt/page-script.js
+++ b/src/fvtt/page-script.js
@@ -130,7 +130,9 @@ class FVTTDisplayer {
                 }
                 return Roll.fromData(r)
             });
-            if (isNewerVersion(fvttVersion, "0.8")) {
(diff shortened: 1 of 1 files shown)


kakaroto:
It should be, yeah, that's what it sounds like.
Why did you close this though?

Aeristoka:
> It should be, yeah, that's what it sounds like. Why did you close this though?

I got in my head about Dice So Nice not picking up the new Roll data (that it's not updated for), I'll re-work it a touch and re-open it.

kakaroto:
Need more testing probably, but I think this looks good, so I'll merge it as is for now 👍

Issue #494, opened 2020-08-19
add a "display in VTT" for feats pages


kakaroto:
For example https://www.dndbeyond.com/feats/lucky

kakaroto:
Fixed by #975

PR #975, opened 2022-04-13
Display Feat in VTT button on the feat pages

IsaacAbramowitz:
This addresses issue #494

### DIFF
--- a/src/dndbeyond/content-scripts/feat.js
+++ b/src/dndbeyond/content-scripts/feat.js
@@ -1,0 +1,98 @@
+console.log("Beyond20: D&D Beyond Feats module loaded.");
+
+class FeatCharacter extends CharacterBase {
+    constructor(global_settings) {
+        super("feat", global_settings);
+    }
+    getDict() {
+        const dict = super.getDict();
+        if (this.avatar) {
+            dict.avatar = this.avatar;
+        }
+        return dict;
+    }
+}
+
+var character = null;
+var settings = getDefaultSettings();
+
+function addDisplayFeatButton() {
+    const icon32 = chrome.runtime.getURL("images/icons/badges/normal32.png");
+    const button = E.a({ class: "ct-beyond20-roll button-alt", href: "#" },
+        E.span({ class: "label" },
+            E.img({ class: "ct-beyond20-item-icon", src: icon32, style: "margin-right: 10px;" }),
+            "Display Feat on VTT")
+    );
+    const feat_name = $(".page-title").text().trim();
+    const feat_type = "";
+    const description = descriptionToString(".item-details .more-info-content, .details-container-feat .details-container-content-description").trim();
+    const feat_tags = $(".details-container-content-footer .tags .tag").toArray().map(elem => elem.textContent);
+    $(".page-heading__content").after(button);
+
+    $(".ct-beyond20-roll").css({
+        "float": "right",
+        "display": "inline-block"
+    });
+    $(".ct-beyond20-roll").on('click', (event) => {
+        sendRoll(character, "item", "0", {
+            "name": feat_name,
+            "description": description,
+            "item-type": feat_type,
+            "tags": feat_tags
+        });
+    }
+    );
+}
+
+function documentLoaded(settings) {
+    cleanupAlertifyComments();
+    character = new FeatCharacter(settings);
+    if (isRollButtonAdded()) {
+        chrome.runtime.sendMessage({ "action": "reload-me" });
+    } else {
+        addDisplayFeatButton();
+        const avatar = $(".details-aside .image a");
+        if (avatar.length > 0) {
+            character.avatar = avatar[0].href;
+            const avatarImg = $(".details-aside .image");
+            if (avatarImg) {
+                addDisplayButton(() => {
+                    sendRoll(character, "avatar", character.avatar, { "name": "Feat" });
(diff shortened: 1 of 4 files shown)


src/dndbeyond/content-scripts/feat.js
```diff
@@ -0,0 +1,98 @@
...
+    const description = descriptionToString(".item-details .more-info-content, .details-container-feat .details-container-content-description").trim();
+    const feat_tags = $(".details-container-content-footer .tags .tag").toArray().map(elem => elem.textContent);
+    $(".page-heading__content").after(button);
+
+    $(".ct-beyond20-roll").css({
+        "float": "right",
+        "display": "inline-block"
+    });
+    $(".ct-beyond20-roll").on('click', (event) => {
+        sendRoll(character, "item", "0", {
```
kakaroto, line 37: I think the type of roll should be "feature", rather than "item". See `displayFeature` in the `content-scripts/character.js`

src/dndbeyond/content-scripts/feat.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+var character = null;
+var settings = getDefaultSettings();
+
+function addDisplayFeatButton() {
+    const icon32 = chrome.runtime.getURL("images/icons/badges/normal32.png");
+    const button = E.a({ class: "ct-beyond20-roll button-alt", href: "#" },
+        E.span({ class: "label" },
+            E.img({ class: "ct-beyond20-item-icon", src: icon32, style: "margin-right: 10px;" }),
```
kakaroto, line 23: Since this is only to display the feat, and not do an actually dice roll, the button shouldn't include the beyond20 dice icon.

src/dndbeyond/content-scripts/feat.js
```diff
@@ -0,0 +1,98 @@
...
+            const avatarImg = $(".details-aside .image");
+            if (avatarImg) {
+                addDisplayButton(() => {
+                    sendRoll(character, "avatar", character.avatar, { "name": "Feat" });
+                }, avatarImg, { small: false, image: true });
+            }
+        }
+        const feat_name = $(".page-title").text().trim();
+        if (settings['subst-dndbeyond'])
+            injectDiceToRolls(".item-details .more-info-content, .details-container-equipment .details-container-content-description-text", character, feat_name);
```
kakaroto, line 66: The injection of dice rolls should likely be on `.details-container-feat .details-container-content-description-text`, not on `.details-container-equipment` and it doesn't need to act on the `.item-details .more-info-content` selector either.

src/dndbeyond/content-scripts/feat.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+function documentLoaded(settings) {
+    cleanupAlertifyComments();
+    character = new FeatCharacter(settings);
+    if (isRollButtonAdded()) {
+        chrome.runtime.sendMessage({ "action": "reload-me" });
+    } else {
+        addDisplayFeatButton();
+        const avatar = $(".details-aside .image a");
```
kakaroto, line 54: The image doesn't seem to be on the same selector. I'm checking on the Piercer feat for example, and it's `.details-container-aside .image-container img`, and since it's an `img` and not an `a`, that changes how the rest of the code would work (i.e: using the `src` attribute, not `href`, and checking for any other consequences)

src/dndbeyond/content-scripts/feat.js
```diff
@@ -0,0 +1,98 @@
...
+function addDisplayFeatButton() {
+    const icon32 = chrome.runtime.getURL("images/icons/badges/normal32.png");
+    const button = E.a({ class: "ct-beyond20-roll button-alt", href: "#" },
+        E.span({ class: "label" },
+            E.img({ class: "ct-beyond20-item-icon", src: icon32, style: "margin-right: 10px;" }),
+            "Display Feat on VTT")
+    );
+    const feat_name = $(".page-title").text().trim();
+    const feat_type = "";
+    const description = descriptionToString(".item-details .more-info-content, .details-container-feat .details-container-content-description").trim();
```
kakaroto, line 28: The `.item-details .more-info-content` portion of the selector can likely be removed here.

kakaroto (changes requested):
Thanks for implementing the feature! 👍 It generally looks good, though I see some left over code from the items handling (which I assume is what you used as template).
Should be easy fixes. If you don't have the time for it, I'll apply them quickly before merging when I work on the next release.
Thanks again!

src/dndbeyond/content-scripts/feat.js
IsaacAbramowitz: Without the dice-roll icon the button looks very short and doesn't stick out as much (see image)

also the `items` pages have a dice as well so you should maybe consider modifying that page too since items don't always have an associated roll either

image

IsaacAbramowitz:
@kakaroto I've addressed your comments, let me know if anything else needs fixing

kakaroto:
Thank you for making the changes. It looks good, and I've merged it! We're getting ready to make a release today and it will be included in it, so yeay! 👍

src/dndbeyond/content-scripts/feat.js
IsaacAbramowitz: Agreed, that's 100% how I quickly find the Beyond20 settings when looking at my character sheet

PR #983, opened 2022-05-01
Implement Extra Damage from Vicious Weapons

Aeristoka:
Vicious Weapons do an extra flat 7 Damage on attack rolls
image

This is similar to Swords of Sharpness:
image

Except DnDBeyond did not deign to implement the extra damage. This change makes the Vicious Weapons appear essentially the same as the Swords of Sharpness in rolls.

image

image

### DIFF
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -223,6 +223,14 @@ async function buildAttackRoll(character, attack_source, name, description, prop
                 crit_damages[0] = damagesToCrits(character, ["8d12"])[0];
             if (roll_properties.name === "Jim’s Magic Missile")
                 crit_damages[0] = damagesToCrits(character, ["3d4"])[0];
+            const matchViciousWeapon = description.match(/When you roll a 20 on your attack roll with this magic weapon, the target takes an extra (\d+) damage of the weapon’s type./);
+            if (matchViciousWeapon) {
+                if (!damages.some(damage => damage === matchViciousWeapon[1])) {
+                    damages.push(matchViciousWeapon[1]);
+                    damage_types.push("Vicious(20 On The Attack Roll)");
+                } 
+            }
+                       
             if (brutal > 0) {
                 const rule = parseInt(character.getGlobalSetting("critical-homebrew", CriticalRules.PHB));
                 let highest_dice = 0;


Aeristoka:
I might figure out a fancier way to do this... not sure it would matter entirely, but I kinda want to poke at it.

kakaroto:
Since they did it for the longsword of sharpness, this might be a case of 'better let DDB so they fix it from their side" otherwise we'll have issues when they do fix it and we're doubling it

Aeristoka:
> Since they did it for the longsword of sharpness, this might be a case of 'better let DDB so they fix it from their side" otherwise we'll have issues when they do fix it and we're doubling it

Then I'd have to go on the forums and request it (like someone did ~2.5 years ago with no reply)

kakaroto:
> > Since they did it for the longsword of sharpness, this might be a case of 'better let DDB so they fix it from their side" otherwise we'll have issues when they do fix it and we're doubling it
> 
> Then I'd have to go on the forums and request it (like someone did ~2.5 years ago with no reply)

lol. Yeah. Maybe a poke will help them, though with the current acquisition/transition, they might be a bit busy.
I'm not a fan though of trying to infer specific properties from the name. I'd accept it though if it did the following : 
- Check description of the item says "When  you roll a 20 on the target ... (\d+) damage" and use that number from the description
- AND, check that the damage isn't already included in the item's damage so it stops adding it if DDB fixes it.

That could work for all items, not just "of sharpness" or "vicious" and might be a cleaner solution.
Thoughts?

Aeristoka:
> > > Since they did it for the longsword of sharpness, this might be a case of 'better let DDB so they fix it from their side" otherwise we'll have issues when they do fix it and we're doubling it
> > 
> > 
> > Then I'd have to go on the forums and request it (like someone did ~2.5 years ago with no reply)
> 
> lol. Yeah. Maybe a poke will help them, though with the current acquisition/transition, they might be a bit busy. I'm not a fan though of trying to infer specific properties from the name. I'd accept it though if it did the following :
> 
> * Check description of the item says "When  you roll a 20 on the target ... (\d+) damage" and use that number from the description
> * AND, check that the damage isn't already included in the item's damage so it stops adding it if DDB fixes it.
> 
> That could work for all items, not just "of sharpness" or "vicious" and might be a cleaner solution. Thoughts?

Sounds like an interesting idea, but I'm so garbage at Regex, probably better that you do that 😄

kakaroto:
> Sounds like an interesting idea, but I'm so garbage at Regex, probably better that you do that 😄

urggg, I'm good with regexp, but the sentences are just soooo freaking different, it's not gonna be easy to regexp it, or even possible. Do you know/have a list of types of weapons with this kind of feature? If it's just vicious and 'of sharpness' weapons, then we can make it check for both separately... Just these checks : 
```js
const matchViciousWeapon = description.match("When you roll a 20 on your attack roll with this magic weapon, the target takes an extra (\d+) damage of the weapon’s type");
if (matchViciousWeapon) {
  damages.push(parseInt(matchViciousWeapon[1]);
  damage_types.push(damage_types[0]);
}
const matchSharpnessWeapon = description.match("When you attack a creature with this weapon and roll a 20 on the attack roll, that target takes an extra (\d+) slashing damage. ");
if (matchSharpnessWeapon) {
  damages.push(parseInt(matchSharpnessWeapon[1]);
  damage_types.push(damage_types[0]);
}
```

+ you know.. the check for whether that exact damage is already in the `damages` array so it doesn't double it once DDB fixes it

Aeristoka:
> > Sounds like an interesting idea, but I'm so garbage at Regex, probably better that you do that 😄
> 
> urggg, I'm good with regexp, but the sentences are just soooo freaking different, it's not gonna be easy to regexp it, or even possible. Do you know/have a list of types of weapons with this kind of feature? If it's just vicious and 'of sharpness' weapons, then we can make it check for both separately... Just these checks :
> 
> ```js
> const matchViciousWeapon = description.match("When you roll a 20 on your attack roll with this magic weapon, the target takes an extra (\d+) damage of the weapon’s type");
> if (matchViciousWeapon) {
>   damages.push(parseInt(matchViciousWeapon[1]);
>   damage_types.push(damage_types[0]);
> }
> const matchSharpnessWeapon = description.match("When you attack a creature with this weapon and roll a 20 on the attack roll, that target takes an extra (\d+) slashing damage. ");
> if (matchSharpnessWeapon) {
>   damages.push(parseInt(matchSharpnessWeapon[1]);
>   damage_types.push(damage_types[0]);
> }
> ```
> 
> * you know.. the check for whether that exact damage is already in the `damages` array so it doesn't double it once DDB fixes it

I think I've got it figured now.

Firstly, only "Vicious" needs the code, "Sharpness" weapons are all implemented properly. Also had to workaround the fact that Versatile weapons can have both damages present, and work the array. I feel pretty good about it now.

src/dndbeyond/base/utils.js
```diff
@@ -223,6 +223,19 @@ async function buildAttackRoll(character, attack_source, name, description, prop
...
                 crit_damages[0] = damagesToCrits(character, ["3d4"])[0];
+            const matchViciousWeapon = description.match(/(When you roll a 20 on your attack roll with this magic weapon, the target takes an extra )(\d+)( damage of the weapon’s type.)/);
+            if (matchViciousWeapon) {
+                let vicious_damage_present = false;
+                damages.forEach(function (damage, index) {
+                    if (damage == matchViciousWeapon[2]) {
+                        vicious_damage_present = true;
+                    }
+                });
+                if (!vicious_damage_present) {
```
kakaroto, line 234: Here's a better way to do the check... instead of a temp variable and using `forEach`, you can do : 
```js
if (damages.some(damage => damage === matchViciousWeapon[2])) {
}
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

src/dndbeyond/base/utils.js
```diff
@@ -223,6 +223,19 @@ async function buildAttackRoll(character, attack_source, name, description, prop
                 crit_damages[0] = damagesToCrits(character, ["8d12"])[0];
             if (roll_properties.name === "Jim’s Magic Missile")
                 crit_damages[0] = damagesToCrits(character, ["3d4"])[0];
+            const matchViciousWeapon = description.match(/(When you roll a 20 on your attack roll with this magic weapon, the target takes an extra )(\d+)( damage of the weapon’s type.)/);
```
kakaroto, line 226: You don't need to use capture groups (parenthesis) around the first/last part of the text. You just need one value (the damage amount) so you only need one capture group. 
```js
const matchViciousWeapon = description.match(/When you roll a 20 on your attack roll with this magic weapon, the target takes an extra (\d+) damage of the weapon’s type./);
```
You can then refer to it with `matchViciousWeapon[1]` instead of `matchViciousWeapon[2]`

kakaroto (commented):
Looks good, though I have a couple of small styling suggestions.

PR #985, opened 2022-05-06
Bugfix for Artificier Arcane Jolt

Aeristoka:
Original code checked to make sure that the weapon being used did not have "Common" in the `item_type`, as Arcane Jolt requires a Magic Weapon.
image

However, when an Artificer uses their Infusion on a weapon, the weapon still has an `item_type` of "Common", but with modified text.
image

This change causes the "common (requires attunement)" check to be added and done first, so it supersedes the exclusion of strictly common items.

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -544,7 +544,7 @@ function handleSpecialWeaponAttacks(damages=[], damage_types=[], properties, set
         if (damages.length > 0 &&
             character.hasClassFeature("Arcane Jolt") &&
             character.getSetting("artificer-arcane-jolt", false) &&
-            item_type.indexOf(", Common") === -1) {
+            (properties["Infused"] || item_type.indexOf(", Common") === -1)) {
             damages.push(character._level < 15 ? "2d6" : "4d6");
             damage_types.push("Arcane Jolt");
         }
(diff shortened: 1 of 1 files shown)


kakaroto:
Humm... Is anything non-common magical? 
I also see it say "common" but the code checks for "Common", shouldn't it be all lower case  and use `item_type.toLowerCase()` for the comparison? 
I think we can also probably just do it with `/, common(?! \(requires attunement\))/i.test(item_type)` as that checks for `common` and can't be followed by `(requires attunement)` and ignores case.

src/dndbeyond/content-scripts/character.js
```diff
@@ -865,7 +866,7 @@ async function rollItem(force_display = false, force_to_hit_only = false, force_
 
         to_hit = handleSpecialGeneralAttacks(damages, damage_types, properties, settings_to_change, {to_hit, item_name});
 
-        to_hit = handleSpecialWeaponAttacks(damages, damage_types, properties, settings_to_change, {item_customizations, item_type, to_hit});
+        to_hit = handleSpecialWeaponAttacks(damages, damage_types, properties, settings_to_change, {item_customizations, item_type, to_hit, is_infused});
```
kakaroto, line 869: Humm... I'd do it a bit differently to keep the arguments to `handleSpecialWeaponAttacks` cleaner.
```js
if (is_infused) {
  properties["Infusion"] = "Yes"
}
```
Then check for the `properties["Infusion"] in the `handleSpecialWeaponAttacks`.
That'd be good for me.

Aeristoka: New code pushed, give it a look 😄

kakaroto:
Much appreciated!

PR #986, opened 2022-05-06
Add Specificity for Genie's Wrath Damage

Aeristoka:
Adds the specific damage type based on the Genie type the Warlock is associated with
image

image

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -509,7 +509,16 @@ function handleSpecialGeneralAttacks(damages=[], damage_types=[], properties, se
             character.hasClassFeature("Genie’s Vessel") &&
             character.getSetting("genies-vessel", false)) {
             damages.push(character._proficiency);
-            damage_types.push("Genie's Wrath");
+            if (character.hasClassFeature("Genie’s Vessel: Genie's Wrath (Dao)"))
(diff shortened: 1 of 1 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -509,7 +509,14 @@ function handleSpecialGeneralAttacks(damages=[], damage_types=[], properties, se
...
             damages.push(character._proficiency);
-            damage_types.push("Genie's Wrath");
+            if (character.hasClassFeature("Genie’s Vessel: Genie's Wrath (Dao)"))
+                damage_types.push("Genie's Wrath (Bludgeoning)")
+            else if (character.hasClassFeature("Genie’s Vessel: Genie's Wrath (Djinni)"))
+                damage_types.push("Genie's Wrath (Thunder)")
+            else if (character.hasClassFeature("Genie’s Vessel: Genie's Wrath (Efreeti)"))
+                damage_types.push("Genie's Wrath (Fire)")
+            else if (character.hasClassFeature("Genie’s Vessel: Genie's Wrath (Marid)"))
+                damage_types.push("Genie's Wrath (Cold)")
```
kakaroto, line 519: **just in case** something changes in the future, this should have an `else damage_types.push("Genie's Wrath")` as a fallback. Otherwise we risk unbalances arrays

kakaroto:
Thanks! 👍

PR #989, opened 2022-06-17
Fix for DnDBeyond changing Name Fields

Aeristoka:
Sensible, human-readable name fields abandoned
Replaced by ridiculous machine-generated field names


kakaroto:
the fix won't work, because the id will change the next time they rebuild the app
it's a random value generated by the build tools
need to find the name differently

I think replacing it with `.ddbc-character-tidbits__heading h1` might be a good working solution for now.

Aeristoka:
As discussed, fixed by @kakaroto in a (hopefully) more permanent way

PR #1014, opened 2022-08-25
Paladin: Auro of Protection not on Death Saves

### LINKED ISSUE #1013: Death Saving Throw for paladin with Aura of Protection incorrectly adds +5
A Paladin character with the Aura of Protection feature adds a +5 to all saving throws, which is all done automatically within D&D Beyond and Beyond20 reflects this.

However, Beyond20 adds a button to automatically roll a death saving throw within the Death Saving Throws tally box that appears when at 0hp. This roll also adds the +5 bonus, but this SHOULD NOT be added in this case, as the paladin is unconscious, and the Aura of Protection should not be active in this case.

Aeristoka:
Fixes #1013


kakaroto:
Looks good! I would have fixed it in the same way 🙄

PR #1015, opened 2022-08-30
Handle Dragon Wing * Ranged Weapons + Flail of Tiamat Damages

Aeristoka:
Lazy setup by DnDBeyond shows all damages for each weapon
Pare down to single additional damage
Allow setting "note" on a weapon to specify damage
"note" must be set to a Damage type available to be infused on the weapon to apply (fallback is "Infused Damage")

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -848,6 +848,29 @@ async function rollItem(force_display = false, force_to_hit_only = false, force_
             }
         }
 
+        // Handle Dragon Wing * Ranged Weapons
+        if (item_name.includes("Dragon Wing") || item_name === "Flail of Tiamat") {
+            damages.splice(2,damages.length - 2);
+            let possible_damages = damage_types.splice(1,damage_types.length - 1);
+            const damage_type = properties.Notes && possible_damages.reduce((found, damage) => {
+                if (found) return found;
+                if (properties.Notes.toLowerCase().includes(damage.toLowerCase())) return damage;
+                return null;
+            }, null);
+            if (damage_type) {
+                damage_types.push(damage_type);
+            }
+            else if (item_name.includes("Dragon Wing")) {
+                damage_types.push("Infused");
+            }
+            else if (item_name === "Flail of Tiamat") {
+                damage_types.push("Chosen Type");
+            }
+            else {
+                damage_types.push("Extra");
+            }
+        }
+
         const weapon_damage_length = damages.length;
         
         // If clicking on a spell group within a item (Green flame blade, Booming blade), then add the additional damages from that spell


Aeristoka:
Default pare down:
image

Note Set for specific type:
image

Aeristoka:
May add similar handling for Flail of Tiamat to this (and generalize the code more as well).

src/dndbeyond/content-scripts/character.js
```diff
@@ -848,6 +848,21 @@ async function rollItem(force_display = false, force_to_hit_only = false, force_
             }
         }
 
+        // Handle Dragon Wing * Ranged Weapons
+        if (item_name.includes("Dragon Wing") || item_name === "Flail of Tiamat") {
+            damages.splice(2,damages.length - 2);
+            let possible_damages = damage_types.splice(1,damage_types.length - 1);
+            if (properties.Notes && possible_damages.includes(properties.Notes)) {
```
kakaroto, line 855: The `includes` here means it needs to match perfectly, so if there's a note like "Damage is of type fire", it won't work.
I'd rewrite this into something like this : 
```js
   const damage_type = properties.Notes && possible_damages.reduce((found, damage) => {
         if (found) return found;
         if (properties.Notes.toLowerCase().includes(damage.toLowerCase())) return damage;
         return null;
   }, null);
   if (damage_type) {
      damage_types.splice(1, 0, damage_type);
  } else if () ....
}
```

src/dndbeyond/content-scripts/character.js
```diff
@@ -848,6 +848,21 @@ async function rollItem(force_display = false, force_to_hit_only = false, force_
...
+                damage_types.splice(1,0,properties.Notes);
+            }
+            else if (item_name.includes("Dragon Wing")) {
+                damage_types.splice(1,0,"Infused");
+            }
+            else if (item_name === "Flail of Tiamat") {
+                damage_types.splice(1,0,"Chosen Type");
+            }
```
kakaroto, line 863: This is prone to possible future developer error, if we add a new handled type, but there's no `else` for it, you end up with `damages` array having one more item than `damage_types` array. Having a fallback `else` would prevent that.

kakaroto (commented):
Thanks! I just noticed a couple of issues that can be fixed to improve the handling. Looks good otherwise 👍

src/dndbeyond/content-scripts/character.js
Aeristoka: I think you mean this is prone to ME making errors 😄

kakaroto: lol, no! 😁 
It's always a good idea to have an `else` when a situation where all the statements fail could cause issues!

Aeristoka: We all now what you meant... 😄

Issue #1022, opened 2022-09-30
Chromatic Orb difficult to cast with advantage / disadvantage


matthewbstroud:
**Describe the bug**
Trying cast chromatic orb with advantage launches a damage type dialog, which doesn't automatically honor the advantage/disadvantage modifier.

**To Reproduce**
1. Have Chromatic Orb in your spell list 
2. Hold down shift and click the attack roll.
![image](https://user-images.githubusercontent.com/17620888/193365728-0a730726-cda5-4074-870e-3aefe8dd5488.png)
3. A popup will show asking for damage type.  
![image](https://user-images.githubusercontent.com/17620888/193365820-a3db8bb6-f7c8-4b20-ba3e-a3b209349c8d.png)
4. Choose a damage type and click okay.
5. A single d20 will roll.

**Expected behavior**
You should only be prompted for a damage type when rolling damage for chromatic orb, since if you miss the damage is meaningless.  _Or_ Clicking the attack roll should honor the advantage/disadvantage modifiers.

**Workaround**
Holding shift when you click the Roll button in the dialog does make the advantage roll work.

**Browser Info (please complete the following information):**
 - Edge
 - Version 105.0.1343.53 (Official build) (64-bit)

**Errors**
![image](https://user-images.githubusercontent.com/17620888/193366588-78b188ab-bd71-4deb-b061-c4edc4864862.png)

kakaroto:
Nice find, thanks for the report. The popup dialog is definitely interfering here!

Aeristoka:
To clarify, no, it should absolutely do the dialogue to specify damage here, since it's going to give you the "Roll Damages" button, as it does when you use DnDBeyond Digital Dice. If it didn't do it here, we've lost the easy facility to do that.

Yes, it absolutely should respect advantage all the way through.

the utils.ts error is unrelated to Beyond20.

matthewbstroud:
![image](https://user-images.githubusercontent.com/17620888/193415887-8a5873b0-6d81-4f63-bdd1-a5369fe6fd23.png)

Aeristoka:
> ![image](https://user-images.githubusercontent.com/17620888/193415887-8a5873b0-6d81-4f63-bdd1-a5369fe6fd23.png)

Context of what you were doing when you saw this?

matthewbstroud:
Clicking on the Chromatic Org to Hit button over and over.

matthewbstroud:
so, this seems to fix the issue
```
[...]

PR #1023, opened 2022-10-01
Issue-1022: Chromatic Orb difficult to cast with advantage / disadvantage

matthewbstroud:
The issue was a race condition in code.  If you were building an attack roll that could launch the spell damage type dialog, the global key_modifiers could change between the initial state and the final state.

I modified the buildAttackRoll to include the advantage type in the arguments passed to sendRoll.  This solves the race condition.

Fixes: https://github.com/kakaroto/Beyond20/issues/1022

### DIFF
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -124,7 +124,8 @@ async function buildAttackRoll(character, attack_source, name, description, prop
         "description": description,
         "rollAttack": force_to_hit_only || !force_damages_only,
         "rollDamage": force_damages_only || (!force_to_hit_only && character.getGlobalSetting("auto-roll-damage", true)),
-        "rollCritical": false
+        "rollCritical": false,
+        "advantage": getAdvantageType(character)
     }
     if (to_hit !== null)
         roll_properties["to-hit"] = to_hit;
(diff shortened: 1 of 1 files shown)


matthewbstroud:
@kakaroto I had never debugged a chrome extension prior to today.  I was working with this file
image

I do not fully understand how the project is built, but found the matching code in the modified file.

matthewbstroud:
I only tested this fix in Chrome.  (well Edge)

kakaroto:
Thanks for contributing the fix. Looks good at first glance, but I'll need to properly review and test this before I merge it.

> I do not fully understand how the project is built, but found the matching code in the modified file.

There are instructions on how to build in the README file, it's basically running `npm install && npm run build`. 
There's also a `npm run gulp` command you can run so it watches for any source changes and automatically rebuilds the output, which is how you should do it during development.

matthewbstroud:
Thanks @kakaroto, I installed node.js and can build now.

kakaroto:
So, your code was fine, if the only purpose was to allow advantage handling via hotkey, but when I looked at the code, there were many other situations where the hotkeys would be ignores. Imagine for example that the user has a hotkey for "apply great weapon master" or "add rage damage", or even a situation where you have the hotkey for advantage but whisper is queried first (or a hotkey for whispering but advantage is queried first).
The only viable solution for handling those use cases that I found were to add a way to pause the handling of key events so the `key_modifiers` are retained until the add of all processing, then queue the events received in the meantime and resume the handling of hotkeys while processing the pending events.
This had the side effect of requiring a major refactor of the code, because so far, if you get a query and you click 'cancel', it just didn't resolve the async call, so nothing happened, but if I were to keep it that way, we'd end up never resuming the handling of hotkey events and our queued events would become enormous, leading to memory leaks and broken behavior. 
This meant that all queries had to return whether it was cancelled or not, and it had a domino effect of recursively checking every caller until we get to the root event handler, making sure a cancellation is properly handled.

See related commits
https://github.com/kakaroto/Beyond20/commit/9686b75e8bbb8b9b6a5f227f003c7dfc5f8275c2
https://github.com/kakaroto/Beyond20/commit/72782db4efce660a1d6921f848e93ed55e2cfb76
https://github.com/kakaroto/Beyond20/commit/c6a31de530e5f907e982e2fe053947b8f2a7677a
https://github.com/kakaroto/Beyond20/commit/fd4403942de84d42bb678a36d5abbbbee7b9276d

matthewbstroud:
That is fine.  I saw the same issue, but wasn't looking to get that deep.
Implementing a lock on key modifiers shouldn't be that difficult.

At this point, I will bow out as my ideas do not seem to be welcome.

PR #1025, opened 2022-10-08
Fix for Brutal/Savage + GWF + Polearm Master

Aeristoka:
Edge case application of GWF to Brutal Critical/Savage Attacks on a Polearm Master - Bonus Attack

### DIFF
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -264,8 +264,9 @@ async function buildAttackRoll(character, attack_source, name, description, prop
                     } else {
                         // Apply great weapon fighting to brutal damage dice
                         if ((character.hasClassFeature("Great Weapon Fighting", true) || character.hasFeat("Great Weapon Fighting", true)) &&
-                            properties["Attack Type"] == "Melee" &&
-                            ((properties["Properties"].includes("Versatile") && character.getSetting("versatile-choice") != "one") || properties["Properties"].includes("Two-Handed"))) {
+                            ((properties["Attack Type"] == "Melee" &&
+                            ((properties["Properties"].includes("Versatile") && character.getSetting("versatile-choice") != "one") || properties["Properties"].includes("Two-Handed"))) ||
+                            name == "Polearm Master - Bonus Attack")) {
                             brutal_dmg += "ro<=2"
                         }
                         crit_damages.push(brutal_dmg);


src/dndbeyond/base/utils.js
```diff
@@ -264,8 +264,9 @@ async function buildAttackRoll(character, attack_source, name, description, prop
                     } else {
                         // Apply great weapon fighting to brutal damage dice
                         if ((character.hasClassFeature("Great Weapon Fighting", true) || character.hasFeat("Great Weapon Fighting", true)) &&
-                            properties["Attack Type"] == "Melee" &&
-                            ((properties["Properties"].includes("Versatile") && character.getSetting("versatile-choice") != "one") || properties["Properties"].includes("Two-Handed"))) {
+                            (properties["Attack Type"] == "Melee" &&
+                            ((properties["Properties"].includes("Versatile") && character.getSetting("versatile-choice") != "one") || properties["Properties"].includes("Two-Handed")) ||
+                            name == "Polearm Master - Bonus Attack")) {
```
kakaroto, line 269: I think the `||` should be before that last parenthesis instead of after it. 
`if (GWF && (melee && ((versatile && not-one-handed) || two-handed) || polearm-master)`

You basically have a `A && B || C` (`melee && versatile-stuff || polearm-master`) which is unreliable/confusing in how it would behave. It should be `(A && B) || C` or `A && (B || C)`

Aeristoka: Fixed.

I blame the horrifying line breaks in that if statement 😅

kakaroto: Yeah, it nearly broke my brain

src/dndbeyond/base/utils.js
```diff
@@ -264,8 +264,9 @@ async function buildAttackRoll(character, attack_source, name, description, prop
                     } else {
                         // Apply great weapon fighting to brutal damage dice
                         if ((character.hasClassFeature("Great Weapon Fighting", true) || character.hasFeat("Great Weapon Fighting", true)) &&
-                            properties["Attack Type"] == "Melee" &&
-                            ((properties["Properties"].includes("Versatile") && character.getSetting("versatile-choice") != "one") || properties["Properties"].includes("Two-Handed"))) {
+                            (properties["Attack Type"] == "Melee" &&
+                            ((properties["Properties"].includes("Versatile") && character.getSetting("versatile-choice") != "one") || properties["Properties"].includes("Two-Handed"))) ||
+                            name == "Polearm Master - Bonus Attack") {
```
kakaroto, line 269: arg, it's still an `A && B || C` situation, but now it's `GWF && melee-and-stuff || polearm` . Would need to add parenthesis around the whole second part.

Aeristoka: And this is a good example of why I shouldn't code things when I can't pay attention... I think it's ok now...

kakaroto: > And this is a good example of why I shouldn't code things when I can't pay attention... I think it's ok now...

Oh, I don't think it would have made any difference. It's really breaking my brain... 
In the end I simplified it to this line and it started to make sense :
`if (GWF && ((melee && ((versatile && not-one-handed) || two-handed)) || polearm)) {`

I think saving some of those as variables and doing a simpler if statement would have been quite helpful in this case.

Issue #1029, opened 2022-10-24
Navigating to a character sheet from "My characters" using View button fails to load extension


kakaroto:
From Discord `@abramTheMan` :
> when you're on the my-characters page on https://www.dndbeyond.com/my-characters and then click View instead of doing a normal redirect it does some kinda fancy React.js redirect where it loads the data on the next page but it never triggers the onContentLoaded event so Beyond20 never gets initialized. you then have to hit refresh if you want Beyond20 to work with your sheet.

kakaroto:
Fixed by #1030

PR #1030, opened 2022-10-24
Navigating to a character sheet from "My characters" using View button fails to load extension

IsaacAbramowitz:
Resolve issue #1029 by adding the 'My Characters' page to the manifest.

Doing this seemed to resolve things for me when I tested it.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -60,7 +60,8 @@
 	"content_scripts": [
 		{
 			"matches": [
-				"*://*.dndbeyond.com/*characters/*"
+				"*://*.dndbeyond.com/*characters/*",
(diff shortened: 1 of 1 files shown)


Aeristoka:
Pinging this to @kakaroto for review, appreciate the PR!

kakaroto:
Thanks! Yep, that was the fix I was going to do for it as well. I just need to test that it works as expected and doesn't have any side effects, then it should be mergeable.
Thanks! 😁

kakaroto:
Confirmed working and not seeing any negative side effects either! 👍 
Though it's missing a similar change in the firefox manifest. I'll do that now.

PR #1035, opened 2022-11-01
Fix issue regarding the restrained condition being mispelled on the VTT

IsaacAbramowitz:
unknown
`Restrained` is being spelled out as `RRestrained` on the virtual table top when updating conditions on a character sheet.

The issue stemmed from the code taking the text inside a `div` and using that as the text for the condition. However the Restrained div had 3 `divs` inside of it, one of which had an `svg` with the letter `R` in it. This updated PR takes the specific div with the desired text instead

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -2095,7 +2095,7 @@ function injectRollButton(paneClass) {
         let exhaustion_level = $(".ct-condition-manage-pane__condition--special .ct-number-bar__option--active,.ct-condition-manage-pane__condition--special .ddbc-number-bar__option--active").text();
         const conditions = [];
         for (let cond of j_conditions.toArray()) {
-            var condition_name = $(cond).find(".ct-condition-manage-pane__condition-name").text();
+            const condition_name = $(cond).find(".ct-condition-manage-pane__condition-name").text();
             conditions.push(condition_name);
         }
         if (exhaustion_level == "")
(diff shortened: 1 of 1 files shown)


Aeristoka:
if only DnDBeyond would stop changing formatting stuff all the time...

Very nice catch!

src/dndbeyond/content-scripts/character.js
```diff
@@ -2094,8 +2094,10 @@ function injectRollButton(paneClass) {
         const j_conditions = $(".ct-condition-manage-pane .ct-toggle-field--enabled,.ct-condition-manage-pane .ddbc-toggle-field--is-enabled").closest(".ct-condition-manage-pane__condition");
         let exhaustion_level = $(".ct-condition-manage-pane__condition--special .ct-number-bar__option--active,.ct-condition-manage-pane__condition--special .ddbc-number-bar__option--active").text();
         const conditions = [];
-        for (let cond of j_conditions.toArray())
-            conditions.push(cond.textContent);
+        for (let cond of j_conditions.toArray()) {
+            var condition_name = $(cond).find(".ct-condition-manage-pane__condition-name").text();
```
kakaroto, line 2098: The variable here should be a `const` rather than `var`

kakaroto (commented):
Ah good catch, thanks for finding it and providing the fix for it! 👍

kakaroto:
> if only DnDBeyond would stop changing formatting stuff all the time...

It's because that SVG image has an "R" in it 😁 
image

PR #1039, opened 2022-11-28
Additional Fixes for MotM Aasimar Radiant Soul

Aeristoka:
MotM Radiant Soul uses Proficiency Bonus instead of Level for damage


kakaroto (approved):
NOW it LGTM 😝

PR #1041, opened 2022-12-02
Add limited support for Goliath "Large Form" trait

mjgiarlo:
This commit adds a partial implementation of the racial trait called "Large Form" that has been introduced for Goliaths as of the latest One D&D playtest. The effects of this bonus action are to (1) grant ADV on STR checks for the duration of the effect, (2) increase the size of the character to Large, and (3) increase the speed of the character by 10 feet. Only the first of these is implemented because it is unclear whether it is even possible (let alone advisable) to temporarily bump the size and speed of the character.

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -773,6 +773,12 @@ const character_settings = {
         "type": "bool",
         "default": true
     },
+    "goliath-large-form": {
+        "title": "Goliath: Large Form",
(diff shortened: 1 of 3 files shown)


Aeristoka:
I believe we won't be implementing anything from One DnD until it's actually official on DnDBeyond (ie, is on character sheets).
This has been the long-standing policy of Beyond20, including ALL UAs.

Aeristoka:
@kakaroto may correct me if I'm misremembering.

mjgiarlo:
@Aeristoka No problem. I'll use this locally and if this trait ultimately lands in DNDB, I can always re-submit.

Aeristoka:
> @Aeristoka No problem. I'll use this locally and if this trait ultimately lands in DNDB, I can always re-submit.

Sounds good!

Issue #1048, opened 2023-01-14
Add hotkey option for display to vtt


kakaroto:
Having a hotkey option that would cause an attack roll to become a "Display" roll instead would be beneficial, especially in monster statblocks where we don't have both buttons available (and adding both would clutter the view).

PR #1054, opened 2023-01-31
Add hotkey to display attack details to the VTT instead of rolling it

IsaacAbramowitz:
This pull request addresses issue #1048.

Not 100% sure on how to properly handle creating hotkeys but this seems to work and I was bored towards the end of my day at work


IsaacAbramowitz:
Note: I have not tested on  Foundry or any other VTT besides Roll20 so there's probably stuff missing

kakaroto:
Hi, thanks for the PR! Sorry for the late response, I've actually fixed this differently 6 weeks ago but the last few weeks have been a bit overwhelming and I didn't get a chance to even just push the code to github.
There were some use cases that weren't properly handled in your proposed changes, so I implemented it differently. Thanks for the contribution though!

You can see the fix here:
https://github.com/kakaroto/Beyond20/commit/067bb044427022d2962e097147af7f98ae87397d

PR #1062, opened 2023-05-01
allow rendered rolls to come from custom domains

watcherdm:
Not as complicated as I imagined on the extension side, really just opened the rendered-roll event listener on the dom. The allows my web application to send dom events up with roll info. Working on documenting the roll data structure in another PR.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -10,7 +10,8 @@
 		"*://beyond20.kicks-ass.org/roll",
 		"*://app.roll20.net/editor/",
 		"*://*.dndbeyond.com/*",
-		"*://*.forge-vtt.com/game"
+		"*://*.forge-vtt.com/game",
+		"*://*.osrbeyond.com/*"
 	],
     "optional_permissions": ["*://*/"],
 	"icons": {
--- a/manifest.json
+++ b/manifest.json
@@ -10,8 +10,7 @@
 		"*://beyond20.kicks-ass.org/roll",
 		"*://app.roll20.net/editor/",
 		"*://*.dndbeyond.com/*",
-		"*://*.forge-vtt.com/game",
-		"*://*.osrbeyond.com/*"
+		"*://*.forge-vtt.com/game"
 	],
     "optional_permissions": ["*://*/"],
 	"icons": {
--- a/src/generic-site/content-script.js
+++ b/src/generic-site/content-script.js
@@ -1,6 +1,7 @@
 console.log("Beyond20: Custom Domain module loaded.");
 
 var settings = getDefaultSettings();
+var characterInternal = null;
 function handleMessage(request, sender, sendResponse) {
     console.log("Received Beyond20 message : ", request);
     if (request.action == "settings") {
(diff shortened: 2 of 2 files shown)


manifest.json
```diff
@@ -10,7 +10,8 @@
 		"*://beyond20.kicks-ass.org/roll",
 		"*://app.roll20.net/editor/",
 		"*://*.dndbeyond.com/*",
-		"*://*.forge-vtt.com/game"
+		"*://*.forge-vtt.com/game",
+		"*://*.osrbeyond.com/*"
```
kakaroto, line 14: Unfortunately, we can't add new things to the manifest, as it will basically disable the extension for everybody until they accept the new permissions. Last time we did that, we left a few hundred thousand users at once. Instead, we use the optional permissions system instead, and the url should be added to the `SUPPORTED_VTTS` constant.

src/generic-site/content-script.js
```diff
@@ -34,5 +35,38 @@ function updateSettings(new_settings = null) {
...
+        event.disconnect();
+    })
+}
+
+function handleIncomingSettings({ character, action, type} = {}) {
+    characterInternal = character;
+    chrome.runtime.sendMessage({
+        action: "hp-update",
+        character
+    })
```
kakaroto, line 55: incoming settings should send an hp-update. I also don't think a settings event usually includes a character 🤔

src/generic-site/content-script.js
```diff
@@ -34,5 +35,38 @@ function updateSettings(new_settings = null) {
...
+    chrome.runtime.sendMessage({
+        action: "settings",
+        type: "character",
+        settings: character.settings,
+        id: character.id
+    })
+}
+
+var registered_events = [];
+registered_events.push(addCustomEventListener("rendered-roll", sendRollToGameLog));
```
kakaroto, line 65: Technically, this sends a message whether it's a `roll` or `rendered-roll` or `hp-update` makes no difference. The function name shouldn't be "sendRollToGameLog" either since it's sending to roll20, not the game log.
So I think the event should be `addCustomEventListener("send-message", sendMessageToBeyond20)` instead.

src/generic-site/content-script.js
```diff
@@ -34,5 +35,38 @@ function updateSettings(new_settings = null) {
     }
 }
 chrome.runtime.onMessage.addListener(handleMessage);
+
+function sendRollToGameLog(message) {
+    console.log(message)
+    chrome.runtime.sendMessage(message)
```
kakaroto, line 41: I think we should probably have a whitelist of which actions are accepted, such as `roll`, `rendered-roll` and `hp-update`. If the `message.action` is different, it should probably block it, since it could be used to affect user settings (such as adding a malicious site to the list of accepted domains), or other internal messaging actions between beyond20 and the background script (can't remember what they are, or if they'd be an issue to let any site send those messages to the extension).

src/generic-site/content-script.js
```diff
@@ -34,5 +35,38 @@ function updateSettings(new_settings = null) {
...
+    chrome.runtime.sendMessage({
+        action: "hp-update",
+        character
+    })
+    chrome.runtime.sendMessage({
+        action: "settings",
+        type: "character",
+        settings: character.settings,
+        id: character.id
+    })
```
kakaroto, line 61: Note that the `settings` action message is always for the global settings, never for character settings, since the character specific settings are always present in the roll request (through request.character.settings).

This is a situation I hadn't thought of though... what about character settings 🤔 Where should they be stored, what would the `id` represent (as it's not a DDB character sheet ID anymore, and we don't want conflicts/collissions).
I think some of those settings should be stored on the site's own `localStorage` if possible, rather than in Beyond20's extension settings... but if we do, we lose that info from the save/load settings option in Beyond20.

Also a site shouldn't be able to send an update to the global settings. That should already be handled on its own via the Beyond20 popup or settings alert. 

I might need to come up with a strategy to allow such sites though to tell the extension "this is a character sheet, and here are the character settings" so that when you click on the beyond20 icon in the address bar, it tells you it's a character, what its name is and lets you customize per-character settings, rather than show you the generic settings popup.

It would also be nice if the extension upon load did a search for `$(".beyond20-placeholder")` for example and injects its dice icon there for users to quick access the settings (like we do near the chat on Roll20 or Foundry, or next to the character name in DDB.

kakaroto (commented):
Thanks for the contribution! I figured this would be simple, but not that simple! Although I see a few situations that need to be thought out a bit more with regards to character settings and the popup.
The biggest challenge will be documenting everything, which is what scares me!

kakaroto (commented):
I've merged this and made some small changes. I disabled the ability to send `settings` messages and changed `send-message` into `SendMessage` and checked against only allowed events to forward.
I've also started documenting the messaging API, though it's not yet complete.
https://beyond20.here-for-more.info/api

Thanks!

PR #1080, opened 2023-08-08
Update beyond20.js

koalana51066:
Add targets to chat messages

### DIFF
--- a/FVTT-module/beyond20/beyond20.js
+++ b/FVTT-module/beyond20/beyond20.js
@@ -1,8 +1,32 @@
-
-
+const moduleName = "beyond20";
 class Beyond20 {
+    static skipTargets = [
+        "Acrobatics",
+        "Animal Handling",
+        "Arcana",
+        "History",
+        "Insight",
+        "Initiative",
+        "Investigation",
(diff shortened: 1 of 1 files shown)


koalana51066:
The change to encounter.js is probably in the wrong place.  The idea is to send the name of the monster from the tracker card on the left.  If there are multiple copies of the same monster, they get appended with (A), (B), etc.  Since I try to make my tokens in Foundry the same, I thought it would be nice to know which one was making the roll even though their stat blocks are the same.  This is just for having an accurate record in case there are questions later.

My code checks to see if the tracker card name (with "is-turn" class) contains the summary card name.  If not, the summary card name is sent.  I so, the tracker card name is sent.

koalana51066:
The change to findToken() was because I thought it would be useful when you don't have an associated request from which to get the name.  I wound up never using it.  That could be removed.

kakaroto:
Hi @koalana51066,
Sorry for the delay in reviewing this, I just wanted you to know that I've seen it and will get to fully reviewing when I get the time to work on Beyond20. 
And thank you for your contribution, it's very much appreciated!

PR #1096, opened 2023-12-11
Support for Healing NPC Attacks

### LINKED ISSUE #1070: Add support for healing NPC attacks
NPC attacks only recognize damage from the "X damage" regexp.
We could add also "regains X hit points" as a recognizable one for Healing damage.
Example in the Unicorn: https://www.dndbeyond.com/monsters/17042-unicorn

ofMoons:
Fixes #1070

Matches text for healing-based actions in monster stat blocks to allow rolling on those abilities.

### DIFF
--- a/src/dndbeyond/base/monster.js
+++ b/src/dndbeyond/base/monster.js
@@ -472,7 +472,7 @@ class Monster extends CharacterBase {
             hit = description.slice(hit_idx);
         // Using match with global modifier then map to regular match because RegExp.matchAll isn't available on every browser
         const damage_regexp = new RegExp(/([\w]* )(?:([0-9]+)(?!d))?(?: *\(?([0-9]*d[0-9]+(?:\s*[-+]\s*[0-9]+)?(?: plus [^\)]+)?)\)?)? ([\w ]+?) damage/)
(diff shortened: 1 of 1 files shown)


src/dndbeyond/base/monster.js
```diff
@@ -489,6 +491,13 @@ class Monster extends CharacterBase {
                 damage_types.push(dmg[4]);
             }
         }
+        for (let dmg of healing_matches) {
+            const damage = dmg[2] || dmg[1];
+            if (damage) {
+                damages.push(damage.replace("plus", "+"));
+                damage_types.push("Healing");
```
kakaroto, line 498: Make the `(?:temporary)?` into a capturing group and check for `dmg[3]` value, so if it's temp hp, the damage type should be "Temp HP" (It's what we used elsewhere, see `src/dndbeyond/content-scripts/charater.js` line 1504)

kakaroto (changes requested):
It looks good, and thanks for adding the the temporary hit points check to it.
I just had one small comment to make this mergeable.
Thanks!

src/dndbeyond/base/monster.js
ofMoons: Thanks for the guidance, I've made the update.

kakaroto (approved):
Looks good, thanks for the contribution!

PR #1097, opened 2023-12-30
Support for War Caster feat

rthorpeii:
Adds support for the advantage granted to concentration check from the War Caster feat.

To avoid having multiple popups, if the character also has Wizard Bladesong, the response from the Bladesong confirmation will also trigger the advantage from War Caster.


kakaroto (approved):
Nice work, thanks!

PR #1115, opened 2024-04-27
initial patch R20discord-optional-perms

jtbg:


jtbg:
(ignore, I PR'd against the wrong upstream)

Issue #1130, opened 2024-06-04
The "Beyond20*" events of the DOM API appear to not be triggered on DnD Beyond itself


carrierfry:
**Describe the bug**
The `Beyond20*` events (like `Beyond20_Loaded(settings)`) do not get triggered when on DnD Beyond

**To Reproduce**
1. Create extension or userscript for DnD Beyond character sheet with following code and enable/load it:
```js
function addBeyond20EventListener(name, callback) {
    const event = ["Beyond20_" + name, (evt) => {
        const detail = evt.detail || [];
        callback(...detail)
    }, false];
    document.addEventListener(...event);
    return event;
}

addBeyond20EventListener("Loaded", settings => {
    console.log("Beyond20_Loaded");
    console.log(settings);
});

console.log("Added event listener");
```
2. Open character sheet on DnD Beyond
3. Open DevTools
4. See missing console logs

**Expected behavior**
The `Beyond20*` events should work even on DnD Beyond itself to make 3rd party extensions be able to listen to those events there

**Screenshots**
![image](https://github.com/kakaroto/Beyond20/assets/20256939/271a6098-8c77-4054-895f-933abb69793c)

**Browser Info (please complete the following information):**
 - Browser Google Chrome
 - Version 125.0.6422.141

kakaroto:
While implementing this I realized that it was already implemented as part of https://github.com/kakaroto/Beyond20/issues/833
You can see it in the commit: https://github.com/kakaroto/Beyond20/commit/54037c7065861140c764019a92888b650973e766
The issue however is that it's sending the requests as is using the action name as the event's name instead of the names described in the API documentation (i.e: `Beyond20_hp-update" instead of "Beyond20_UpdateHP" and "Beyond20_roll" instead of "Beyond20_Roll").

I've now fixed it to be consistent with the documented DOM API.

PR #1135, opened 2024-06-05
Added Beyond20_Loaded and Beyond20_NewSettings events to DnD Beyond itself

carrierfry:
As described in #1130 all the custom events are missing on DnD Beyond itself, making it hard to interact with the DOM API when things like settings are important. Since I am not too familiar with Beyond20's code yet, I only added the event triggers for Beyond20_Loaded and Beyond20_NewSettings for now.

Let me know if something is wrong or missing!


kakaroto (approved):
Looks good to me, thanks!

PR #1136, opened 2024-06-05
Stop settings button overlaying tidbits DDB

### LINKED ISSUE #1129: [minor] Overlapping elements at small browser widths
**Describe the bug**
At small browser widths, the Beyond20 button overlaps with other elements:

image

**To Reproduce**
1. Go to any character sheet
2. Resize the browser to around 1000px wide

**Expected behavior**
Maybe the other buttons could collapse to just icons? I'm not sure. But the overlapping is annoying.

 - Browser Chrome
 - Version Latest

lindenmckenzie:
Fixes #1129

Overrides tidbit min-width:300px which is causing the overlap between the settings button and the tidbit section. This is happening between 768px and around 812px. Chose to use CSS override due to wanting to keep in line with DDBs general approach for styling buttons at tablet width rather than going straight down to an icon. 

Testing screenshot taken using Chrome in responsive mode at 768px wide:

Screenshot 2024-06-05 at 20 04 34

### DIFF
--- a/src/extension/beyond20.css
+++ b/src/extension/beyond20.css
@@ -392,3 +392,8 @@ span.dice-roll.beyond20-tooltip-content {
 .ct-beyond20-icon {
     vertical-align: middle;
 }
+
+/* Stops Beyond20 buttons overlaying ddb tidbits */
+.ct-character-header-info .ddbc-character-tidbits {
+    min-width:auto;
+}


kakaroto:
Sounds good, thanks! I was thinking of having the button become just the icon, but your reasoning for your approach makes sense 👍

Issue #1127, opened 2024-05-23
Add support for FoundryVTT v12


JeshinBBRP:
Currently you can still link/cast spells but skill checks and weapon attacks are not sending to the VTT from DNDBeyond and there is no error message.

To Reproduce
Go to any character sheet
Link to FoundryVTT
Attempt to roll skill check
Attempt to roll weapon attack
Attempt to cast firebolt cantrip

Date of issue was 24-05-22 with newest Chrome and newest Beyond20 versions.

tickleyfeet29:
Same issue - Foundry v12.  However, damage rolls go through, weirdly.

surestart:
I'm having the same problem since updating to v12, and disabling all modules doesn't seem to affect the problem. Attacks don't appear in the Foundry chat log, but they do cause dice from Dice So Nice to roll on the screen with the same values as the rolls happening on DNDbeyond, so Foundry _is_ receiving the data in some form. I am also seeing damage rolls work normally, as well as spells linking when the description of the spell is being sent, but not when the spell is asking for a save or sending an attack roll.

kakaroto:
So weird, I wasn't expecting posting chat messages to break with v12.
V12 was just released though, we always recommend people wait a while before switching to new foundry versions because every new version seems to break something in existing modules. I haven't even tried v12 yet, but once I do, I'll tackle the v12 compatibility in beyond20.
Thanks for the reports!

JStarks85:
I am running v2.9.4 and Foundry V12.327.  

Attack rolls and skill checks do not appear in the chat when rolled.

I tested Roll 20 and it works there.

I just updated Beyond 20 Companion Module to 1.5.2. and the issue still persists.

malaandm:
Same:
I am running v2.9.4 and Foundry V12.325.
No d20 rolls (Attack rolls, ability, saving throw, and skill checks) appear in the chat when rolled.
I am using Chrome, and on The Forge

albertholaso:
In case it's related, the Dndbeyond Monster rolls also do not come through

xuri-ajiva:
The issue seams to be some kind of chat message parsing from the D&D 5e System:
```js
SyntaxError: Expected [%*/], [+\-], end of input, or whitespace but "u" found. (at foundry-esm.js:62945:22)
    at new peg$SyntaxError (foundry-esm.js:62945:22)
    at peg$buildStructuredError (foundry-esm.js:63320:14)
[...]

PR #1144, opened 2024-07-09
Foundry 12 / DnD5e expects the term to have operator or number set

AlanWaiss:
In troubleshooting Beyond20 with Foundry v12, I found that the terms expect an "operator" for "OperatorTerm" terms or "number" for "NumericTerm" terms.

I'm also only adding the + OperatorTerm if the previous term isn't already an OperatorTerm and using foundry.utils.isNewerVersion when available since the global isNewerVersion is deprecated.

[Issue 1127](https://github.com/kakaroto/Beyond20/issues/1127)

### DIFF
--- a/src/fvtt/page-script.js
+++ b/src/fvtt/page-script.js
@@ -1,9 +1,11 @@
 var settings = null;
 var extension_url = "/modules/beyond20/";
 var fvttVersion = game.version || game.data?.version;
+//isNewerVersion is deprecated in Foundry 12
+const fvtt_isNewer = window.foundry && foundry.utils && foundry.utils.isNewerVersion ? (v1, v0) => foundry.utils.isNewerVersion(v1, v0) : isNewerVersion;
 // v10 uses .document for Placeables (tokens), or the Base object itself, instead of .data field
 // We use the new fields to avoid spamming the log with deprecation warnings
-const docData = (doc) => isNewerVersion(fvttVersion, "10") ? doc.document || doc : doc.data;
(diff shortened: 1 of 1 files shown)


SethStanowski:
Just tested this on my Foundry VTT server works great, needed this fix thank you

FVTT Version: 12
Build: 328 
Game System: dnd5e v 3.3.0

tested on chrome version 126.0.6478.182 (Official Build) (64-bit)

kakaroto (approved):
Finally got the time/wits to review this. My main worry is if it causes issues on 0.8.x to v11, so I'll want to test it out first before merging, but code wise, it looks good and makes sense to me.
It looks like the change in v12 is that they stopped using `expression` and now use `operator` or `number` fields instead, for some reason. Wondering if it could have been fixed just by adding those two fields. (i believe multiple + operators in succession isn't an issue, but i like your solution to it) 
The only possible issue is that i don't remember what those roll parts might contain and if we could have two formula parts with no operators in between since you remove the autoadd of operator, and only apply it during adding a numeric value.
Will test and if nothing breaks on older versions, I'll merge.
Thanks again and apologies for the delay in getting it looked at!

kakaroto:
Thanks, it's merged! I tested on v9 and it turns out even in v9, it was using `operator` and `number` 🤦 I've slightly [cleaned it up](https://github.com/kakaroto/Beyond20/commit/f63aa7adff9b676ad33ae83bc784f8328dcfd76a) and merged!

PR #1154, opened 2024-07-26
added fix for deactivequickrolls for initiative

dmportella:
the selector was returning two elements and it seemed that this would trigger multiple clicks

@kakaroto the selector was returned 2 elements, i changed it to return now only one, and it seems to work with everything, desktop, tablet and mobile.

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -2312,8 +2312,8 @@ function deactivateQuickRolls() {
     const spells = $(".ct-spells-spell .ct-spells-spell__action,.ddbc-spells-spell .ddbc-spells-spell__action");
     const spells_to_hit = $(".ct-spells-spell .ct-spells-spell__tohit .integrated-dice__container, .ddbc-spells-spell .ddbc-spells-spell__tohit .integrated-dice__container");
     const spells_damage = $(".ct-spells-spell .ct-spells-spell__damage .integrated-dice__container, .ddc-spells-spell .ddc-spells-spell__damage .integrated-dice__container");
-    const initiative = $(".ct-combat__summary-group--initiative .integrated-dice__container, .ct-combat__summary-group--initiative span[class*='styles_numberDisplay'], .ct-combat-tablet__extra--initiative .integrated-dice__container, .ct-combat-tablet__extra--initiative span[class*='styles_numberDisplay'], .ct-combat-mobile__extras > section[class*='styles_boxMobile'] span[class*='styles_numberDisplay']");
-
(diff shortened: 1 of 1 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -2312,8 +2312,8 @@ function deactivateQuickRolls() {
     const spells = $(".ct-spells-spell .ct-spells-spell__action,.ddbc-spells-spell .ddbc-spells-spell__action");
     const spells_to_hit = $(".ct-spells-spell .ct-spells-spell__tohit .integrated-dice__container, .ddbc-spells-spell .ddbc-spells-spell__tohit .integrated-dice__container");
     const spells_damage = $(".ct-spells-spell .ct-spells-spell__damage .integrated-dice__container, .ddc-spells-spell .ddc-spells-spell__damage .integrated-dice__container");
-    const initiative = $(".ct-combat__summary-group--initiative .integrated-dice__container, .ct-combat__summary-group--initiative span[class*='styles_numberDisplay'], .ct-combat-tablet__extra--initiative .integrated-dice__container, .ct-combat-tablet__extra--initiative span[class*='styles_numberDisplay'], .ct-combat-mobile__extras > section[class*='styles_boxMobile'] span[class*='styles_numberDisplay']");
-
+    const initiative = $(".ct-combat__summary-group--initiative .integrated-dice__container, .ct-combat-tablet__extra--initiative span[class*='styles_numberDisplay'], .ct-combat-mobile__extras > section[class*='styles_boxMobile'] span[class*='styles_numberDisplay']");
```
kakaroto, line 2315: ah, good catch, thanks!
I think the integrated dice container also contained that `span[class*='styles_numberDisplay']`, that's why it showed up twice. One possibility would be to do `.eq(0)` at the end of the request so it just picks the first one, or remove the duplicates like you did. However, you removed the span, and kept integrated-dice__container, so you fixed the double rolls when digital dice are on, and it breaks when digital dice are disabled.

kakaroto:
I've just pushed a commit that fixes it. When trying to apply your fix, I realized that if we only keep the `span` and remove teh `integrated-dice__container` from the selector, it will cause the click to trigger DDB's digital dice (even if beyond20 has the option disabled, and likely would roll digital dice twice with the option enabled). So we do need to hook it into the digital dice container to prevent the default click handler.
That also explained why I had the issue of digital dice rolling when I clicked in the mobile layout, so I fixed that as well, and switched to using the `.eq(0)` solution instead, which is simpler (https://api.jquery.com/eq/).
Thanks!

dmportella:
Awesome glad it work out in the end

PR #1161, opened 2024-08-16
added support again to avatar panel

dmportella:
character management pane also changed name and the display in vtt for the avatar was missing.

Added character panel support again that was missing, the display in vtt button

had to add a different way to identify the pane since it is a pane that does not have a header (no title) to select on like all other panes. But it should be oke.

it works on all break points

image


kakaroto (approved):
Looks good, thanks! 👍

PR #1166, opened 2024-09-19
Roll 20 Discord Activity

dmportella:
I merged all the changes together at the moment, there is some testing still to do.

The button is added successfully, the settings for adding discord permission works also.

I noticed at the moment that the rolls are displayed twice on roll20 activity for some reason which I need to figure out why.

@kakaroto been a few weeks since i last looked at this, (i had a bunch of things getting in the way) I had recreate my branch the code space i had open died or i deleted by mistake (probably i deleted XD) 

Need to test the permission for someone that already has the extension and when they receive the update it still works and lets them ask for update their permission.

Also need to figure out why rolls are being sent twice. -  I think it is todo with the port stuff and the normal tab handler (might not need the port stuff in the end)

This will not work on the application only in the browser.

image

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -32,6 +32,7 @@ const SRC_FILES = {
         "src/extension/background.js"
     ],
     options: [
+        "src/common/constants.js",
         "src/common/utils.js",
         "libs/FileSaver.min.js",
         "src/common/settings.js",
--- a/manifest.json
+++ b/manifest.json
@@ -186,6 +186,22 @@
 				"libs/jquery-3.4.1.min.js",
 				"dist/roll20.js"
 			]
+		},
+		{
+			"matches": [
+				"*://1199271093882589195.discordsays.com/*"
+			],
+			"css": [
+				"libs/css/alertify.css",
+				"libs/css/alertify-themes/default.css",
+				"libs/css/alertify-themes/beyond20.css"
+			],
+			"js": [
+				"libs/alertify.min.js",
+				"libs/jquery-3.4.1.min.js",
+				"dist/roll20.js"
+			],
+			"all_frames": true
 		}
 	]
 }
--- a/src/common/constants.js
+++ b/src/common/constants.js
@@ -1,30 +1,35 @@
-ROLL20_URL = "*://app.roll20.net/editor/*";
-FVTT_URL = "*://*/*game";
-DNDBEYOND_CHARACTER_URL = "*://*.dndbeyond.com/*characters/*";
-DNDBEYOND_MONSTER_URL = "*://*.dndbeyond.com/monsters/*";
-DNDBEYOND_ENCOUNTERS_URL = "*://*.dndbeyond.com/my-encounters";
-DNDBEYOND_ENCOUNTER_URL = "*://*.dndbeyond.com/encounters/*";
-DNDBEYOND_COMBAT_URL = "*://*.dndbeyond.com/combat-tracker/*";
-DNDBEYOND_SPELL_URL = "*://*.dndbeyond.com/spells/*";
-DNDBEYOND_VEHICLE_URL = "*://*.dndbeyond.com/vehicles/*";
-DNDBEYOND_SOURCES_URL = "*://*.dndbeyond.com/sources/*";
-DNDBEYOND_CLASSES_URL = "*://*.dndbeyond.com/classes/*";
-DNDBEYOND_EQUIPMENT_URL = "*://*.dndbeyond.com/equipment/*";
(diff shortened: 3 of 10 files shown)


dmportella:
@kakaroto I think turning off the content port stuff removes the issue with the double messages being sent

This code commented out all works oke. Needs further looking in to it, i wont commit this yet until i test a bit more.

image

Custom Sites code works normally

image

dmportella:
Yep I think both Roll20 Activity (browser) and Roll20 Website work fine with out the content port stuff.

dmportella:
if you can confirm I will make another commit with the content port stuff removed.

dmportella:
tested that and it is all working with out the content port.

I dont think we will get it working on the electrum app in discord.

dmportella:
tested it with a clean update and things seem to work ok with a small issue, just one thing that after updating the permissions (which pop up when you load discord activity) you need give a refresh for everything to pick up properly.

kakaroto:
oh nice, I was unsure the port stuff was needed since it's supposed to already be handled transparently by the browser, and I wanted to double check that it was indeed needed. Glad to see it wasn't 😁 
Looks like the only change from what I did last time with the permissions, was adding the host to the manifest? I'm not sure if that was necessary since I think the background script manually injects the scripts, but that `all_frames: true` there might have been the clue to getting it to work.
I haven't tested this yet, but it's cool if it's that easy to get it working. I'll try to tackle it this weekend and release it alongside the hotfix.
Thanks for tackling this and testing it!

dmportella:
> oh nice, I was unsure the port stuff was needed since it's supposed to already be handled transparently by the browser, and I wanted to double check that it was indeed needed. Glad to see it wasn't 😁 Looks like the only change from what I did last time with the permissions, was adding the host to the manifest? I'm not sure if that was necessary since I think the background script manually injects the scripts, but that `all_frames: true` there might have been the clue to getting it to work. I haven't tested this yet, but it's cool if it's that easy to get it working. I'll try to tackle it this weekend and release it alongside the hotfix. Thanks for tackling this and testing it!

Yeah there wasnt a lot needed, most of the code you made was enough.

kakaroto:
> The host in the manifest is probably the only thing you need there was a couple other changes that harmless that might else well add

Ok, lots of testing (to the point where my F5 key is angry with me) and I can see that:
- without the section in the manifest, it simply doesn't work
- that section in the manifest also seems to be adding that site as a new permission, so it might be defeating all that optional permissions stuff.
- if I don't give the permissions through the settings tab, then it doesn't work, unless I click on the Beyond20 icon in the address bar before I launch the activity (as expected, since that gives it permission).
- if I remove the `discord.com` from the `ROLL20_DISCORD_ACTIVITY_DOMAINS` constant, then it stops working (I assume because I can't give it permission to the page, but it still fails to work even if the permission is already explicitly given then I remove the address from constants and reload the extension)
- If discord.com is not authorized, then the error is one of a disconnected port. See screenshot below 

image


You can get that error by launching the roll20 activity without having given explicit permission to discord.com then clicking the beyond20 icon next to the chat and clicking "More Options" in the quick settings popup dialog. That usually sends a message to the background script, and that fails due to the closed port. I assume this is why @rileydutton opened a separate connection to the background script.
If we give permission to discord.com at the same time, then it works without that extra port connection, so I think we're good.

Basically, it sounds like for some reason, the connection between the content script and the background script gets interrupted/closed immediately after it injects the scripts into the page (or possibly it never gets opened) **unless** we also injected a content script in the the parent page of the iframe. If we don't want to have discord.com counting as a 'custom vtt domain', then we would need to add the port request, otherwise, it's not needed.

Now my main concern is how I'd be able to remove that section from the manifest and inject the right scripts into that iframe. I might need a small content-script that just looks for the correct iframe and injects the roll20 script into it if it appears into the discord.com page.

dmportella:
That makes sense, I haven't spent a lot of time with Google extensions before but I could tell things were perfect. But sounds like the option you have outlined might work. Definitely a good shot.

kakaroto:
So, I managed to get it to work even without discord.com in the SUPPORTED_VTT list, and without opening a new connection port to the background script, and the issue with the trying to open the settings popup was because the popup.html page that loads tries to send to its parent tab the request to open the settings, which failed if we only load into the iframe.

Anyways, I did confirm that having the url in the manifest.json caused the extension to get disabled as soon as the update lands, so I removed it from there and instead used an option permission of `webNavigation` which I use to check for the correct iframe loading and to inject the roll20 scripts into that specific iframe only.
It seems to work quite well like this.

Issue #1163, opened 2024-09-08
"Sorcerous Burst" rolls all damage types at the same time


nschoenwald:
It would be great if you could choose the damage type like you can for Chromatic Orb.

dmportella:
Redoing the change now :3, created the PR to early T,  T

kakaroto:
Fixed via ##1168

PR #1167, opened 2024-09-20
WIP added support for Sorcerous Burst dmg type selection

dmportella:
same as chromatic orb

Fixes https://github.com/kakaroto/Beyond20/issues/1163


dmportella:
this will just handle the dmg type, I will look at the Burst aspect of the spell soon.

PR #1168, opened 2024-09-20
added support for Sorcerous Burst Dmg type selection like Chromatic Orb

dmportella:
Does not include Burst aspect yet

Related to -> https://github.com/kakaroto/Beyond20/issues/1163

image

There is an issue with dnd beyond that they are not setting the DMG types properly for high level characters.

image

So this PR will need to wait until it is done, dnd beyond needs to sort out their end or we have to write some code to circumvent this 

Created a forum post for this -> https://www.dndbeyond.com/forums/d-d-beyond-general/bugs-support/206716-new-spell-sorcerous-burst-dmg-types-are-incorrect

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -579,7 +579,6 @@ class Beyond20RollRenderer {
                 }
             }
         
-
             await this._roller.resolveRolls(request.name, all_rolls, request)
             
             // Moved after the new resolveRolls so it can access the roll results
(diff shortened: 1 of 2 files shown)


dmportella:
added support to the burst ability

There is an issue that **there is not way to figure out which ability mod the a specific spell is using** 

So i added the Query that asks the User which Class Spell Modifier they want to use.

I would like to add a Config that the User can default which class to use, not done this yet so its not included here

Example DMG type selection
image

Example DMG in Roll20
image

Example of in case of a Burst which Spell Ability Modifier you are using for the Spell
image

Example DMG in Roll20 for roll that includes Burst's
image

Let me know what you think @kakaroto

src/common/roll_renderer.js
```diff
@@ -610,6 +609,52 @@ class Beyond20RollRenderer {
...
+                        // Prompt for the ability modifier if needed (only if there are multiple options)
+                        if (mods.length > 1 && selectedModifier === null) {
+                            const result = await this.queryGeneric("Spellcasting Ability Modifier", "Burst Ability Modifier", mods);
+                            selectedModifier = (result !== null) ? mods[result] : mods[0]; // Use first one if not selected.
+                        } else { 
+                            selectedModifier = mods[0];
+                        }
+
+                        const modifier = parseInt(request.character.spell_modifiers[selectedModifier]);
+                        burstRollsLeft = (burstRollsLeft != 0) ? burstRollsLeft : modifier; // Set burstRollsLeft to the modifier value if not already set
```
kakaroto, line 633: This can lead to a bug if you have for example modifier=2, and your burst (delta) is 2, it brings `burstRollsLeft` back to 0, then the next iteration of the loop, it sets it back to 2.

This can be simplified if you always use the modifier from sorcerer, so you can initialize the variable to the correct modifier before entering the loop, and you never have to reset it.

src/common/roll_renderer.js
```diff
@@ -610,6 +609,52 @@ class Beyond20RollRenderer {
...
+                        
+                        // Calculate how many rolls are allowed for the burst (delta)
+                        const delta = Math.min(burstCount, burstRollsLeft); // Only allow as many burst rolls as available in burstRollsLeft
+                        
+                        // Reduce burstRollsLeft by delta, ensuring it doesn't go below zero
+                        burstRollsLeft -= delta;
+
+                        // Perform the burst rolls if delta > 0
+                        if (delta > 0 && burstRollsLeft >= 0) {
```
kakaroto, line 642: This should only check for `delta`, otherwise if burstRollsLeft decreases to 0 after you substract delta from it on the previous line, it will make you skip this block by mistake.

kakaroto (changes requested):
Thanks for implementing this. It looks like there's a couple of potential corner cases that could lead to a bug, so I made some comments on that.

src/common/roll_renderer.js
```diff
@@ -610,6 +609,52 @@ class Beyond20RollRenderer {
...
+                for (let [i, dmg_roll] of damage_rolls.entries()) {
+                    const [dmg_type, roll, flags] = dmg_roll;
+                    const faces = parseInt(roll.dice[0].faces);
+                    const burstCount = roll.dice[0].rolls.filter(x => x.roll === faces).length; // Count how many max rolls occurred
+
+                    if (burstCount > 0) {
+                        burstRollCount++;
+                        // Prompt for the ability modifier if needed (only if there are multiple options)
+                        if (mods.length > 1 && selectedModifier === null) {
+                            const result = await this.queryGeneric("Spellcasting Ability Modifier", "Burst Ability Modifier", mods);
```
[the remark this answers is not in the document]
dmportella: it is possible sadly, with pact of the tome and other features I have been able to add the spell to any class combination that does not need sorcerer

src/common/roll_renderer.js
```diff
@@ -610,6 +609,52 @@ class Beyond20RollRenderer {
...
                 }
+            } else if (request.name.includes("Sorcerous Burst")) {
+                const mods = Object.keys(request.character.spell_modifiers);
+                let burstRollsLeft = 0;
+                let burstRollCount = 0;
+                let selectedModifier = null;
+                for (let [i, dmg_roll] of damage_rolls.entries()) {
+                    const [dmg_type, roll, flags] = dmg_roll;
+                    const faces = parseInt(roll.dice[0].faces);
+                    const burstCount = roll.dice[0].rolls.filter(x => x.roll === faces).length; // Count how many max rolls occurred
```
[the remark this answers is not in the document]
dmportella: yep that makes sense I will make the change

dmportella:
I will make some changes and push as soon as I have the time

dmportella:
@kakaroto made a bunch of changes I think this version is way better, it will use sorcerer if available if not will present the class selector. it deals with the burst count properly now i think, I am doing tests at the moment. Let me know what you think

src/common/roll_renderer.js
```diff
@@ -610,6 +609,54 @@ class Beyond20RollRenderer {
...
+                const faces = parseInt(damage_rolls[0][1].dice[0].faces);
+                const initialBurstCount = damage_rolls[0][1].dice[0].rolls.filter(x => x.roll === faces).length;
+                
+                if (initialBurstCount > 0) {
+                    // Determine the selected modifier, prompt only if multiple options and no sorcerer
+                    if (!hasSorcerer && Object.keys(mods).length > 1) {
+                        const result = await this.queryGeneric("Spellcasting Ability Modifier", "Burst Ability Modifier", Object.keys(mods));
+                        selectedModifier = result !== null ? Object.keys(mods)[result] : Object.keys(mods)[0];
+                    } else {
+                        selectedModifier = Object.keys(mods)[0];
```
kakaroto, line 628: should be 
```js
} else if (hasSorcerer) {
   selectedModifier = Object.keys(mods).find(x => x.toLowerCase() === "sorcerer");
} else {
  // ...
```

kakaroto (approved):
Looks good, thanks! I just found one small issue that I've fixed locally. I'll be testing this now then will merge.

dmportella:
Yeah to change those index zero to the proper way of fetching. Well spotted, my vision was already blurred XD

PR #1169, opened 2024-09-23
added support to the HOMEBREW_MAX crit rules

dmportella:
Error when using Perfect Rolls
image

@kakaroto Hi just did some test today and noticed that the refactor was erroring with Perfect Roll Crit rule, So I patched that by moving the code below to the beginning of the main `if` statement. I also noticed that Bursts are not getting included on the crit section, you double the roll if it is a crit which its fine all works out in the end, It might confuse people XD. Perfect Rolls need to be added to the Crit since they are not included in the doubled calculation should we add to the burst result or to crit?

```
const critRule = parseInt(character.getGlobalSetting("critical-homebrew", CriticalRules.PHB));
const doubleForCrit = is_critical && [CriticalRules.PHB, CriticalRules.HOMEBREW_REROLL, CriticalRules.HOMEBREW_MOD].includes(critRule);

```
Standard Crit Rule the Burst is not included on the Crit is that intended?
image

Homebrew MAX rolls
image

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -649,10 +649,12 @@ class Beyond20RollRenderer {
                 const sorcererMod = Object.keys(mods).find(x => x.toLowerCase() === "sorcerer");
                 const mainDamage = damage_rolls.find((roll) => roll[2] === DAMAGE_FLAGS.REGULAR);
                 const critDamage = damage_rolls.find((roll) => roll[2] === (DAMAGE_FLAGS.REGULAR | DAMAGE_FLAGS.CRITICAL));
+                const critRule = parseInt(character.getGlobalSetting("critical-homebrew", CriticalRules.PHB));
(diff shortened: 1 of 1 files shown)


kakaroto:
Thanks for testing and catching it! Looks like I might need a hotfix again soon (although I think it only affects a very specific set of users, since it's just for burst and with that homebrew rule... so likely not urgent).
I'll review this soon (I hope).

dmportella:
> Thanks for testing and catching it! Looks like I might need a hotfix again soon (although I think it only affects a very specific set of users, since it's just for burst and with that homebrew rule... so likely not urgent). I'll review this soon (I hope).

yep it is very niche (it affect my games XD we all use perfect rolls) 

but yes its not a big problem.

kakaroto (approved):
It took me a while to wrap my head around what you were doing, but once I did, it made sense and looks good to me! 👍 
Thanks!

dmportella:
Awesome glad it made sense in the end

Issue #1164, opened 2024-09-18
Class Features seem to be broken


dmportella:
**Describe the bug**
Class Features seem to be broken, they changed classes for them so the selector is not working (style component class names again).

![image](https://github.com/user-attachments/assets/d86a5aeb-b65d-40f3-bebc-cf68aa83b601)

genie warlock (dao) but character option is missing.

**Expected behavior**
Should see Genie Wrath being applied.

**Screenshots**
Broken Code now (selector line)

![image](https://github.com/user-attachments/assets/1cc13e59-7177-416b-807a-ffcc7cc3eada)

HTML now is styled componenterised >:(

![image](https://github.com/user-attachments/assets/b2febfa6-0c25-4a6d-9460-c3e08ada49b7)

**Browser Info (please complete the following information):**
 - Browser chrome
 - Version 128.0.6613.138 

**Additional context**
Adding fix for it shortly

DMMeleedy:
Is this why my barbarian character no longer has a Rage button to slide on in the beyond20 interface?

kakaroto:
> Is this why my barbarian character no longer has a Rage button to slide on in the beyond20 interface?

yes indeed!

DMMeleedy:
Do you know when your fix will make it to the official chrome extension?

dmportella:
> Do you know when your fix will make it to the official chrome extension?

not sure, change has been accepted and merge so soon as it is published

Kodokbotak:
yeah also blood hunter's crimson rite toggle has disappeared

dmportella:
> yeah also blood hunter's crimson rite toggle has disappeared

that should be fixed with the PR that was merged yesterday, i will test it

dmportella:
> yeah also blood hunter's crimson rite toggle has disappeared

with the fix the toggle comes back <3

Wait for a couple days for it to be pulished

![image](https://github.com/user-attachments/assets/01c4f038-d827-4eea-80d8-de23828cd44f)

Kodokbotak:
nooice!

DMMeleedy:
This is not showing up for me in beyond20 either. Is this a feature that will be fixed by this bug fix, or was it not supported to begin with?

Savage Attackerfree-rules, pg. 201
From
Soldier
Once per turn when you hit a target with a weapon, you can roll the weapon’s damage dice twice and use either roll against the target.

dmportella:
[...]

PR #1171, opened 2024-09-30
added support to 2024 Great Weapon Master and Sharpshooter and small bug fixes

dmportella:
added support for 2024 versions of feats, added settings, and fixed features options and choices as they not getting caught.

Adds more fixes for https://github.com/kakaroto/Beyond20/issues/1164

Great Weapon master 2024
image

New setttings
image

Sharpshooter 2024
image

Sharpshooter no longer gives extra damage
```
General Feat (Prerequisite: Level 4+, Dexterity 13+)

You gain the following benefits.

Ability Score Increase. Increase your Dexterity score by 1, to a maximum of 20.

Bypass Cover. Your ranged attacks with weapons ignore [Half Cover](https://www.dndbeyond.com/sources/basic-rules/combat#Cover) and [Three-Quarters Cover](https://www.dndbeyond.com/sources/basic-rules/combat#Cover).

Firing in Melee. Being within 5 feet of an enemy doesn’t impose Disadvantage on your attack rolls with Ranged weapons.

Long Shots. Attacking at long range doesn’t impose Disadvantage on your attack rolls with Ranged weapons.
```

sharpshooter 2014
image

great weapon master 2014
image

For funs BOTH versions of the feat together!!!
image

Add choices support in addition to options which i think they changed it abd broken the features removing s bunch of things from the features list so now they all show where available

image

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -755,7 +755,7 @@ const character_settings = {
         "title": "Feat: Great Weapon Master 2024 (Apply to next roll only)",
         "description": "Heavy Weapon Mastery. Apply extra damage equals your Proficiency Bonus.",
         "type": "bool",
-        "default": false
+        "default": true
     },
     "sharpshooter": {
         "title": "Feat: Sharpshooter (Apply to next roll only)",
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -751,6 +751,12 @@ const character_settings = {
         "type": "bool",
         "default": false
     },
+    "great-weapon-master-2024": {
+        "title": "Feat: Great Weapon Master 2024 (Apply to next roll only)",
+        "description": "Heavy Weapon Mastery. Apply extra damage equals your Proficiency Bonus.",
+        "type": "bool",
+        "default": false
+    },
     "sharpshooter": {
         "title": "Feat: Sharpshooter (Apply to next roll only)",
         "description": "Apply Sharpshooter -5 penalty to roll and +10 to damage",
--- a/src/dndbeyond/base/character.js
+++ b/src/dndbeyond/base/character.js
@@ -225,8 +225,9 @@ class Character extends CharacterBase {
         const features = $(selector).find(".ct-feature-snippet > .ct-feature-snippet__heading, .ct-feature-snippet--class > div[class*='styles_heading'], .ct-feature-snippet--racial-trait > div[class*='styles_heading'], .ct-feature-snippet--feat > div[class*='styles_heading']")
         const feature_list = [];
         for (let feat of features.toArray()) {
-            const feat_reference = $(feat).parent().find("span[class*='styles_metaItem'] > p[class*='styles_reference'] > span[class*='styles_name']");
-            const feat_name = this.getFeatureVersionName(feat.childNodes[0].textContent.trim(), feat_reference.length > 0 ? feat_reference[0].textContent.trim() : undefined);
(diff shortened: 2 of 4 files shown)


src/dndbeyond/base/character.js
```diff
@@ -225,19 +225,39 @@ class Character extends CharacterBase {
         const features = $(selector).find(".ct-feature-snippet > .ct-feature-snippet__heading, .ct-feature-snippet--class > div[class*='styles_heading'], .ct-feature-snippet--racial-trait > div[class*='styles_heading'], .ct-feature-snippet--feat > div[class*='styles_heading']")
         const feature_list = [];
         for (let feat of features.toArray()) {
-            const feat_name = feat.childNodes[0].textContent.trim();
+            const feat_reference = $(feat).parent().find("span[class*='styles_metaItem'] > p[class*='styles_reference'] > span[class*='styles_name']");
+            const feat_name = this.getFeatureVersionName(feat.childNodes[0].textContent.trim(), feat_reference.length > 0 ? feat_reference[0].textContent.trim() : undefined);
```
kakaroto, line 229: For clarity/maintainability, I'd rewrite it like this:
```js

            const feat_reference = $(feat).parent().find("span[class*='styles_metaItem'] > p[class*='styles_reference'] > span[class*='styles_name']").eq(0).text();
            const feat_base_name = feat.childNodes[0].textContent.trim()
            const feat_name = this.getFeatureVersionName(feat_base_name, feat_reference);
```

src/common/settings.js
```diff
@@ -751,6 +751,12 @@ const character_settings = {
         "type": "bool",
         "default": false
     },
+    "great-weapon-master-2024": {
+        "title": "Feat: Great Weapon Master 2024 (Apply to next roll only)",
+        "description": "Heavy Weapon Mastery. Apply extra damage equals your Proficiency Bonus.",
+        "type": "bool",
+        "default": false
+    },
```
kakaroto, line 759: Reading the feat description, I don't think the setting is necessary anymore, since there's no "make a choice before an attack". It seems to always apply, and there's no limit to it, so it should just always add the GWM proficiency bonus on heavy attacks.
https://www.dndbeyond.com/feats/21-great-weapon-master
vs
https://www.dndbeyond.com/feats/1789149-great-weapon-master

The only possibility is if you attack on someone else's turn (opportunity attack), so maybe have a setting but enabled by default? Although the breakdown of the damages is usually meant for people to cherry-pick which damages actually apply depending on the situation, so I wouldn't say it's mandatory either.

kakaroto (changes requested):
Looks good, though we'd need some changes to the behavior due to the way they changed it, means it should just always apply, without a need to enable a toggle at every attack.

src/common/settings.js
dmportella: yep can make those changes np

src/dndbeyond/base/character.js
dmportella: sure thing

dmportella:
I have added the changes, made the setting true/on by default and changed the code to reflect that. also changed the suggested edit on the variables.

kakaroto:
Looks good, thanks! I'll merge then edit the setting description to remove that "apply to next roll only".
Thanks!

dmportella:
> Looks good, thanks! I'll merge then edit the setting description to remove that "apply to next roll only". Thanks!

awesome, yeah i forgot the remove that thank you 🙏🏼

PR #1180, opened 2024-10-14
fixed Great Weapon Fighting, Great Weapon Master and Polearm Master …

dmportella:
…to work as per 2024 rules

fixes: https://github.com/kakaroto/Beyond20/issues/1186

added support gwf for fighting style
added support gwf for feat
added support for gwf in additional damages
added support for gwf and gwm for polearm master Pole Strike
fixed ability and saving throw pane selection since there were changes to the panes by dndbeyond

Great weapon Fighting as per 2024 rules works like elemental adept where each die can only role a minimum of 3
which is different from 2014 which was reroll 1s and 2s.

i had to mark the difference between them and also review other cases like Great Weapon master and Polearm Master which work differently in 2024

now all cases, GWF applies to all damages including additional damages like the frostbrand extra cold damage

Polearm Master changed the action name to "Pole Strike" from the old Polearm Master ones, so patched that and made it work with GWF and GWM

everything has been tested with all crit rules, common, magical weapons and cantrip damages like booming blade etc.

Mind you GWF as per old ways does not apply to booming blade but if it applies to the additional damages i think should apply to it too, but i have not made that change. everything is per existing functionality (besides the core rules changes on the styles)

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -2632,13 +2632,14 @@ function documentModified(mutations, observer) {
             // the page is modified, giving us enoguh time to catch "Initiative" string and add the class name
             // to the div, before the text gets translated
             const initRegex = /\bInitiative\s\(([-+]?\d+)\)/i;
-            if (sidebar.parent().find(".ddbc-ability-score-manager").length > 0) {
+            if (sidebar.parent().find("svg[class*='ddbc-ability-icon']").length > 0 && sidebar.find("div[class*='styles_interactive']").length === 0) { // not saving throw
                 const paneClass = SPECIAL_PANES.ability;
                 markPane(sidebar, paneClass);
                 handlePane(paneClass);
-            } else if (sidebar.find("svg[class*='ddbc-ability-icon']").length > 0) {
+            } else if (sidebar.parent().find("svg[class*='ddbc-ability-icon']").length > 0 && sidebar.find("div[class*='styles_interactive']").length > 0) {
                 // Saving throws have no specific class, but the preview icon has the ddbc-ability-icon class
                 // so we can use that to detect saving throws (excluding ability scores, already handled above)
+                // Also checking for saving throws text bit exist since dndbeyond added that.
                 const paneClass = SPECIAL_PANES.savingThrow;
                 markPane(sidebar, paneClass);
                 handlePane(paneClass);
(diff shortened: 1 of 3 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -1131,11 +1155,20 @@ async function rollAction(paneClass, force_to_hit_only = false, force_damages_on
...
         // Polearm master bonus attack using the other end of the polearm is considered a melee attack.
-        if (action_name.includes("Polearm Master") && (character.hasClassFeature("Great Weapon Fighting", true) || character.hasFeat("Great Weapon Fighting", true))) {
-            damages[0] = damages[0].replace(/[0-9]*d[0-9]+/g, "$&ro<=2");
+        if (action_name.includes("Polearm Master") && 
+            ((character.hasClassFeature("Great Weapon Fighting", true) || character.hasFeat("Great Weapon Fighting", true)) && 
+            (!character.hasClassFeature("Great Weapon Fighting 2024", true) || !character.hasFeat("Great Weapon Fighting 2024", true)))) {
+            damages[0] = damages[0].replace(/[0-9]*d[0-9]+/g, "$&min3");
+        }
+
+        if (action_name.includes("Pole Strike") && (character.hasFeat("Polearm Master 2024", true)) &&
```
kakaroto, line 1164: Can you have 2014 Polearm Master alongside 2024 Great Weapon Fighting?
If yes, then I see a bug, since you check for 2014 PM with 2014 GWF, and 2024 PM with 2024 GWF.

I'd suggest:
```js
const isPolearm = action_name.includes("Polearm Master") || (action_name.includes("Pole Strike") && character.hasFeat("Polearm Master 2024", true));

if (isPolearm && 
    (character.hasClassFeature("Great Weapon Fighting") || 
      character.hasFeat("Great Weapon Fighting"))) {
            damages[0] = damages[0].replace(/[0-9]*d[0-9]+/g, "$&ro<=2");
} else if (isPolearm &&
    (character.hasClassFeature("Great Weapon Fighting 2024") || 
     character.hasFeat("Great Weapon Fighting 2024"))) {
            damages[0] = damages[0].replace(/([0-9]*)d([0-9]+)([^\s+-]*)(.*)/g, (match, amount, faces, roll_mods, mods) => {
                return new Array(parseInt(amount) || 1).fill(`1d${faces}${roll_mods}min3`).join(" + ") + mods;
            });
}
```

kakaroto (changes requested):
Thanks for this PR! It was a lot of work and a lot of fixes!
I noted down a few possible issues I noticed, but I'm not 100% sure on all of them if the behavior is correct, so let me know if you agree/disagree on anything, then I should be able to clean it up and merge it before release.

dmportella:
@kakaroto there i think all is working now (besides the brutal strike) I tested every permutation of the feats/features mixed up everything is working on my end

kakaroto:
Merged the changes and added a code cleanup commit on  top: https://github.com/kakaroto/Beyond20/commit/2c397845ed15fd1a6cd0b23fe5fae439581ab5c4

PR #1182, opened 2024-10-26
added fix for Monster actions for 2024

dmportella:
fixed to hit regex to support 2014 and 2024 monsters 
added saving throw actions support for 2024 monsters

tested with most monsters from 2014 and counter parts in 2024

2014: https://www.dndbeyond.com/monsters/17020-sprite
2024: https://www.dndbeyond.com/monsters/4775845-sprite

2014 BEFORE change
image

2024 BEFORE Change
image

2014 creature after change
image

2024 creature after change
image

This pull requests changes the To hit regex to take into account the changes in 2024 creatures where they have `Melee Attack Roll` Instead of `Melee Weapon Attack` and also the to Hit now doesnt have `to hit` on the text anymore. Some smaller changes were made to take into account other things. But this works now for both versions.

Change the save parsing by a bit I had to split the regex into 2014 and 2024 versions and extract it into a method to make it cleaner. They were very different from one another. (could be improved where we dont need to have strength,charisma etc written out.

There not a lot of 2024 monsters so there is ton of abilities that I can not test because they have not been released. Expect further breaks in this area when all monsters come out.

until then this should fix all the 2024 monsters.

### DIFF
--- a/src/dndbeyond/base/monster.js
+++ b/src/dndbeyond/base/monster.js
@@ -458,11 +458,24 @@ class Monster extends CharacterBase {
     }
 
     parseAttackInfo(description) {
-        const m = description.match(/(Melee|Ranged)(?: Weapon| Spell)? Attack:.*?(\+[0-9]+) to hit.*?, (?:reach |ranged? |Gibletish )?(.*?)(?:,.*?)?\./i)
+        // changed for 2024 info descriptions
+        const m = description.match(/(Melee|Ranged|Spell)(?: Weapon| Spell| Attack) ?(?: Attack|Roll):.*?(\+[0-9]+)(?: to hit.*?|\s?,) (?:reach |ranged? |Gibletish )?(.*?)(?:,.*?)?\./i);
         if (m)
             return m.slice(1, 4);
         else
             return null;
+    }
+
+    parseSaveInfo(description) {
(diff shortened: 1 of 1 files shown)


src/dndbeyond/base/monster.js
```diff
@@ -458,13 +458,26 @@ class Monster extends CharacterBase {
     }
 
     parseAttackInfo(description) {
-        const m = description.match(/(Melee|Ranged)(?: Weapon| Spell)? Attack:.*?(\+[0-9]+) to hit.*?, (?:reach |ranged? |Gibletish )?(.*?)(?:,.*?)?\./i)
+        // changed for 2024 info descriptions
+        const m = description.match(/(Melee|Ranged|Spell)(?: Weapon| Spell| Attack) ?(?: Attack|Roll):.*?(\+[0-9]+)(?: to hit.*?|\s?,) (?:reach |ranged? |Gibletish )?(.*?)(?:,.*?)?\./i);
```
kakaroto, line 462: Wondering if instead of `/(Melee|Ranged|Spell)(?: Weapon| Spell| Attack) ?(?: Attack|Roll):` we used `/(Melee|Ranged|Spell)(?: Weapon Attack| Spell Attack| Attack Roll) ?:`, if it'd be better.
Extremely minor and not an actual issue.

kakaroto (approved):
Looks good to me, thanks!

PR #1187, opened 2024-11-11
Fix for Actor Creation Issue `bug` 

zambo:
Fixed "Create Actors" functionality by updating the Actor object to use the ownership property instead of permission, aligning with FoundryVTT v12 API changes and Actor Schema.

### DIFF
--- a/FVTT-module/beyond20/beyond20.js
+++ b/FVTT-module/beyond20/beyond20.js
@@ -36,8 +36,8 @@ class Beyond20 {
                     user: game.user.id
                 }
             },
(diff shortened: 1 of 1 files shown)


kakaroto:
Ah, thank you! I need to review the v12 changes for this. Were there any other issues you had with v12 and beyond20 that I should fix at the same time? 
Since Beyond20 needs to stay compatible with a reasonable number of Foundry versions, I'll need to adapt your change so it uses the right field whether the user is on v9, v10, v11 or v12.
Thanks!

PR #1188, opened 2024-11-14
added fix for monster side panel

dmportella:
monster side panel css was removed

ct-monster-panel was removed, so i made the same change we have done before when a panel is changed.

There is now a b20-monster-panel class to let all the code find and get the pane

**Before change**
image

**After the changes side panel is now working.**
image

**Tested with 2024 and 2014 monsters too to ensure nothing was affected (does not seem like it has)**
image
image

### DIFF
--- a/src/dndbeyond/base/monster.js
+++ b/src/dndbeyond/base/monster.js
@@ -829,15 +829,15 @@ class Monster extends CharacterBase {
         let hp = null;
         let max_hp = null;
         let temp_hp = null;
-        const groups = $(".ct-creature-pane .ct-collapsible__content .ct-creature-pane__adjuster-group,.ct-creature-pane .ddbc-collapsible__content .ct-creature-pane__adjuster-group");
+        const groups = $(".b20-creature-pane .ct-collapsible__content .b20-creature-pane__adjuster-group,.b20-creature-pane .ddbc-collapsible__content .b20-creature-pane__adjuster-group");
         for (let item of groups.toArray()) {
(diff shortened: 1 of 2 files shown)


src/dndbeyond/base/monster.js
```diff
@@ -829,15 +829,15 @@ class Monster extends CharacterBase {
         let hp = null;
         let max_hp = null;
         let temp_hp = null;
-        const groups = $(".ct-creature-pane .ct-collapsible__content .ct-creature-pane__adjuster-group,.ct-creature-pane .ddbc-collapsible__content .ct-creature-pane__adjuster-group");
+        const groups = $(".b20-creature-pane .ct-collapsible__content .b20-creature-pane__adjuster-group,.b20-creature-pane .ddbc-collapsible__content .b20-creature-pane__adjuster-group");
```
kakaroto, line 832: The change to `.b20-creature-pane` makes sense, since you add that class, but the others wouldn't work.
There's no `.b20-creature-pane__adjuster-group` but I do see a `styles_adjusterGroup_*` which we should be using.
{29C459FC-D902-49F1-AAD8-E749D15B0083}

kakaroto (commented):
looks good for the panel detection but saw a bug in how the HP values are getting read.
I'll fix this later myself before merging, so no need to update the PR at this time.

src/dndbeyond/base/monster.js
dmportella: Yeah i did not mean to change that one it was a find and replace mistake, i should be more careful. 
yeah you are right tho

PR #1191, opened 2024-11-21
Added support to identify what version of the rules the character is 2014 or 2024

dmportella:
Every 2024 class has a Core Class Traits, the 2014 do not

For the purpose of rules, any character multiclassing 2024 classes with 2014 class will be treated as 2024 characters

This change will add a version field (number) to the character object.

This is enabling PR for other work.

Dndbeyond 2024 classes were not made to work with 2014 classes but dnd beyond lets you do it anyway.
image


kakaroto (approved):
LGTM!

PR #1192, opened 2024-11-21
added small fix to remove GWM 2024 from Polearms feat attacks since t…

dmportella:
added small fix to remove GWM 2024 from Polearms feat attacks since they only work on 2014 GWM

Also the bonus actions dont have a properties section so we can see if they are heavy or not 

the text has been split now so it checks for heavy or if it is a pole arms attack from the feat (bonus action)
image
image


kakaroto (approved):
LGTM

PR #1198, opened 2024-11-24
added support to display the mastery for the weapon if present 

dmportella:
added support to display the mastery for the weapon if present and mastered

# Mastery

> Dndbeyond adds the star to weapon the character has mastery on, so we get the mastery text from the item properties and use that to display on the roll that weapon attack can use mastery

image

> Example of a dagger attack that has mastery

image

> example of an attack that does not have mastery

image


kakaroto (commented):
Ouh that's nice, I hadn't heard of this Mastery thing.
Not sure though why we'd need to put the Mastery property into the request. The roll renderer can already do `request["Properties"]["Mastery"]` without needing to do `request["mastery"]` since that's just duplicating the data (and if you add it as a new field for the request, then the API docs would need to be modified).
Also wondering if there's any way to display that mastery in roll20 🤔

dmportella:
The weapon has mastery but the mastery is the fact that the user has mastered the weapon

dmportella:
soa weapon has a mastery property that you can only take advantage if you mastered it

# example of a mastered weapon
> it shows that you mastered by the fact you have the star on it
image
> you also get the additional section on the weapon showing the mastery you have
image

may be the mastery name could be changed to `is-mastered` ?

# example where you DONT have mastery
image
image

dmportella:
Happy to update the docs, I havent done that yet, but let see what you think about this change first

also no idea if there is a way to display in roll20 or foundry, 

Each mastery give an effect, we could have a On Hit Effects list that is displayed that can be extended in the future to add more things . . .

Here are the specifics of mastery
>https://www.dndbeyond.com/sources/dnd/phb-2024/equipment#MasteryProperties
image

dmportella:
The other problem @kakaroto is that if we use the properties mastery we only get that the weapon has a mastery of NIck for example, and not Dagger (Nick) homebrew items or magical weapons dont have a type we can extract.

kakaroto:
Updated the docs here: https://github.com/kakaroto/Beyond20/commit/346f0e440493d8441ae9d41c16823a234b4dd1cb

PR #1194, opened 2024-11-22
added support for GWF 2014 and 2024 for brutal critical

dmportella:
known issue with versatile weapon is applying versatile brutal dmg to 1 handed dmg too will address this in a later commit

Pole arm attack examples with gwf 2024
Screenshot 2024-11-22 172202
Screenshot 2024-11-22 172227

polearm master 2014 with gwf 2014
image
image

pole strike 2024 + gwf 2014
image
image

long sword (versatile)
image
image

One handed should be using 1d8 for the brutal strike dmg and it is not (known issue)
image

GWF 2024 great axe
image

GWF 2024 long sword versatile
image

pole arm 2014 GWF 2024
image

pole strike 2024 GWF 2024
image

PR #1193, opened 2024-11-22
add support for brutal strike 2024 feature for barbarian

dmportella:
brutal strike replaces brutal critical
add 1d10 or 2d10 to the damage that matches the weapon dmg type

image

image

image

Great Axe attack with GWF
Screenshot 2024-11-22 162234
Screenshot 2024-11-22 162259

Unarmed strikes with out GWF
Screenshot 2024-11-22 164741
Screenshot 2024-11-22 164748

2014 GWF 
image

New setting that is auto turned off once used
Screenshot 2024-11-22 164832

Brutal Strike Applies to ONE attack per turn that you can choose.

Improved the Polearm Master and Pole Strike and consolidated it so it also applies to them if the players wants too use brutal strike.

PR #1201, opened 2024-11-27
Two brutals fix

dmportella:
Merged both PR's so that we can use the better function on brutal critical

https://github.com/kakaroto/Beyond20/pull/1193
https://github.com/kakaroto/Beyond20/pull/1194

I will close the other PR's

This commit should have resolved all review comments


src/dndbeyond/content-scripts/character.js
```diff
@@ -811,28 +811,13 @@ async function rollItem(force_display = false, force_to_hit_only = false, force_
...
-                    if (versatile_damage != "") {
-                        versatile_damage = versatile_damage.replace(/([0-9]*)d([0-9]+)([^\s+-]*)(.*)/g, (match, amount, faces, roll_mods, mods) => {
-                            return new Array(parseInt(amount) || 1).fill(`1d${faces}${roll_mods}min3`).join(" + ") + mods;
-                        });
-                    } else {
-                        damage = damage.replace(/([0-9]*)d([0-9]+)([^\s+-]*)(.*)/g, (match, amount, faces, roll_mods, mods) => {
-                            return new Array(parseInt(amount) || 1).fill(`1d${faces}${roll_mods}min3`).join(" + ") + mods;
-                        });
-                    }
+
```
kakaroto, line 814: There was a check here for `if (damages.length == 0)` which disappeared in the refactor.
I think it was because this code runs in a loop, so if the weapon has multiple damages (flaming sword?) then it could get applied multiple times, so `if (damages.length == 0)` makes sure it only executes it once.

src/dndbeyond/content-scripts/character.js
```diff
@@ -1174,27 +1160,22 @@ async function rollAction(paneClass, force_to_hit_only = false, force_damages_on
...
+            let brutal_strike_dmg = applyGWFIfRequired(action_name, properties, `${strikeDieCount}d10`);
+
+            damages.push(brutal_strike_dmg);
+            damage_types.push("Brutal Strike");
+
+            settings_to_change["brutal-strike"] = false;
         }
 
+        // apply for normal die
+        damages[0] = applyGWFIfRequired(action_name, properties, damages[0]);
```
kakaroto, line 1177: Slight difference here in the code, old code was checking for `action_name.includes("Polearm Master")` while the new code (in utils) checks for `action_name == "Polearm Master - Bonus Attack"`
Also noticed there's a check here for `&& character.hasFeat("Polearm Master 2024", false)` which disappeared from the new utils function. Are we introducing a regression here possibly for some weird use case?

kakaroto (commented):
It looks so much cleaner and easier to review this way, I love it 😁 Two small comments that I'd like your input on before merging. Thanks

src/dndbeyond/content-scripts/character.js
dmportella: good point yeah that should be the case if there is multiple damage die. things like flamesword the fire doesnt get gwf

src/dndbeyond/content-scripts/character.js
dmportella: good point those should be back (the hasfeat check) the action name we should be more lenient and check if it includes polearm master

dmportella:
@kakaroto i think the changes are good now

src/dndbeyond/base/utils.js
```diff
@@ -305,8 +305,8 @@ function applyGWFIfRequired(action_name, properties, damage) {
         (
             (properties["Properties"].includes("Versatile") && character.getSetting("versatile-choice") != "one") || 
             properties["Properties"].includes("Two-Handed"))) ||
-            action_name == "Polearm Master - Bonus Attack" ||
-            action_name == "Pole Strike")
+            action_name.includes("Polearm Master") && character.hasFeat("Polearm Master", false)  ||
+            action_name.includes("Pole Strike") && character.hasFeat("Polearm Master 2024", false))
```
kakaroto, line 309: due to logical operator precedence and mixing of `&&` and `||`, I think this check needs parenthesis for more clarity/maintainability. I'll add them as I push

kakaroto (approved):
LGTM, thx!

PR #1202, opened 2024-11-27
fix for ability mod not being found

dmportella:
impacted dice injection and number of other things - rather obscure problem it was not showing on all instances

This fix is important there was friend of mine Tomo that raised an issue that his deflect missiles were not working (it was not including the dex mod and it was rolling the incorrect die)

image
image
image

# I could not reproduce it on 2024

image
image
image

# Using a 2014 character

But when using his character it was seeing the issue, this was a 2014 issue.

image

# Stylecomponents hits again

## They changed the little box where character used this format of abilities 
image

## But all works on this format
image

dndbeyond added styledcomponents to only the mod of the ability which was not getting populated

Making the mod part of the all abilities in i the character miss the mod.

This made the very niche example of "you dexterity modifier ..." fail on dice injection.

This is all working now

### DIFF
--- a/src/dndbeyond/base/character.js
+++ b/src/dndbeyond/base/character.js
@@ -123,7 +123,7 @@ class Character extends CharacterBase {
             let modifier = $(ability).find(".ct-ability-summary__primary .ct-signed-number,.ddbc-ability-summary__primary .ddbc-signed-number,.ddbc-ability-summary__primary span[class*='styles_numberDisplay']").text();
             let value = $(ability).find(".ct-ability-summary__secondary,.ddbc-ability-summary__secondary").text();
             if (modifier == "") {
-                modifier = $(ability).find(".ct-ability-summary__secondary .ct-signed-number,.ddbc-ability-summary__secondary span[class*='styles_numberDisplay'], .ct-ability-summary__secondary span[class*='styles_numberDisplay']").text();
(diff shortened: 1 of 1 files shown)


src/dndbeyond/base/character.js
```diff
@@ -123,7 +123,7 @@ class Character extends CharacterBase {
             let modifier = $(ability).find(".ct-ability-summary__primary .ct-signed-number,.ddbc-ability-summary__primary .ddbc-signed-number,.ddbc-ability-summary__primary span[class*='styles_numberDisplay']").text();
             let value = $(ability).find(".ct-ability-summary__secondary,.ddbc-ability-summary__secondary").text();
             if (modifier == "") {
-                modifier = $(ability).find(".ct-ability-summary__secondary .ct-signed-number,.ddbc-ability-summary__secondary .ddbc-signed-number, .ct-ability-summary__secondary span[class*='styles_numberDisplay']").text();
+                modifier = $(ability).find(".ct-ability-summary__secondary .ct-signed-number,.ddbc-ability-summary__secondary span[class*='styles_numberDisplay'], .ct-ability-summary__secondary span[class*='styles_numberDisplay']").text();
```
kakaroto, line 126: Humm.. because of how complex this can be, I'd prefer to keep the old selector in too, so I'd replace it with this instead:
```js
modifier = $(ability).find(".ct-ability-summary__secondary .ct-signed-number,.ddbc-ability-summary__secondary .ddbc-signed-number,.ddbc-ability-summary__secondary span[class*='styles_numberDisplay'], .ct-ability-summary__secondary span[class*='styles_numberDisplay']").text();
```
(adding `ddbc-ability-summary__secondary span[class*='styles_numberDisplay']` rather than modifying the existing selector)

dmportella: sure thing

kakaroto (approved):

PR #1203, opened 2024-11-28
fixed small typo and add valid choices to cunning strike

dmportella:
Just small fixes there was a typo that made the page break and I added a valid choices so only selected and valid choice is shown

### DIFF
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -261,6 +261,7 @@ async function buildAttackRoll(character, attack_source, name, description, prop
             if (choice === null) return null; // Query was cancelled;
         }
 
+        // TODO: refactor into a method and remove it from build attack roll 
         if (character.hasClass("Rogue") && character.hasClassFeature("Sneak Attack 2024") && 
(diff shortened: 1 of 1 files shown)


src/dndbeyond/base/utils.js
```diff
@@ -273,12 +274,14 @@ async function buildAttackRoll(character, attack_source, name, description, prop
...
                 const choices = await queryCunningStrike();
+                const validChoices = [];
                 for (const choice of choices) {
                     if (choice.action === "None") continue;
                     if (choice["die"] > sneakDieCount) continue;
                     sneakDieCount -= choice["die"];
+                    validChoices.push(choice);
                 }
-                roll_properties["cunning-strike-effects"] = choices.filter(f => f.action !== "None").map(m => m.action).join(", ") || undefined;
+                roll_properties["cunning-strike-effects"] = validChoices.filter(f => f.action !== "None").map(m => m.action).join(", ") || undefined;
```
kakaroto, line 284: doesn't need the `filter` anymore 😁

kakaroto (approved):
Thanks, good catches!

PR #1205, opened 2024-11-30
WIP: Spell Effects

dmportella:
Adding support for popular spell effects

Currently only doing: **Bless, Enlarge and Reduce**

# New sections called Effects
image

# Bless
image

NOT DONE BLESS ATTACK BONUS YET

# Enlarge
image
image

# Reduce
image
image

### DIFF
--- a/src/common/roll_renderer.js
+++ b/src/common/roll_renderer.js
@@ -503,7 +503,12 @@ class Beyond20RollRenderer {
         const {rolls} = this.getToHit(request, "", data, type)
         await this._roller.resolveRolls(title, rolls, request);
         this.processToHitAdvantage(request.advantage, rolls);
-        return this.postDescription(request, title, null, {}, null, rolls);
+
+        const roll_info = [];
+        if (request["effects"] !== undefined && request["effects"].length !== 0)
+            roll_info.push(["Effects", request["effects"].join(', ')]);
+
+        return this.postDescription(request, title, null, {}, null, rolls, roll_info);
     }
 
     async rollSkill(request, custom_roll_dice = "") {
@@ -786,6 +791,8 @@ class Beyond20RollRenderer {
         }
 
         const roll_info = [];
+        if (request["effects"] !== undefined)
+            roll_info.push(["Effects", request["effects"]]);
         if (request["mastery"] !== undefined)
             roll_info.push(["Mastery", request["mastery"]]);
         if (request["save-dc"] != undefined)
@@ -861,6 +868,8 @@ class Beyond20RollRenderer {
                 }
             }
         }
+        if (request["effects"] !== undefined)
+            roll_info.push(["Effects", request["effects"]]);
         if (request["mastery"] !== undefined)
             roll_info.push(["Mastery", request["mastery"]]);
         if (request["save-dc"] !== undefined)
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -355,6 +355,19 @@ async function buildAttackRoll(character, attack_source, name, description, prop
                     crit_damage_types.push(isBrutal && isSavage ? "Savage Attacks & Brutal" : (isBrutal ? "Brutal" : "Savage Attacks"));
                 }
             }
+
+            // effects
+            if(properties["Attack Type"] == "Melee" || properties["Attack Type"] == "Unarmed Strike") {
+                if(character.getSetting("effects-enlarge", false)) {
+                    damages.push("+1d4");
+                    damage_types.push("Enlarge");
+                    addEffect(roll_properties, "Enlarge");
+                } else if(character.getSetting("effects-reduce", false)) {
+                    damages.push("-1d4");
+                    damage_types.push("Reduce");
+                    addEffect(roll_properties, "Reduce");
+                }
+            }
         }
 
         roll_properties["critical-damages"] = crit_damages;
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -13,6 +13,12 @@ function sendRollWithCharacter(rollType, fallback, args) {
     return sendRoll(character, rollType, fallback, args);
 }
 
+function addEffect(rollProperties, effect) {
+    rollProperties["effects"] = rollProperties["effects"] || [];
+    if (!rollProperties["effects"].includes(effect)) {
+        rollProperties["effects"].push(effect);
+    }
+}
 
 async function rollSkillCheck(paneClass) {
     const skill_name = $("." + paneClass + "__header-name").text();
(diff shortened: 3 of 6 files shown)


dmportella:
@kakaroto added all the effects related stuff for bless, and enlarge and reduce, future i will add bane, hunters mark and hex. Hunters mark will allow us to apply the Foe Slayer class feature that changes hunters mark damage to 1d10. and i will probably stop there, and not add any other effect.

src/common/roll_renderer.js
```diff
@@ -503,7 +503,12 @@ class Beyond20RollRenderer {
         const {rolls} = this.getToHit(request, "", data, type)
         await this._roller.resolveRolls(title, rolls, request);
         this.processToHitAdvantage(request.advantage, rolls);
-        return this.postDescription(request, title, null, {}, null, rolls);
+
+        const roll_info = [];
+        if (request["effects"] !== undefined && request["effects"].length !== 0)
```
kakaroto, line 508: I'm thinking of a third party sending a roll and putting `request.effects = null` or even `request.effect = "Bless"` which would cause a crash here.

A safer check would be:
```js
if (Array.isArray(request["effects"]) && request["effects"].length > 0)
```

src/dndbeyond/content-scripts/character.js
```diff
@@ -277,9 +294,84 @@ function rollAbilityOrSavingThrow(paneClass, rollType) {
...
             roll_properties.d20 = "1d20min10";
 
+    // effects 
+    if(rollType == "saving-throw" && character.getSetting("effects-bless", false)) {
+        roll_properties["modifier"] += "+1d4";
+        addEffect(roll_properties, "Bless");
+    } else if(rollType == "saving-throw" && character.getSetting("effects-bane", false)) {
+        roll_properties["modifier"] += "-1d4";
+        addEffect(roll_properties, "Bane");
+    }
```
kakaroto, line 304: I think doing the bless/bane in one place is better. Here you're doing it for ability and saving throws, but then the check happens again in `sendRoll` but only checks for attack and spell-attack.
It's better to remove it from here and expand the list of roll types the check in `sendRoll` applies it to.

src/dndbeyond/content-scripts/character.js
```diff
@@ -277,9 +294,84 @@ function rollAbilityOrSavingThrow(paneClass, rollType) {
...
+        [RollType.SUPER_DISADVANTAGE]: RollType.DISADVANTAGE
+    };
+
+    if (roll_properties["advantage"] === undefined) {
+        roll_properties["advantage"] = RollType.ADVANTAGE;
+    } else if (roll_properties["advantage"] in advantageMapping) {
+        roll_properties["advantage"] = advantageMapping[roll_properties["advantage"]];
+    }
+    
+    if(key_modifiers.normal_roll) {
```
kakaroto, line 332: Unfortunately, you can't override the key modifiers here, because if the hotkey is `N` for normal and `A` for advantage, and you switch it, then when the user releases the letter `N`, the `key_modifiers.normal_roll` will get set to `false` (which it already was) and `key_modifiers.advantage` will not be modified, since the letter `A` was never pressed by the user... and you end up with a "stuck hotkey".

kakaroto (changes requested):
This is great! Thank you for doing it. I think people will love the feature and I like how you added it to the popup!
Only big concern with this PR is the `key_modifiers` issue, and I'm not sure how to resolve that.... 🤔 humm.. do you even need to set a hotkey to `true`? You're already setting the advantage value, so I think if you just set the values to `false` then it won't be an issue?
Too tired to think clearly, but hopefully I'm on the right track. Let me know what you think!

dmportella:
@kakaroto thank you for the review and all the comments, i all the changes suggested and also removed all that modifier stuff out, after testing further and confirming what you said, i moved it to sendroll, which made it a lot simpler. tested with key modifiers and all is working (i no longer change the modifiers) and only changing the advantage variable and all works as intended.

src/dndbeyond/content-scripts/character.js
```diff
@@ -183,6 +190,15 @@ async function rollSkillCheck(paneClass) {
...
     }
 
+    // effects
+    if(roll_properties.ability == "STR" && character.getSetting("effects-enlarge", false)) {
+        adjustRollAndKeyModifiersWithAdvantage(roll_properties);
+        addEffect(roll_properties, "Enlarge");
+    } else if(roll_properties.ability == "STR" && character.getSetting("effects-reduce", false)) {
+        adjustRollAndKeyModifiersWithDisadvantage(roll_properties);
+        addEffect(roll_properties, "Reduce");
+    }
```
kakaroto, line 200: Forgot to remove the enlarge/reduce effect on the skill checks now that you moved it to `sendRoll`

src/dndbeyond/base/utils.js
```diff
@@ -465,6 +477,22 @@ async function sendRoll(character, rollType, fallback, args) {
...
+            addEffect(req, "Bane");
+        }
+
+        if(req.character.settings["effects-bless"] && req.type === "saving-throw") {
+            req["modifier"] += "+1d4";
+            addEffect(req, "Bless");
+        } else if(req.character.settings["effects-bane"] && req.type === "saving-throw") {
+            req["modifier"] += "-1d4";
+            addEffect(req, "Bane");
+        }
```
kakaroto, line 495: The modifier is usually the +x modifier of the ability, so I'd say the bless/bane effect should still be added as a `custom-roll-dice` so you can bundle the check for saving-throw in the previous `if` statement.
```ks
if (req.character.settings["effects-bless"] && 
   (req.type === "saving-throw" || (["attack", "spell-attack"].includes(req.type) && req["to-hit"])) {
```
Actually, I think in the case of attack/spell-attack the custom-roll-dice isn't used if there is no to-hit, so you shouldn't need to check for it, so it could be simplified into
```js
if (req.character.settings["effects-bless"] && ["attack", "spell-attack", "saving-throw"].includes(req.type)) {
```

src/dndbeyond/base/utils.js
```diff
@@ -485,6 +513,14 @@ async function sendRoll(character, rollType, fallback, args) {
         }
     }
 
+    // effects 
+    if((rollType == "saving-throw" || rollType == "ability" || rollType == "skill") && req.ability == "STR" && character.getSetting("effects-enlarge", false)) {
+        adjustRollAndKeyModifiersWithAdvantage(req);
+        addEffect(req, "Enlarge");
+    } else if((rollType == "saving-throw" || rollType == "ability") && req.ability == "STR" && character.getSetting("effects-reduce", false)) {
```
kakaroto, line 520: Missing the check for "skill" on the reduce, you only added it on enlarge.
Also, could benefit from using `["saving-throw", "ability", "skill"].includes(rollType)` for cleaner code readability

kakaroto (changes requested):
Found a few small issues, though nothing major

kakaroto (approved):
Thanks for the quick fixes!

src/extension/popup.js
```diff
@@ -333,7 +356,8 @@ function populateCharacter(response) {
             // When loading settings, the discord target combobox gets replaced in order to be filled,
             // so we need to fetch it again to add the advanced-option class to it
             $("#beyond20-option-discord-target").addClass("advanced-option");
-        }
+        
+        options.find(".effects-option").hide();}
```
kakaroto, line 360: Late night coding and I just noticed this in code, the line seems to have been written before the closing `}` instead of after... so it's inside of the `if` scope above it.
Fixed it, pushing soon!

PR #1206, opened 2024-12-01
added support for Halfling Luck 2024 feature

raystuart:
## Description

The 2024 Halfling racial trait is called "Luck", while the 2014 racial trait is called "Lucky".

This PR adds logic to check for "Luck" racial trait in order to display the settings form and to apply the roll modification to `args.d20`.


kakaroto (approved):
LGTM! Thanks!

PR #1208, opened 2024-12-09
Paladin: Fix for 'Radiant Strikes'

Lott:
The Paladin feature "Improved Divine Smite" was renamed to "Radiant Strikes" in the 2024 PHB.  This PR updates the popup and damage script to check for either name.

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -608,7 +608,7 @@ const character_settings = {
         "default": true
     },
     "paladin-improved-divine-smite": {
-        "title": "Paladin: Improved Divine Smite",
+        "title": "Paladin: Radiant Strikes (Improved Divine Smite)",
         "description": "Roll an extra 1d8 radiant damage whenever you hit with a melee weapon",
         "type": "bool",
         "default": true
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -409,8 +409,8 @@ function handleSpecialMeleeAttacks(damages=[], damage_types=[], properties, sett
     }
 
     if (character.hasClass("Paladin")) {
-        //Paladin: Improved Divine Smite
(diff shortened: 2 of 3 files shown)


dmportella:
You can't change the existing 2014 feature to 2024

There 1000s of people that still play 2014 and do not own 2024 content at all

This will need a new config.

dmportella:
you did add both, sorry oke the or statement, The name should contain both names instead of just changing to 2024 version.

dmportella:
this will likely need a new setting

2014 version only supports weapon strikes as the 2024 needs to support unarmed strikes too.

Lott:
> this will likely need a new setting
> 
> 2014 version only supports weapon strikes as the 2024 needs to support unarmed strikes too.

Forgot about the unarmed strikes.  Let me see if I can manage it with one, otherwise I'll break it into two settings.

dmportella:
> > this will likely need a new setting
> > 2014 version only supports weapon strikes as the 2024 needs to support unarmed strikes too.
> 
> Forgot about the unarmed strikes. Let me see if I can manage it with one, otherwise I'll break it into two settings.

please break it into two settings

I have not looked at the area of code yet for this features but the new one will likely need to be done on another part of the code to affect unarmed strikes

dmportella:
Thank you tho, I had not yet reviewed the paladin changes, done a bunch of the others which are just waiting to be released.

kakaroto:
Hi, 
Thank you for the contribution! I've rewritten it as two separate settings and committed it here: https://github.com/kakaroto/Beyond20/commit/91ef3600eaaf528a56cc50e1594cf9771eb2c4c4 
I thought I could keep it on a single setting, but it got a little too ugly for me, so it was best to keep them separate.
I've also realized that Improved Divine Smite was acting already on unarmed strikes, so I fixed that at the same time. 
I think that was the last thing needed for the release, thanks!

PR #1217, opened 2025-01-08
fix: reordered selector keeping old one, to fetch properties of actions

dmportella:
Added hotfix for Properties from actions that were not being selected.

This is likely a change in the on how the properties are done has broken the original selector.

All actions were broken but this fix corrects that.

image

image

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -1150,7 +1150,7 @@ async function rollAction(paneClass, force_to_hit_only = false, force_damages_on
         return displayAction(paneClass);
     }
     // b20-action-pane and ct-custom-action-page both use ct-action-detail for the details
-    const properties = propertyListToDict($("." + paneClass + " .ct-action-detail [role=list] > div"));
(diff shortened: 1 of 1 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -1150,7 +1150,7 @@ async function rollAction(paneClass, force_to_hit_only = false, force_damages_on
         return displayAction(paneClass);
     }
     // b20-action-pane and ct-custom-action-page both use ct-action-detail for the details
-    const properties = propertyListToDict($("." + paneClass + " .ct-action-detail [role=list] > div"));
+    const properties = propertyListToDict($("." + paneClass + " [role=list] > div, .ct-action-detail [role=list] > div"));
```
kakaroto, line 1153: Should be 
```js
    const properties = propertyListToDict($(`.${paneClass} [role=list] > div, .${paneClass} .ct-action-detail [role=list] > div`));
```
You missed adding the `"." + paneClass ` part on the old selector.
Fixed on my side, thanks!

Issue #1219, opened 2025-01-10
DnDBeyond Custom Action, Description & Pics wont post to VTT... Thanks for addressing :))


tjagz123:
**Describe the bug**
**DnDBeyond Custom Action wont post to VTT**

**To Reproduce**
1. Go to - **open DnD Beyond Character Sheet**
2. Click on - **Actions, Manage Custom, Click any custom action,** 

**Expected behavior**
**When you you click on Beyond 20 button the description would display on VTT, and also Display to VTT button is missing** 

**Screenshots**
To help explain your problem, add screenshots of where you clicked in D&D Beyond, a screenshot of your Beyond20 character settings if applicable and the result you saw in the VTT.

![image](https://github.com/user-attachments/assets/2f92ca91-1ba1-4a62-a35a-943521f3edd7)

###This how is used to look before today's patch...

![image](https://github.com/user-attachments/assets/58e5fb2c-fbbd-4acd-8974-847f6c626793)

**Browser Info (please complete the following information):**
 - Browser **[chrome and firefox]**
 - Version [**1 / 9 / 25 Version]**
 - If applicable, any dev console errors (F12 to open the console) from D&D Beyond and the VTT

**Additional context**
Add any other context about the problem here, such as a link to the character sheet or D&D Beyond page that shows the problem.

kakaroto:
Which version were you on? I released 2.11.1 yesterday which addressed this, so I think you just didn't yet have your browser updating the extension for you but if you update it, it should work.
Please confirm, thanks!

tjagz123:
Hi there, by the way love Beyond20 app... I'm a Ko-Fi supporter :))
Yes I'm using  2.11.1 in both Firefox & Chrome. Both of these browsers are up to date. 
I checked the Beyond20 Quick setting on the DnD Beyond Character page too... nothing odd there. 
I'm sure it's something simple... Any Thoughts? Thanks in advance!!

tjagz123:
Just rebooted PC, no change. 🤔

dmportella:
@kakaroto yeah not I got the main attacked working, but had not seen the descriptions not showing, i found the issue - styledcomponent change on the description, I will make a patch for it now.

tjagz123:
Hi @kakaroto @dmportella ... Thanks for the followup and communication... :)) I just checked DnD Beyond and the issue still persists. I rebooted my PC and tried Firefox & Chrome. Does the patch take a few days to publish?

[...]

PR #1221, opened 2025-01-10
fix: changed description selector to match new styledcomponent css

dmportella:
Small fix for description of custom actions 

Descriptions have been changed to use styledcomponent which break the css selector.

Fixes: https://github.com/kakaroto/Beyond20/issues/1219


kakaroto (approved):

kakaroto (commented):
oups, I just noticed the string is using `""` not backticks, so the `${paneClass}` won't be replaced

kakaroto:
> oups, I just noticed the string is using `""` not backticks, so the `${paneClass}` won't be replaced

Fixed in https://github.com/kakaroto/Beyond20/commit/da89fdcfdadd231b97493c3a48831fe6d31b64d5

PR #1224, opened 2025-01-15
feature: added support to exhaustion rules from 2024/2014

dmportella:
added setting effect that allows any character of what ever version to use 2024 exhaustion rules on their rolls.

# Added New effect setting

> Setting is NOT dependent on the character being a 2024 class, giving the ability for people to decide to use 2024 rules if they want it. it is turned off by default

image

# Rolls that it effects

> The following rolls are impacted: `attack`, `spell-attack`, `initiative`, `ability`, `saving-throw`, `skill`, `death-save`. The Effect is shown in the effects list displaying the level of exhaustion.

image
image
image

# Exhaustion for 2014
image

## Attacks

> Attacks and Spell Attacks

image

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -508,6 +508,13 @@ const character_settings = {
 
     // effects
 
+    "effects-exhaustion": {
+        "title": "Effect: Exhaustion",
+        "description": "Applies 2024 rules for exhaustion. This condition is cumulative and effects all D20 roles applying a -2 * exhaustion level modifier to the roll.",
+        "type": "bool",
+        "default": false
+    },
+
     "effects-bless": {
         "title": "Effect: Bless",
         "description": "Whenever a target makes an attack roll or a saving throw before the spell ends, the target adds 1d4 to the attack roll or save.",
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -508,9 +508,16 @@ const character_settings = {
 
     // effects
 
-    "effects-exhaustion": {
-        "title": "Effect: Exhaustion",
-        "description": "Applies 2024 rules for exhaustion. This condition is cumulative and effects all D20 roles applying a -2 * exhaustion level modifier to the roll.",
+    "effects-exhaustion-2024": {
+        "title": "Effect: Exhaustion 2024",
+        "description": "Applies 2024 rules for exhaustion. This condition is cumulative and effects all D20 rolls applying a -2 * exhaustion level modifier to the roll.",
+        "type": "bool",
+        "default": false
+    },
+
+    "effects-exhaustion-2014": {
+        "title": "Effect: Exhaustion 2014",
+        "description": "Applies 2014 rules for exhaustion. This condition is cumulative and applies disadvantage on ability checks at exhaustion >= 1 and disadvantage on attacks and saving throws on exhaustion >= 3.",
         "type": "bool",
         "default": false
     },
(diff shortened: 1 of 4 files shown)


src/common/settings.js
```diff
@@ -508,6 +508,13 @@ const character_settings = {
 
     // effects
 
+    "effects-exhaustion": {
+        "title": "Effect: Exhaustion",
+        "description": "Applies 2024 rules for exhaustion. This condition is cumulative and effects all D20 roles applying a -2 * exhaustion level modifier to the roll.",
```
kakaroto, line 513: typo s/roles/rolls/ :)

src/common/settings.js
```diff
@@ -508,6 +508,13 @@ const character_settings = {
 
     // effects
 
+    "effects-exhaustion": {
```
kakaroto, line 511: Should be `effects-exhaustion-2024` as the 2014 rules for exhaustion also affect rolls, so I think we ought to have both 2014 and 2024 exhaustion effects supported.
Unless... we check if the character sheet is 2024 or 2014 and apply the exhaustion rules depending on the character sheet version, in which case the description needs to be version agnostic.

src/common/settings.js
```diff
@@ -508,6 +508,13 @@ const character_settings = {
...
+        "title": "Effect: Exhaustion",
+        "description": "Applies 2024 rules for exhaustion. This condition is cumulative and effects all D20 roles applying a -2 * exhaustion level modifier to the roll.",
+        "type": "bool",
+        "default": false
```
kakaroto, line 515: I think this is the kind of effect that deserves having a default of `true`, since it won't have any effect if the character has no exhaustion levels.

src/dndbeyond/base/utils.js
```diff
@@ -485,6 +485,13 @@ async function sendRoll(character, rollType, fallback, args) {
             req.character.settings["custom-roll-dice"] = (req.character.settings["custom-roll-dice"] || "") + " -1d4";
             addEffect(req, "Bane");
         }
+
+        if (req.character.exhaustion && req.character.exhaustion != 0 && req.character.settings["effects-exhaustion"] && 
+            (["attack", "spell-attack"].includes(req.type) && req["to-hit"] || 
+            ["initiative", "ability", "saving-throw", "skill", "death-save"].includes(req.type))) {
+            req.character.settings["custom-roll-dice"] = (req.character.settings["custom-roll-dice"] || "") + ` -${req.character.exhaustion * 2}`;
+            addEffect(req, `Exhaustion (${req.character.exhaustion})`);
```
kakaroto, line 493: Add a check here for the character sheet version, if 2014, then force disadvantage on ability checks at `exhaustion >= 1` and disadvantage on attacks and saving throws on `exhaustion >= 3`
This way it would support both legacy and 2014 exhaustion rules (if 2014, name the effect as `Legacy Exhaustion (${req.character.exhaustion)`).

kakaroto (changes requested):
Nice work! I'm suggesting adding support for Legacy exhaustion too.
I can tackle it today as I think I'll push a hotfix update tonight

src/common/settings.js
dmportella: yeah i kept the exhaustion not checking for version because someone might want to use 2024 exhaustion on a 2014 character

src/common/settings.js
dmportella: I would turn it true only if it was version 2024 character

src/dndbeyond/base/utils.js
dmportella: I will

src/common/settings.js
dmportella: sorted

kakaroto (approved):
Looks good, thanks!

PR #1225, opened 2025-01-18
fix: added fixes for tidbits, attributes of monster extras

dmportella:
Added a bunch of fixes to ensure attributes and tidbits worked again for monster extras

HP, HP formula, AC, AC Meta, Saving throws, skills were all broken

# Attributes
image

### DIFF
--- a/src/dndbeyond/base/extras.js
+++ b/src/dndbeyond/base/extras.js
@@ -64,24 +64,33 @@ class MonsterExtras extends CharacterBase {
         const attributes = stat_block.find("div[class*='styles_attribute__']");
         for (let attr of attributes.toArray()) {
             const label = $(attr).find("h2[class*='styles_attributeLabel']").text().trim();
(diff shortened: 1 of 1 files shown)


src/dndbeyond/base/extras.js
```diff
@@ -64,24 +64,33 @@ class MonsterExtras extends CharacterBase {
         const attributes = stat_block.find("div[class*='styles_attribute__']");
         for (let attr of attributes.toArray()) {
             const label = $(attr).find("h2[class*='styles_attributeLabel']").text().trim();
-            const values = $(attr).find("p[class*='styles_attributeValue']");
+            const data = $(attr).find("p");
```
kakaroto, line 67: You can do `$(attr).find("p").eq(0)` so you don't need to do `$(data[0])` afterwards (unless you need a potential array response for the `addAttribute` which uses the entire `data` object)

kakaroto (approved):
LGTM!

PR #1227, opened 2025-01-20
fix: fixing dndbeyond incorrect sorcerous burst damage values

dmportella:
DND Beyond when they set up the new cantrip sorcerous burst they set the damage values incorrectly

# Incorrect values
image

> We use the properties of the spell to show which dmg type the person wants, this fix will use the first damage value which is the correct one and apply it to all damage types.

#
> [!IMPORTANT]
> There is NO check to see if the first number if the highest, this is simple fix to set all values the same. it relies on dndbeyond damage values to be correct.

## Casting the spell
image
image

### DIFF
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -218,7 +218,7 @@ async function buildAttackRoll(character, attack_source, name, description, prop
             const choice = await queryDamageTypeFromArray(roll_properties.name, damages, damage_types, ["Acid", "Cold", "Fire", "Lightning", "Poison", "Thunder"]);
             if (choice === null) return null; // Query was cancelled;
         } else if (roll_properties.name === "Sorcerous Burst") {
-            // fix dndbeyond dmg issue
+            // HACK ALERT: temporarily use acid damage....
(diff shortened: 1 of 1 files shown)


src/dndbeyond/base/utils.js
```diff
@@ -218,6 +218,9 @@ async function buildAttackRoll(character, attack_source, name, description, prop
             const choice = await queryDamageTypeFromArray(roll_properties.name, damages, damage_types, ["Acid", "Cold", "Fire", "Lightning", "Poison", "Thunder"]);
             if (choice === null) return null; // Query was cancelled;
         } else if (roll_properties.name === "Sorcerous Burst") {
+            // fix dndbeyond dmg issue
+            damages = damages.map(m => damages[0]); // set everything to the correct acid dmg
```
kakaroto, line 222: Hummm.. are we sure this happens very very early on? Before any other damages are added, like sneak attack/rage/etc... (not that they'd apply for a spell, but you know what I mean).

dmportella: not for this spell, other spells like shadow blade sneak does apply

kakaroto (approved):

PR #1229, opened 2025-01-26
Fix support for steel defender formula

kakaroto:
This checks if the formula we parsed ends in "+" then check if the sibling node is a modifier and adds it to the formula (removing the sibling node entirely)

This is how it looks in ddb originally:
image


dmportella:
image
image
Looking good.

dmportella (approved):
yep, right now as far as we know it is only steel defender that has the problem, so the check for integer is the best we can do.

PR #1230, opened 2025-01-29
fix: added fix for hp syncing due to dndbeyond styledcomponent changes

dmportella:
# Fix for HP sync issues

> they recently changed the hp controls so they use styledcomponent ones

image

# work around is to use the sidebar
image

# Styled components now in this controls

image

> added a fix that make the control work again.


kakaroto (approved):
LGTM and I tested it on all 3 configurations (mobile, tablet and desktop), it seems to work perfectly now 👍

Issue #1234, opened 2025-02-08
HP doesn't get updated


SetesMaat:
So when ever i change the HP by healing or damaging of my character, it doesn't update it in foundry.

**To Reproduce**
1. Go to ddb
2. Put any number into the health bar and apply heal or damage
3. go to foundry 
4. see that the HP is still the same as befor

**Expected behavior**
That the HP change to the value the character has

**Screenshots**

![Image](https://github.com/user-attachments/assets/7dc9a6c2-1145-4623-b09e-cc3346f99a79)

![Image](https://github.com/user-attachments/assets/2da7d9b2-d0de-4136-990b-59af3e7d203e)

![Image](https://github.com/user-attachments/assets/e605ae73-2b83-48ee-9882-d3430da711b5)

**Browser Info (please complete the following information):**
 - Opera GX(chromium) but we tried it with other browsers like firefox
 - Version 116.0.5366.103
 -  Chromium-Version:131.0.6778.266

**Additional context**
For some it seems to work for other it doesn't don't now why we even exchanged settings so there is nothing wrong I think

SetesMaat:
https://github.com/user-attachments/assets/2daacf6a-db89-4c05-bb98-cb3e3289c07c

So new Infos apparently it updates when you update the window itself.
Not sure how that works but it does...

SetesMaat:
Again new info when you use the small site view (seen in the video in the end) it always updates like normal

Maybe the name of the value in the big view changed and beyond20 doesn't get the right value?

albertholaso:
I'm noticing this same issue that DnDBeyond character HP is not updating on FoundryVTT character sheet and tokens.

albertholaso:
I found that if the **HP Management panel is open**, hitpoints in FoundryVTT updates properly.
Without that panel open, hitpoints does not update in FoundryVTT.

![Image](https://github.com/user-attachments/assets/8fc5f4f7-f5fe-42e4-bc73-dd3a34aeab74)

dmportella:
Yep that has been sorted already, the PR just needs to be accepted.

Lindharin:
I've updated to 2.13.0, and HPs are not being updated in Roll20 still.  I'm not sure if the fix in 2.13 only applied to Foundry, or if the fix wasn't successful.  I've confirmed with four friends that they are having the same issue even after updating the extension.  We've tried in both Chrome and Firefox.

cgplayer2000:
[...]

PR #1236, opened 2025-02-16
fix: Hp syncing and death saving throws

dmportella:
# HP sync

> Fixes: https://github.com/kakaroto/Beyond20/issues/1234

# HP sync

> hp sync now works both in destop and mobile

image
image

# side panel class

> side panel now works again

image

# Death saving throws

> Fixed death saving throws, added roll dice back

image

### DIFF
--- a/src/dndbeyond/base/character.js
+++ b/src/dndbeyond/base/character.js
@@ -135,18 +135,33 @@ class Character extends CharacterBase {
     }
 
     updateHP() {
-        const health_pane = $(".ct-health-manager");
+        const health_pane = $(".b20-health-manage-pane");
         let hp = null;
         let max_hp = null;
         let temp_hp = null;
         if (health_pane.length > 0) {
-            hp = parseInt(health_pane.find(".ct-health-manager__health-item--cur .ct-health-manager__health-item-value").text());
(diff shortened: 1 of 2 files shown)


src/dndbeyond/base/character.js
```diff
@@ -135,35 +135,62 @@ class Character extends CharacterBase {
...
+                label: "label[class*='styles_label'], span[class*='styles_label']",
+                current: "button[class*='styles_valueButton']",
+                max: "div[class*='styles_number']",
+                temp: ".ct-health-summary__hp-group--temp .ct-health-summary__hp-item--temp .ct-health-summary__hp-item-content, div[class*='styles_temp']",
+                tempValue: ".ct-health-summary__hp-number, button[class*='styles_valueButton'], input[class*='styles_input']"
+            }
+        };
+        
+        let hp = 0, max_hp = 0, temp_hp = 0;
+        if(healthPane || quickAccessPane){
```
kakaroto, line 180: This will always be true, I think you want `if (healthPane.length > 0 || quickAccessPane.length > 0)`

kakaroto (approved):
I like how you set the `selectors` object! Makes for much cleaner code.
Found just a small bug, but otherwise looks good.
I'll fix the bug before I push the release, so no worries, and thanks!

Issue #1241, opened 2025-03-19
Rogue Cunning Strike issues


Adamant11746:
**Describe the bug**
The Cunning Strike toggle only works once before it turns itself off. It also doesn't work at all if you roll Sneak Attack separately.

**To Reproduce**
1. Toggle Sneak Attack and Cunning Strike
2. Roll an attack
3. Cunning Strike works and then turns itself off
4. Toggle Cunning Strike but not Sneak Attack
5. Roll Sneak Attack from the feature (requires it to be set to "Display as Attack" which is on by default)
6. Cunning Strike doesn't trigger, and you get the full number of sneak attack dice.

**Expected behavior**
Cunning Strike doesn't say next roll only, so I expected it to stay toggled and ask each time whether I wanted to Cunning Strike (especially important for rogues with only 1 attack). It should also work when rolling the separate action for Sneak Attack (especially important for multiclassed or dual wielding rogues with more than 1 attack.

**Screenshots**
To help explain your problem, add screenshots of where you clicked in D&D Beyond, a screenshot of your Beyond20 character settings if applicable and the result you saw in the VTT.

**Browser Info (please complete the following information):**
 - Browser: Firefox
 - Version: Whatever the current version is, I don't know how to check but it's fully up to date

Adamant11746:
I see. It looks like it's mostly working as intended. I think it should be possible to toggle cunning strike but not sneak attack though, since otherwise you are forced to roll sneak on every attack just to get the option to drop a die (or dice) for cunning strike. At the moment it seems easier to manually drop the last die for the simpler cunning strike options rather than toggling things on and off every attack, and manually roll a reduced number of dice for the more expensive options. I don't think that the ideal solution for me should be to completely ignore the cunning strike (and even the sneak attack) toggle and just do the math myself...

At this point it seems to be more of a feature request than a bug report, so maybe change the tag?

Also, making a hotkey has the side effect of being inadvertently toggled every time I type something with that key, and there are very few keys I can use for things that don't get used when I type. In fact, it's just the 2 alt keys and right ctrl (and `, but that's my push to talk key). I know it's partially redundant with the hotkey system, but a separate setting (as initially envisioned in [https://github.com/kakaroto/Beyond20/issues/1160](url)) seems like a very useful thing to have when you think about the limited number of available hotkeys and the large number of features across all the classes.

dmportella:
I could do something like this
[...]

PR #1242, opened 2025-03-22
feature: added support for locking checkboxes

dmportella:
> [!WARNING]
> Hi @kakaroto what do you think about this? I made it as separate setting because i didnt want to have to deal with merging of settings from stored character settings already etc, and i didnt want to create another setting type, so i opted to add the "lock" setting to the bool option types as a reference to another setting that is not displayed separately. see this ticket as reference for the feature - https://github.com/kakaroto/Beyond20/issues/1241

> [!IMPORTANT]
> Code is not optimized yet just wanted to get you to give an eyeball at it

# Feature

> Padlock for options, adds a padlock to certain settings that tells beyond 20 to ignore the reverting of the settings after an action was taken, example SNEAK ATTACK

# Sneak Attack

> Some users suggested that they would like to use the sneak attack action and apply cunning strike to it, without having to have the sneak attack option turned on for all attacks.

# Padlock

> For Cunning strike there is not a padlock next to it that will tell the extension NOT to turn off cunning strike when it is used. this will work for normal attacks and sneak attack action.

> [!NOTE]
> Padlock will override the reverting of the setting on all uses including when cunning strike is set by a hotkey 

image
image
image

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -736,8 +736,13 @@ const character_settings = {
         "default": false
     },
     "rogue-sneak-attack": {
-        "title": "Rogue: Sneak Attack",
+        "title": "Rogue: Sneak Attack (Apply to next roll only)",
         "description": "Send Sneak Attack damage with each attack roll",
+        "type": "bool",
+        "default": true,
+        "lock": "rogue-sneak-attack-lock"
+    },
+    "rogue-sneak-attack-lock": {
         "type": "bool",
         "default": true
     },
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -579,6 +579,11 @@ const character_settings = {
         "title": "Bard: College of Whispers: Psychic Blades",
         "description": "Use your Bardic Inspiration to deal extra psychic damage (Apply to next roll only)",
         "type": "bool",
+        "default": false,
+        "lock": "bard-psychic-blades-lock"
+    },
+    "bard-psychic-blades-lock": {
+        "type": "bool",
         "default": false
     },
     "bard-spiritual-focus": {
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -832,13 +832,23 @@ const character_settings = {
         "title": "Feat: Great Weapon Master (Apply to next roll only)",
         "description": "Apply Great Weapon Master -5 penalty to roll and +10 to damage",
         "type": "bool",
+        "default": false,
+        "lock": "great-weapon-master-lock"
+    },
+    "great-weapon-master-lock": {
+        "type": "bool",
         "default": false
     },
     "great-weapon-master-2024": {
-        "title": "Feat: Great Weapon Master 2024",
+        "title": "Feat: Great Weapon Master 2024 (Apply to next roll only)",
         "description": "Heavy Weapon Mastery. Apply extra damage equals your Proficiency Bonus.",
         "type": "bool",
-        "default": true
+        "default": true,
+        "lock": "great-weapon-master-2024-lock"
+    },
+    "great-weapon-master-2024-lock": {
+        "type": "bool",
+        "default": false
     },
     "sharpshooter": {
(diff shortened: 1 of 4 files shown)


src/common/settings.js
```diff
@@ -974,12 +979,17 @@ function createHTMLOptionEx(name, option, short = false, {advanced=false}={}) {
...
     const description_p = description ? description.split("\n").map(desc => E.p({}, desc)) : [];
     const title = short ? option.short : option.title;
+    const lock = option.lock;
     let e = null;
     if (option.type == "bool") {
         e = E.li({ class: "list-group-item beyond20-option beyond20-option-bool" },
             E.label({ class: "list-content", for: name },
                 E.h4({}, title),
                 ...description_p,
+                (_ => lock ? E.div({ class: 'material-switch--padlock pull-right' },
```
kakaroto, line 989: Why are you making it into an arrow function that you're calling? Why can't it just be a regular tertiary operator? 
```js
(lock ? E.div(...) : null),
```
That being said, the `null` appears as text in your screenshots, so I think it might be better to approach it this way:
```js
    const lock = option.lock;
    const lockDiv = lock ? E.div({ class: 'material-switch--padlock pull-right' },
                      E.input({ id: lock, class: "beyond20-option-input", lock, type: "checkbox" }),
                    E.label({ for: lock })
                ): null;
```
and in the call itself, use this to avoid having an extraneous argument to the function
```js
                ...(lock ? [lockDiv] : []),
```

### PoC:
image

src/dndbeyond/base/utils.js
```diff
@@ -266,7 +266,8 @@ async function buildAttackRoll(character, attack_source, name, description, prop
 
         // TODO: refactor into a method and remove it from build attack roll 
         if (character.hasClass("Rogue") && character.hasClassFeature("Sneak Attack 2024") && 
-            character.getSetting("rogue-sneak-attack", false) && 
+            (character.getSetting("rogue-sneak-attack", false) || 
+            name.includes("Sneak Attack") && !character.getSetting("rogue-sneak-attack", false) && character.getSetting("rogue-cunning-strike", false)) &&
```
kakaroto, line 270: Very confusing statement because of the mix of `||` and `&&` without separating them in parenthesis (I know there's a priority order on the AND and OR operators, but without grouping them, it's hard to wrap my head around the order.
You also don't need to check for `!rogue-sneak-attack` since you're already in an `||` from the truthy check, so if it checks for the `name.includes`, we already know it's disabled.

kakaroto (commented):
I'm loving the feature. We've received often the request to make a way to not reset some options after each use, and we've also got the request to clear the sneak attack after each use... I've never thought of a simple way to do it.. I was leaning towards making those into dropdowns instead of toggles, but I love your approach and I think it works really well!

I'd say, add the "(apply to next roll)" text on the cunning strike label, since that should have been there, and add a `lock` option for all of the other options that have "apply to next roll" in them.
Would also love to have the lock option added to the `rogue-sneak-attack` option (but with `default: true` so we don't change the expected behavior.

src/common/settings.js
dmportella: thank u i will apply the changes

dmportella:
Hey @kakaroto ,

Thanks so much for the suggestions! I'm really glad you loved the idea - that means a lot. 🙌

I went ahead and made all the changes you suggested. The label for Cunning Strike now includes the "(apply to next roll)" text, as it should’ve from the start. I've also added the lock option to all other settings that include "apply to next roll," just as you requested.

And yes! The rogue-sneak-attack option now has a lock as well, with the default set to true, so the expected behavior remains unchanged.

Let me know if there's anything else you'd like adjusted!

dmportella:
image

kakaroto (approved):
Looks good, thanks! I'll just change the text to say "apply to next roll only, **unless locked**" for clarity. Will do that now once I merge.
Thanks!

PR #1243, opened 2025-04-05
Per-Character VTT Setting

bknode:
Beyond20 works great for us when using Roll20, but a few DMs are starting to experiment with D&D Beyond's Maps.

In those cases, the extension keeps popping up a dialog with the final roll result when it can't find a VTT. I understand this is useful for Discord users, but when we're just using Maps, it becomes a bit repetitive. We could have our players enable or disable the extension based on whose campaign we're playing that night, but I thought it might be easier to have a character-specific setting that enables or disables the VTT for that particular character.

Please take a look at my changes and let me know what you think.

Thanks for all your hard work on this great extension!

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -481,6 +481,13 @@ const character_settings = {
             "two": "Use weapon Two-handed"
         }
     },
+    "use-vtt": {
+        "short": "Use VTT",
+        "title": "Use VTT",
+        "description": "Send rolls to the default VTT.",
+        "type": "bool",
+        "default": true
+    },
     "custom-roll-dice": {
         "title": "Custom Roll dice formula bonus",
         "description": "Add custom formula to d20 rolls (Bless, Guidance, Bane, Magic Weapon, etc..)",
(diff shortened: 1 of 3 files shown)


dmportella:
Yea I get it, the dnd beyond maps we have not done anything on them yet, I will review and let you know

kakaroto:
Thanks for the PR, though I don't quite like the approach taken, mostly because the setting's name and description is not very clear to new users, and I don't like the setting being checked in the background script. It'd be better to check it on the character page script instead.
There are two other possible paths I can see that might be better. First we have this option: 
image

I think we can change it to also say "hide when using digital dice" (which is required in your case for using the DDB Maps.

The other possibility would be adding a new option to this existing setting:
image

If selecting "D&D Beyond Maps", it would just automatically force the use of digital dice and not send the rolls to the background script.

That being said, I'd prefer we don't go that route because the digital dice are much inferior to the output of Beyond20, as you don't get to see all the actual damages separated (would enable Rage make any difference for example?) or separate the output/calculate totals between one handed and two handed weapon use, etc... So I'd rather not add "DDB Maps" as a feature unless we actually support sending our full output to it.
I'd go instead with an option to hide the "no VTT" warning if digital dice are enabled instead, which sounds much more useful and likely to be a use case that players would use.

PR #1249, opened 2025-04-14
Fix Sacred Weapon

ibaaske:
Devotion Paladin's Sacred Weapon is not populating in the options window.
Screenshot 2025-04-14 at 8 28 59 AM

This is because it's coming through as a Class Feature and as "Channel Divinity: Channel Divinity: Sacred Weapon"
Screenshot 2025-04-13 at 10 48 58 PM

This change listens for that instead.

Testing with the change...it now shows.
Screenshot 2025-04-13 at 10 55 44 PM

### DIFF
--- a/src/extension/popup.js
+++ b/src/extension/popup.js
@@ -283,7 +283,7 @@ function populateCharacter(response) {
             e = createHTMLOption("wildfire-spirit-enhanced-bond", false, character_settings);
             options.append(e);
         }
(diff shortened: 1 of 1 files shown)


dmportella:
There is further problems with this, 2014 has channel divinity BUT 2024 does not - same subclass but different versions of the paladin class name the feature differently

# 2024 class with devotion subclass
image

# 2014 class with devotion subclass
image

You need check for both

I will accept the change and make the remaining changes ty

Issue #1252, opened 2025-04-25
Tavern brawler not included on all monk unarmed attacks


mblair415:
The tavern brawler feat isn't being included on all of the monk's unarmed attacks.

There is a regular monk unarmed attack action with no feat as an option in case someone would like to attack without their feat.  There is also an enhanced unarmed attack that will take the feat into account and reroll an attack roll of 1 with the "ro<1" roll20 functionality (technically, I think "ro" and "ro<1" both do the trick).

The monk also has two bonus action unarmed attacks.  Neither of these are taking the feat into account.  There is one unarmed bonus attack.  There is also a bonus attack flurry of blows.  

It feels like the enhanced version should replace the regular version.  There's never a time that it deactivates unlike something like the savage attacker feat which only activates once per turn.

**To Reproduce**
Build a monk and get the tavern brawler origin feat (sailor background or human species).  Notice that there is a regular action unarmed attack that doesn't include the feat.  There is also an enhanced attack that does include it.  Look at the bonus action attacks.  There are two different ways for a monk to use an unarmed attack as a bonus action and neither of these include the feat's functionality. 

**Expected behavior**
There's no way to turn the feat off.  Technically, every version of the monk's unarmed attacks should include this functionality. 

**Screenshots**
To help explain your problem, add screenshots of where you clicked in D&D Beyond, a screenshot of your Beyond20 character settings if applicable and the result you saw in the VTT.

**Browser Info (please complete the following information):**
 - Chrome
 - Version 135.0.7049.96 (Official Build) (arm64)

dmportella:
Will triage and check the intended behaviour ty for raising it.

dmportella:
thank you made the fixes its should be in the next release <3

PR #1258, opened 2025-05-17
fix: added tavern brawler 2024 to all unarmed strike actions from mon…

dmportella:
I have added tot he tavern brawler 2024 feat

> [!NOTE]
> Fixes: https://github.com/kakaroto/Beyond20/issues/1252 

> Now all unarmed strike have the r<=1 also made a few smaller changes, these attacked were not being recognised as melee so other feats were not being applied to them. The attacks were, "enhanced unarmed strike", "flurry of blows" and "unarmed strike".

> [!IMPORTANT]
> This has highlighted also that the flurry of blows was not taking advantage of other feats that affect melee attacks.

# tavern brawler effect now will show in the attacks
Image

# Unarmed Strike
Image

# Enhanced Unarmed Strike
Image

# Bonus action Unarmed Strike
Image

# Flurry of Blows
Image

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -575,13 +575,15 @@ function handleSpecialGeneralAttacks(damages=[], damage_types=[], properties, se
             const intelligence = character.getAbility("INT") || {mod: 0};
             const mod = parseInt(intelligence.mod) || 0;
             const psychic_action = action_name.toLocaleLowerCase();
-            if(["psionic power: psionic strike"].includes(psychic_action)) {
+            if(["psionic power: psionic strike", "psionic power: protective field"].includes(psychic_action)) {
                 // Use full modifier, even if deeply negative
                 damages[0] = ensureModifier(damages[0], mod);
-            } else if(["psionic power: protective field"].includes(psychic_action)) {
-                // Clamp the modifier to minimum of -1, not below that
-                const cappedMod = mod < -1 ? -1 : mod;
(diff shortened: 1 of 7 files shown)


dmportella:
@kakaroto this one should be done too

src/dndbeyond/content-scripts/character.js
```diff
@@ -1263,11 +1264,27 @@ async function rollAction(paneClass, force_to_hit_only = false, force_damages_on
...
+            "hooves",
+            "talons",
+            "thunder gauntlets",
+            "unarmed fighting",
+            "arms of the astral self",
+            "shadow blade",
+            "predatory strike",
+            "enhanced unarmed strike",
+            "flurry of blows"
+        ].includes(action_name.toLocaleLowerCase());
```
kakaroto, line 1287: Not quite, because you changed the previous `action_name.includes(x)` into `action_name == x`. I'm not sure, but I **think** som eof these actually match partially the action name, so a safer solution would be:
```js
const isMeleeAttack = [ .... ].some(action => action_name.includes(action))
```

kakaroto (changes requested):
small tweak needed!

src/dndbeyond/content-scripts/character.js
dmportella: yep i can fix that

dmportella:
@kakaroto changes done 07

dmportella:
@kakaroto just making sure you see it, it should be done now. I rebased with master to make sure i had all the changes.

kakaroto (approved):

Issue #1251, opened 2025-04-25
Fighter's psi warrior subclass not adding intelligence modifier to abilities


mblair415:
The fighter class has a subclass of psi warrior.  At level 3 the class gets an ability called Psionic Power which grants two abilities that have the same issue.

This is two bugs written as a single ticket.  Sorry.

In both cases they're not adding the intelligence modifier. 

There is a complication with this as the ability states that if the intelligence modifier is lower than -1 then it will cap out with a -1 modifier.  I don't know of any roll20 modifiers that will capture this functionality.

This is a core feature for this subclass granted at level 3, so this would impact every single person who builds this subclass with an intelligence modifier that isn't exactly 0.

**To Reproduce**
Build a 3rd level fighter with 12 or more intelligence.  Take the psi warrior subclass.  Click on the Psionic Strike ability or click on the Protective Field ability.

Both will roll the raw psionic die and not add in any modifier.

**Expected behavior**
When either of these abilities are used it should add in the intelligence modifier to the psionic die result, and if the intelligence modifier is -2 or less it should instead use a modifier of -1.

I don't know if that last part is possible due to roll20 limitations.

**Browser Info (please complete the following information):**
 - Chrome
 - Version 135.0.7049.96 (Official Build) (arm64)

mblair415:
<img width="211" alt="Image" src="https://github.com/user-attachments/assets/f4174ae7-25ba-4c66-9609-95273d7c703e" />
<img width="494" alt="Image" src="https://github.com/user-attachments/assets/c2e4a302-843f-496a-9493-d297ba4604d1" />

Two images attached.  There are two places to trigger the Protective Field ability.  They do not work the same.  One appears to work correctly (didn't check with a -2 or less modifier).  Triggering the ability as a reaction did not add in the intelligence modifier.  Triggering it from the second location did add the modifier.

One image includes the reaction options that show the two places it can be triggered.  One as a reaction by clicking on the crossed swords.  The other if you click on the beyond20 logo below where it says 1D6+2 (which is accurate).  The second image shows the dice rolls.

It looks like the Psionic Strike attack modifier works the same way with two ways to trigger it and one works correctly and the other doesn't work right.

dmportella:
I will check the behaviour and try reproduce, but yes it is possible to add the modifier to the abilities.

dmportella:
[...]

PR #1259, opened 2025-05-17
fix: Psionic skills

dmportella:
> [!NOTE]
> Fixes: https://github.com/kakaroto/Beyond20/issues/1251 

# added mod to the Psionic warrior actions

> Added the int mod protective field is a min of 1 roll

## Psionic Power: Psionic Strike
image

## Psionic Power: Protective Field
image

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -570,14 +570,17 @@ function handleSpecialGeneralAttacks(damages=[], damage_types=[], properties, se
     // Class Specific
     if (character.hasClass("Fighter")) {
         if(character.hasClassFeature("Psionic Power")) {
-            // HACK ALERT: fixes dndbeyond missing mods but incase they are added in the future we ensure the mdo is applied if not present if it is present it is not applied
+            // HACK ALERT: fixes dndbeyond missing mods but incase they are added in the future we ensure the mod is applied if not present if it is present it is not applied
             const intelligence = character.getAbility("INT") || {mod: 0};
             const mod = parseInt(intelligence.mod) || 0;
             const psychic_action = action_name.toLocaleLowerCase();
             if(["psionic power: psionic strike"].includes(psychic_action)) {
+                // Use full modifier, even if deeply negative
                 damages[0] = ensureModifier(damages[0], mod);
             } else if(["psionic power: protective field"].includes(psychic_action)) {
-                damages[0] = ensureModifier(damages[0], mod).replace(/[0-9]*d[0-9]+/g, "$&min1");
+                // Clamp the modifier to minimum of -1, not below that
+                const cappedMod = mod < -1 ? -1 : mod;
+                damages[0] = ensureModifier(damages[0], cappedMod);
             }
         } 
     }
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -422,7 +422,7 @@ function handleSpecialMeleeAttacks(damages=[], damage_types=[], properties, sett
                 damage_types.push("Symbiotic Entity");
         }
     }
-
+    
     if (character.hasClass("Paladin")) {
         // Paladin: Improved Divine Smite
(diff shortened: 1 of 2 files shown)


kakaroto:
> @kakaroto the min1 is not taken into consideration, do i have to do anything else different to make sure that after the roll is cast even with a negative mod it would be a min of 1?

I'm not sure how "min1 is not taken into consideration" since a minimum roll of a dice is already 1 😛 
I don't think you can have a `min1` for the whole formula, I think we only support it as a dice roll modifier. What you could do is build the min value dynamically... if the modifier is `- 3` then you'd do `1d20min4 - 3` basically.

src/dndbeyond/content-scripts/character.js
```diff
@@ -568,6 +568,23 @@ function handleSpecialGeneralAttacks(damages=[], damage_types=[], properties, se
...
+            const intelligence = character.getAbility("INT") || {mod: 0};
+            const mod = parseInt(intelligence.mod) || 0;
+            const psychic_action = action_name.toLocaleLowerCase();
+            if(["psionic power: psionic strike"].includes(psychic_action)) {
+                // Use full modifier, even if deeply negative
+                damages[0] = ensureModifier(damages[0], mod);
+            } else if(["psionic power: protective field"].includes(psychic_action)) {
+                // Clamp the modifier to minimum of -1, not below that
+                const cappedMod = mod < -1 ? -1 : mod;
+                damages[0] = ensureModifier(damages[0], cappedMod);
```
kakaroto, line 583: This is the wrong fix, you shouldn't be changing the modifier, because if it's -3, and you roll a 20, you should have a total of 17, not 19.
Instead, this was my suggestion:
```js
// We set the minimum dice roll to be the minimum value needed so that the minimum total with the mod is 1
const min = mod >= 0 ? 1 : Math.abs(mod) + 1;
damages[0] = ensureModifier(damages[0], mod).replace(/[0-9]*d[0-9]+/g, `$&min${min}`)`;
```

dmportella:
@kakaroto I know what you mean but after reading the feature i realized the the MOD can be a minimum of -1 it can not be more then that so it was not needed to do the min function.

kakaroto:
> `Protective Field. When you or another creature you can see within 30 feet of you takes damage, you can use your reaction to expend one Psionic Energy die, roll the die, and reduce the damage taken by the number rolled plus your Intelligence modifier (minimum reduction of 1), as you create a momentary shield of telekinetic force.`
> 
> So the minimum reduction you can have on damage is zero ( a minus one )

Yeah, I read it too before commenting and it confused me at first in the same way as you, but really, it's not the reduction to the dice based on intelligence modifier which is a minimum of 1 (i.e: "intelligence modifier minimum of -1") but rather the damage is reduced by a minimum of 1.

> "**REDUCE** the damage taken [...] (minimum **REDUCTION** of 1)"...

in other words, the `dice + intelligence modifier` is a minimum of 1, in order to reduce the damage received by 1 at least.
So yeah, the modifier can be -3, but the total formula minimum is 1.

dmportella:
> > `Protective Field. When you or another creature you can see within 30 feet of you takes damage, you can use your reaction to expend one Psionic Energy die, roll the die, and reduce the damage taken by the number rolled plus your Intelligence modifier (minimum reduction of 1), as you create a momentary shield of telekinetic force.`
> > So the minimum reduction you can have on damage is zero ( a minus one )
> 
> Yeah, I read it too before commenting and it confused me at first in the same way as you, but really, it's not the reduction to the dice based on intelligence modifier which is a minimum of 1 (i.e: "intelligence modifier minimum of -1") but rather the damage is reduced by a minimum of 1.
> 
> > "**REDUCE** the damage taken [...] (minimum **REDUCTION** of 1)"...
> 
> in other words, the `dice + intelligence modifier` is a minimum of 1, in order to reduce the damage received by 1 at least. So yeah, the modifier can be -3, but the total formula minimum is 1.

Yeah I know what you mean, I will fix that so that the mod still applied but the result can not be less then 1.

kakaroto (approved):

PR #1265, opened 2025-06-05
Foundry v13 fix

sirrio:
This one actually enables rolls for foundry again and fixes various errors which came up the on the way.

Feel free to change whatever is needed, i will probably work the next weeks on the deprecation errors too, so the plugin will use the new foundry roll renderer.

All rolls types worked for me and were rendered correctly

image

### DIFF
--- a/src/common/utils.js
+++ b/src/common/utils.js
@@ -74,11 +74,15 @@ function isExtensionDisconnected() {
 }
 
 // Taken from https://stackoverflow.com/questions/9515704/insert-code-into-the-page-context-using-a-content-script;
-function injectPageScript(url) {
+function injectPageScript(url, callback = null) {
     const s = document.createElement('script');
     s.src = url;
     s.charset = "UTF-8";
-    s.onload = () => s.remove();
+    s.onload = () => {
+        s.remove();
+        if (callback)
+            callback();
+    };
     (document.head || document.documentElement).appendChild(s);
 }
 
--- a/src/common/utils.js
+++ b/src/common/utils.js
@@ -117,11 +117,15 @@ function isRoll20(title) {
 }
 
 function isFVTT(title) {
-    return title.includes("Foundry Virtual Tabletop");
+    return title.includes("Foundry Virtual Tabletop") || title.includes("Foundry VTT");
 }
 
 function fvttTitle(title) {
-    return title.replace(" • Foundry Virtual Tabletop", "");
+    return title
+        .replace(" • Foundry Virtual Tabletop", "")
+        .replace(" • Foundry VTT", "")
+        .replace(" - Foundry VTT", "")
+        .replace(" - Foundry Virtual Tabletop", "");
 }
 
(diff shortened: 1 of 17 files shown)


src/fvtt/content-script.js
```diff
@@ -57,7 +57,10 @@ function injectSettingsButton() {
 
 const observer = new window.MutationObserver(titleSet);
 observer.observe(document.getElementsByTagName("title")[0], { "childList": true });
+if (document.getElementsByTagName("title")[0].textContent)
+    titleSet(null, observer);
```
kakaroto, line 61: This line seems to call `titleSet` if the title has a value, I assume as a way to avoid "missing" the observer change, but the title tag will always have a value, we're not using the observer to see it get set, we listen for it getting changed, which is done by the `fvtt_script.js` script, which we inject after attaching the observer, so we can't miss it anyway.

The title gets set only after the Foundry world has loaded and the chat UI has rendered so we don't register our tab to the background script until we can actually start sending rolls to it.
(note that the title gets changed to add "Foundry Virtual Tabletop" to it, which allows background script and other areas of the logic to recognize if a tab is running FVTT or not)

src/common/utils.js
```diff
@@ -113,11 +117,15 @@ function isRoll20(title) {
 }
 
 function isFVTT(title) {
-    return title.includes("Foundry Virtual Tabletop");
+    return title.includes("Foundry Virtual Tabletop") || title.includes("Foundry VTT");
 }
 
 function fvttTitle(title) {
-    return title.replace(" • Foundry Virtual Tabletop", "");
+    return title
```
kakaroto, line 124: This change shouldn't be necessary since Beyond20 is the one that sets the title and forces it to have " • Foundry Virtual Tabletop"

See https://github.com/sirrio/Beyond20/blob/641629eb130bf6ff33f2b1a0a9f30950ee68457e/src/fvtt/page-script.js#L573-L577

Having that string in it means that Beyond20 is properly loaded on it and the chat UI is rendered so it can receive rolls.

kakaroto (changes requested):
Thank you for this! I've reviewed the changes, everything looks good apart from one small misunderstanding on how Beyond20 uses the FVTT page's title.

src/fvtt/content-script.js
sirrio: I'm still not 100% sure how the injection is manage in the fvtt_script.js but it is not working correctly and I needed this bandaid to make rolls work again. Maybe I'm missing something. I added a condition to this titleSet call here to be more specific though.

kakaroto: Humm.. it's probably because of the check for the chat controls, or the `renderChatLog` hook might be different, so the page script isn't able to find the chat log and never sets the title. I'll check what it should be to propose a fix

kakaroto (approved):
Sorry for the delay. Thanks for the quick fix!

PR #1266, opened 2025-06-05
feature: Added support to roll both damages of toll the dead

dmportella:
# New Feature: Roll both damages for Toll the Dead

> [!IMPORTANT]
> I had to reuse the versatility field for getting this done, I am not happy with this use but i racked my brain for ages trying to think of a way to implement this, like damage groups, etc etc it quickly got complicated and finally i though that we could use the term Versatile for spells and weapons. This needs a lot of tidying up and optimisations, BUT the feature is pretty much done and working. I will optimise in the next few months.

# New setting

> New Setting for Toll the Dead allow user select if they want the new functionality - the default is the Query Dialog so it does not impact or break anything people are used too.

image

# Damage Description 

> Damage description uses the versatile code to display the separate damages, all is working right.

image

### DIFF
--- a/src/dndbeyond/base/utils.js
+++ b/src/dndbeyond/base/utils.js
@@ -164,10 +164,6 @@ async function queryCunningStrike() {
     return selection;
 }
 
-function upgradeDice(input, fromDie = 8, toDie = 12) {
-  let regex = new RegExp(`(\\d+)d${fromDie}`, 'g');
-  return input.replace(regex, `$1d${toDie}`);
-
 async function applyRogueSneakAttack(character, name, properties, damages,
                                      damage_types, roll_properties,
                                      settings_to_change) {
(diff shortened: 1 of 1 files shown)


dmportella:
@kakaroto this one should be fixed now there was a small merge conflict

src/dndbeyond/base/utils.js
```diff
@@ -292,10 +296,17 @@ async function buildAttackRoll(character, attack_source, name, description, prop
...
+        } else if (roll_properties.name.includes("Toll the Dead")) {
+            if (character.getSetting("toll-choice") === "both") {
+                damages.splice(1, 0, upgradeDice(damages[0], 8, 12));
+                damage_types.splice(1, 0, damage_types[0] + ' (Missing HP)');
+                damage_types[0] = damage_types[0] + ' (Full HP)';
+            } else {
+                const ttd_dice = await dndbeyondDiceRoller.queryGeneric(roll_properties.name, "Is the target missing any of its hit points ?", { "d12": "Yes", "d8": "No" }, "ttd_dice", ["d12", "d8"]);
+                if (ttd_dice === null) return null;
+                damages[0] = damages[0].replace("d8", ttd_dice);
+                damage_types[0] = damage_types[0] + ttd_dice === "d8" ? ' (Full HP)' : ' (Missng HP)';
```
kakaroto, line 308: Typo `Missng`

src/dndbeyond/base/utils.js
```diff
@@ -292,10 +296,17 @@ async function buildAttackRoll(character, attack_source, name, description, prop
...
-            damages[0] = damages[0].replace("d8", ttd_dice);
+        } else if (roll_properties.name.includes("Toll the Dead")) {
+            if (character.getSetting("toll-choice") === "both") {
+                damages.splice(1, 0, upgradeDice(damages[0], 8, 12));
+                damage_types.splice(1, 0, damage_types[0] + ' (Missing HP)');
+                damage_types[0] = damage_types[0] + ' (Full HP)';
+            } else {
+                const ttd_dice = await dndbeyondDiceRoller.queryGeneric(roll_properties.name, "Is the target missing any of its hit points ?", { "d12": "Yes", "d8": "No" }, "ttd_dice", ["d12", "d8"]);
+                if (ttd_dice === null) return null;
+                damages[0] = damages[0].replace("d8", ttd_dice);
```
kakaroto, line 307: Could replace this with `upgradeDice` too (or use the same `.replace`. I don't think we need the additional `upgradeDice` method for this simple replacement)

kakaroto (commented):
Humm... I'm not sure that I necessarily want to add this feature because it's quite specific and the prompt makes sense. I assume this is because you (or one of your players) uses TtD often and so answering the prompt every time gets annoying ... 🤔 
Anyways, I had some small notes but otherwise looks good (though hacky, as expected)

dmportella:
> Humm... I'm not sure that I necessarily want to add this feature because it's quite specific and the prompt makes sense. I assume this is because you (or one of your players) uses TtD often and so answering the prompt every time gets annoying ... 🤔 Anyways, I had some small notes but otherwise looks good (though hacky, as expected)

Yep you are not wrong, I had lot of people hit me up asking if this could be changed to just roll both damages at the same time to save time. There is a few games with multiple players that use this cantrip alot.

dmportella:
@kakaroto got all the changes done and i fixed some small issue i noticed

PR #1271, opened 2025-07-10
fix: tooltip position on fvtt

dmportella:
> [!NOTE]
> CSS changes to allow tooltip to show properly, all working

 
image
image

### DIFF
--- a/FVTT-module/beyond20/beyond20.css
+++ b/FVTT-module/beyond20/beyond20.css
@@ -131,6 +131,10 @@
 
 .beyond20-roll-cell,
 .beyond20-roll-die-result {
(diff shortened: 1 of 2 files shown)


kakaroto (commented):
I don't like this because it uses `!important` (which I've been told to never use) and it changes the overflow value for `.message-content` which will affect all messages, not just Beyond20 messages.

I found this to work just as well:
```css
.chat-message .message-content:has(> .beyond20-message) {
    overflow: visible;
}
```
I've pushed that fix instead.

dmportella:
Yep make sense I hasn't aware of the has operator hope it works ok

Issue #1274, opened 2025-07-21
Adding die together when rolling adv/disadv


Jackie-Knight:
**Describe the bug**
When I roll with advantage or disadvantage the die add instead of taking the highest or lowest number.

**To Reproduce**
1. When in a character sheet on dndbeyond
2. Right click on the Hit/DC
3. Select either Advantage or Disadvantag
4. In a VTT see that the roll is added together instead of taking the highest or lowest number

**Expected behavior**
To take the highest or lowest roll (based on rolling advantage or disadvantage). Instead it adds the two rolls together

![Image](https://github.com/user-attachments/assets/781d0720-2975-413b-ba42-4b80725af74e)

**Screenshots**
To help explain your problem, add screenshots of where you clicked in D&D Beyond, a screenshot of your Beyond20 character settings if applicable and the result you saw in the VTT.

**Browser Info (please complete the following information):**
 - Browser chrome
 - Version 137.0.7151.120 (Official Build) (64-bit)
 - If applicable, any dev console errors (F12 to open the console) from D&D Beyond and the VTT

**Additional context**
Add any other context about the problem here, such as a link to the character sheet or D&D Beyond page that shows the problem.
character sheet: https://www.dndbeyond.com/characters/147285962

DenzelKingsbury:
https://www.dndbeyond.com/characters/136699593 my character sheet is public. 
settings: 
<img width="698" height="944" alt="Image" src="https://github.com/user-attachments/assets/325e2093-45a2-4520-9426-4479ed1fdeb4" /> <img width="698" height="892" alt="Image" src="https://github.com/user-attachments/assets/2634241b-4cfe-4405-8a68-2baf5dbb0082" /> <img width="699" height="730" alt="Image" src="https://github.com/user-attachments/assets/ae09d9b4-cddb-40bf-bdbf-afb2ef53247f" />
----------------------------------------

Attaching console messages from Chrome and Firefox for comparison. Both performed now with fresh installs of the extension within  the browser. 

[Chrome - dndbeyond_character.js4494 Sending.txt](https://github.com/user-attachments/files/21368697/Chrome.-.dndbeyond_character.js4494.Sending.txt)
[Firefox - dndbeyond_character.js4494 Sending.txt](https://github.com/user-attachments/files/21368698/Firefox.-.dndbeyond_character.js4494.Sending.txt)

Jackie-Knight:
[...]

PR #1279, opened 2025-08-19
fix: digital dice adv and dis rolls

dmportella:
> [!NOTE]
> Had to fix the digital dice parsing, things changed so the results were not getting interpreted properly anymore,  fixes: https://github.com/kakaroto/Beyond20/issues/1274

# Results

image

image


kakaroto:
Was hard to parse, but once I figured out the issue, it made sense, thanks!

Issue #1299, opened 2025-09-27
Great Weapon Fighting Style damage being calculated incorrectly


joshuasimmons33:
The formula for GW fighting style has the modifier being added to the second dice roll in weapons with multiple damage dice (e.g., a greatsword), which is causing the comparison to the minimum 3 value to sometimes be evaluated incorrectly.

dmportella:
Can you provide examples please.

joshuasimmons33:
<img width="279" height="130" alt="Image" src="https://github.com/user-attachments/assets/86780d8d-2b2d-47bf-8e9e-6896d78961af" />

Shouldn't that +5 in the second comparison be outside the brackets for comparison?

dmportella:
I had further investigation around this turns out is a known behavior and one that will not as it is correct.

<img width="836" height="259" alt="Image" src="https://github.com/user-attachments/assets/581d6c90-2577-4dab-a4dd-e631d40446a7" />

> [!IMPORTANT]
> When you roll damage for an attack you make with a Melee weapon that you are holding with two hands, you can treat any 1 or 2 on a damage die as a 3. The weapon must have the [Two-Handed](https://www.dndbeyond.com/sources/dnd/free-rules/equipment#Two-Handed) or [Versatile](https://www.dndbeyond.com/sources/dnd/free-rules/equipment#Versatile) property to gain this benefit. 

The GW fighting means that you roll a minimum of 3 on EACH damage die -  So a great sword with **2d6 + 3** becomes **1d6min3 + 1d6min3 + 3** the modifier has not impact on the calculation.

| roll         | min | max |
|--------------|------:|:--------:|
| 2d6 +3 | 5 | 15 |
| 1d6min3 + 1d6min3 +3 | 9 | 15 |

You can see the results below.

<img width="515" height="297" alt="Image" src="https://github.com/user-attachments/assets/ae39e57c-805a-489d-8230-729225d60d4b" />

joshuasimmons33:
Here is the same greatsword with a +3 damage bonus, rolling a 1 and a 1 on the damage dice. 

<img width="323" height="192" alt="Image" src="https://github.com/user-attachments/assets/c4509d9c-42e3-4394-a190-8af553f9dc30" />

dmportella:
Please give me the link to your dndbeyond character so i can check

dmportella:
I am not seeing this issue, this why I asked before for a link to your character so that i can test.

Example of the double bladed scimitar working for me.

[...]

PR #1300, opened 2025-10-08
fix: roll20 genroll fix for min value rolls in roll20

dmportella:
> [!NOTE]
> Fixes -> https://github.com/kakaroto/Beyond20/issues/1299

Moved the modifier to the outside of the roll for any roll with min value

like **"1d4min3+9"** becomes **"{1d4, 0d0 + 3}kh1+9"** 

> [!WARNING]
> if the roll has 1d4min3 - 1 and you roll 1, the die should be min3 but the mod will turn it to 2.

### DIFF
--- a/src/roll20/content-script.js
+++ b/src/roll20/content-script.js
@@ -46,7 +46,8 @@ function escapeRoll20Macro(text) {
 
 function genRoll(dice, modifiers = {}) {
     dice = dice.replace(/ro<=([0-9]+)/, "ro<$1");
-    dice = dice.replace(/(^|\s)+([^\s]+)min([0-9]+)([^\s]*)/g, "$1{$2$4, 0d0 + $3}kh1");
+    dice = dice.replace(/(^|\s)+([^\s]+)min([0-9]+)([^\s]*)/g, "$1{$2, 0d0 + $3}kh1$4");
+  
     let roll = "[[" + dice;
     for (let m in modifiers) {
         let mod = modifiers[m].trim();


dmportella:
Thank you @kakaroto i will test things and see what happens.

dmportella:
Yes So @kakaroto I checked and the issue still persist with proposed changed. 

So I changed the regex to ensure it stop $4 before + / - / whitespace / end of string

**Original**

`dice = dice.replace(/(^|\s)+([^\s]+)min([0-9]+)([^\s]*)/g, "$1{$2, 0d0 + $3}kh1$4");`

**New**

`dice = dice.replace(/(^|\s)+([^\s]+)min(\d+)([^\s+\-]*?)(?=[+\-\s]|$)/g, "$1{$2$4, 0d0 + $3}kh1");`

* `([^\s+\-]*?)` grabs trailing roll suffixes like ro<1 but not +4 or -2.

* `(?=[+\-\s]|$)` peeks ahead to ensure we stop right before an arithmetic modifier, whitespace, or the end.

* Changed ([0-9]+) to (\d+) since it does the same thing

**Examples**

1d4min3+4 → {1d4, 0d0 + 3}kh1+4

1d4min3ro<1+4 → {1d4ro<1, 0d0 + 3}kh1+4

1d6ro<1min2-1 → {1d6ro<1, 0d0 + 2}kh1-1

I tested with a Halfing Fighter, Rogue, bard with reliable talent and silver tongue and great weapon fighting and it all seems to work alright now

Issue #1301, opened 2025-10-18
Custom Damages added twice for Wild Shapes


dmportella:
> [!NOTE]
> Ticket for Discord support request related to WIld Shape and Custom Damages being added twice.

> Kirbish — 10/10/2025 19:24
> Hey, not sure if anyone can help but I am having some issues with Beyond20 hotkeys. I have set up a hotkey to add 3 to a damage roll that I can use when I am in Wild Shape (+1 for Insignia of Claws and +2 for Circle of Mutation).
> 
> It works if I do it with normal attacks from my character sheet, but when I click to roll on an attack on a beast stat sheet, for some reason it adds the bonus twice. 
> Character attack shows:
> 
> Moon Sickle, +1
> 19
> Slashing Damage: 5
> Custom Damage: 3
> Total Damage: 8 
> Beast (Dire Wolf) attack shows:
> 
> Bite
> 8
> Piercing Damage: 8
> Custom Damage: 3
> Custom Damage: 3 
> As an aside note, is there a way to do +1 to the attack roll and then +3 to the damage with a single hotkey? At the minute I have it set up to work by holding two different hotkeys.

PR #1302, opened 2025-10-18
fix: stop custom damages being added twice and tiding up code

dmportella:
> [!NOTE]
> Fixes:  https://github.com/kakaroto/Beyond20/issues/1301

The code was adding the custom damages twice once for wild shape and then again if it was a attack which it is.

I consolidated the code again to use the util function instead and added the if statement to an else if

tested it both normal attacks and wild shape attacked and all is working

> [!IMPORTANT]
> Added a new feature to capture the Circle of Mutation class feature Circle Forms: Predatory Strike that adds extra damage to the wild shape form. includes setting (turned off by default) to enable it and disable it

image

image

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -938,6 +938,12 @@ const character_settings = {
         "type": "bool",
         "default": false
     },
+    "druid-circle-of-mutation-circle-forms": {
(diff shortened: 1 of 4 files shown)


kakaroto (approved):
LGTM!

PR #1303, opened 2025-10-21
fix: unknown combatant issue

dmportella:
> [!NOTE]
> Fixes issues an issue with Foundry where the token is not linked properly, Leaving it called "Unknown Combatant" 
image

### DIFF
--- a/src/fvtt/page-script.js
+++ b/src/fvtt/page-script.js
@@ -414,7 +414,7 @@ async function addInitiativeToCombat(roll) {
                         if (combatant) {
                             await game.combat.updateEmbeddedDocuments("Combatant", [{ "_id": docData(combatant)._id, "initiative": roll.total }]);
                         } else {
-                            await game.combat.createEmbeddedDocuments("Combatant", [{ "tokenId": token.id, "hidden": docData(token).hidden, "initiative": roll.total }]);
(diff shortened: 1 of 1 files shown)


src/fvtt/page-script.js
```diff
@@ -414,7 +414,7 @@ async function addInitiativeToCombat(roll) {
                         if (combatant) {
                             await game.combat.updateEmbeddedDocuments("Combatant", [{ "_id": docData(combatant)._id, "initiative": roll.total }]);
                         } else {
-                            await game.combat.createEmbeddedDocuments("Combatant", [{ "tokenId": token.id, "hidden": docData(token).hidden, "initiative": roll.total }]);
+                            await game.combat.createEmbeddedDocuments("Combatant", [{ "tokenId": token.id, "hidden": docData(token).hidden, "initiative": roll.total, "actorId": token.actor.id, "sceneId": token.scene.id, "img": token.actor.img }]);
```
kakaroto: A token could be unlinked from an actor, so `token.actor.id` could crash (Foundry has its own issues sometimes due to similar assumptions in some parts of their code).

kakaroto: yep, it's what I did before merging

dmportella: perfect. 👍🏼

PR #1304, opened 2025-10-22
fix: only apply blessed strikes to weapon attacks

dmportella:
> [!NOTE]
> fixes issue where Blessed Strikes are applying on cantrip that are not weapon attacks
image
image

# After fix

image
image

### DIFF
--- a/src/common/utils.js
+++ b/src/common/utils.js
@@ -30,18 +30,22 @@ function cleanRoll(rollText) {
     return rollText;
 }
 
-// Taken from https://stackoverflow.com/questions/45985198/the-best-practice-to-detect-whether-a-browser-extension-is-running-on-chrome-or;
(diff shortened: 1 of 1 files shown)


kakaroto (changes requested):
With some research, I see that we have a 2014/2024 issue here. The Blessed Strikes support was originally added in 2020, and was using the 2014 rules:
image

Then "Improved Blessed Strikes" was added in 2024, but that class feature was introduced in the 2024 version of the Cleric, which itself has its own different-flavor of Blessed Strikes:
image

So essentially, 2014 version applies radiant damage to cantrips OR weapon attacks, and the 2024 version applies to weapon attacks only but lets you choose radiant or necrotic damage. There's also Potent Spellcasting as another option for Blessed Strikes. I assume we'd need to pull what option the user chose from their class features list, to know how to apply it.

Will need to refactor to differentiate between 2014 and 2024, activate them differently based on version, check which option was chosen for the 2024 version and only apply Improved Blessed Strikes in the 2024 case.

Issue #1315, opened 2025-12-10
Rolls always "ask whisper" regardless of setting


S8RAKI:
**Describe the bug**
When I try to roll a dice, I will always be asked whether or not to whisper regardless of what I set the "whisper rolls" setting to.

**To Reproduce**
1. Go to any character sheet with a VTT open
2. Click on anything that rolls a dice (attack, ability check, etc)
3. See error (the dice roller in the bottom left corner asks whether you want to whisper or not)

**Expected behavior**
The dice rolls right after clicking the button (unless there is something necessary preventing it from doing so), and does not have you manually select whether to whisper or not if "whisper rolls" is not set to "always ask".

**Screenshots**
What happens when I click to roll a dice
<img width="264" height="49" alt="Image" src="https://github.com/user-attachments/assets/7e9a31e4-71d5-4527-9077-7d8982431028" />
Asking whether to whisper or not
<img width="245" height="449" alt="Image" src="https://github.com/user-attachments/assets/11d4dd18-dcd5-436d-b367-d70346460ece" />

**Browser Info (please complete the following information):**
 - Browser: Opera GX
 - Version: LVL 7, core 123.0.5669.76

**Additional context**
I believe this issue was caused by a D&DBeyond update rather than an issue with Beyond20 itself, though I've never used D&DBeyond without Beyond20 so I'm unsure.

GHKela:
I also had this issue (Firefox): Temp fix - Extention options, uncheck "Use D&D Beyond's Digital Dice". 
(With the option enabled I was unable to send any rolls other than initiative.)

kakaroto:
Fixed in the 2.17.0 release

PR #1316, opened 2025-12-10
fix: added roll button retry reroll

dmportella:
> [!NOTE]
> Fixes:  https://github.com/kakaroto/Beyond20/issues/1315 and all the other issues with the digital dice

This pattern:

At the moment _makeRoll runs, React/MUI hasn’t finished updating the DOM for that toolbar yet.

Possibly opening the whisper target menu (dice-toolbar__target-menu-button) before fallback.

Those actions can trigger a re-render or state update, so when the code immediately runs the selector, the element might not be “there” yet from the point of view of the current frame.

The retry implement will work around this issue.

### DIFF
--- a/src/dndbeyond/base/digital-dice.js
+++ b/src/dndbeyond/base/digital-dice.js
@@ -193,24 +193,22 @@ class DigitalDiceManager {
         }
     }
     static _clickRollButtonWithRetry(retries = 5, delay = 50) {
-        const tryClick = (remaining) => {
-            const $rollButton = $(".dice-toolbar.rollable .dice-toolbar__target button")
-                .not(".dice-toolbar__target-menu-button")
-                .first();
+        const tryClick = (remaining, _delay) => {
+            const $rollButton = $(".dice-toolbar.rollable .dice-toolbar__target button:not(.dice-toolbar__target-menu-button)").first();
 
             if ($rollButton.length) {
-                $rollButton[0].click();
+                $rollButton.click();
                 return;
             }
 
             if (remaining > 0) {
-                setTimeout(() => tryClick(remaining - 1), delay);
+                setTimeout(() => tryClick(remaining - 1, _delay), _delay);
             } else {
                 console.warn("DigitalDiceManager: roll button not found after retries");
             }
         };
 
-        tryClick(retries);
+        tryClick(retries, delay);
     }
     static _selectWhisperTargetWithRetry(retries = 5, delay = 50) {
         return new Promise((resolve) => {
@@ -223,7 +221,7 @@ class DigitalDiceManager {
             // Open the target selection menu
             menuButton.click();
 
-            const trySelect = (remaining) => {
+            const trySelect = (remaining, _delay) => {
                 const $options = $("#options-menu ul ul > div");
 
                 if ($options.length > 0) {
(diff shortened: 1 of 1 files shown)


src/dndbeyond/base/digital-dice.js
```diff
@@ -197,9 +197,31 @@ class DigitalDiceManager {
...
-            $(".dice-toolbar__roll, .dice-toolbar.rollable button:not(.dice-toolbar__target-menu-button)").click();
+            // At the moment _makeRoll runs, React/MUI hasn’t finished updating the DOM for that toolbar yet.
+            // so the roll buttons doesnt exist yet. We need to retry to roll the button a few times until it is done.
+            DigitalDiceManager._clickRollButtonWithRetry();
         }
     }
+    static _clickRollButtonWithRetry(retries = 5, delay = 50) {
+        const tryClick = (remaining) => {
+            const $rollButton = $(".dice-toolbar.rollable .dice-toolbar__target button")
+                .not(".dice-toolbar__target-menu-button")
```
kakaroto: Why did you change it from the `.dice-toolbar__roll, .dice-toolbar.rollable button:not(.dice-toolbar__target-menu-button)` selector?

kakaroto (changes requested):
Found a small issue and I had a question.
Thanks, I think the retry logic makes sense and is much easier than trying to wait for observer changes. I wonder if we could also have some kind of fallback method, for essentially doing "digital dice fails, fallback to rolling without it" in case someone's browser freezes for a few seconds or something and causes no DOM updates for more than 250ms..

src/dndbeyond/base/digital-dice.js
```diff
@@ -172,7 +172,7 @@ class DigitalDiceManager {
             dice.click()
         return amount || 0;
     }
-    static _makeRoll(roll) {
+    static async _makeRoll(roll) {
```
kakaroto: (Note to self) Will need to investigate further, but noticing a potential issue:
as `_makeRoll` is called by `_submitRoll` which itself is called by `rollDigitalDice` which is async and returns a promise for when the roll is completed. With the `_submitRoll` -> `_makeRoll` now becoming async and not actually rolling until the promise resolves, I think the async needs to be propagated upwards in the chain so the topmost async properly awaits the roll (or ensure that all the calls are not expecting it to be done when it returns)

dmportella: I think given the current code, a second call would never be run because of the `_pendingRolls.length === 1 check` the length would be 2 and does not call ` this._submitRoll(roll);` again. 

The Notification mis-assignment can happen, but that is because `_handleNewNotification` assumes FIFO over notifications and pendings, not because `_makeRoll` is async. We could refactor against the "wrong notification" case, but that would be a different refactor - tagging rolls with some id or recongnizable pattern or only parson notification that match. But is independent to `_makeRoll` being async. Two options we could do: 
1 - We could make `_makeRoll` sync again and keep all the whisper logic as fire and forget without returning a promise
OR
2 - We make `_submitRoll` and `rollDigitalDice` async and await on `_makeRoll` but dont change when the out promise resovles. that doesnt change the semantics of “when is the roll completed” (still `_finishPendingRoll`) but makes the chain more consistent if we ever want to await "roll has been sent to DNDBeyond" somewhere

kakaroto:
Merged manually, but then I had to rebase, so github doesn't recognize it as having been merged.

PR #1318, opened 2025-12-12
fix: added avatar roll fix

dmportella:
> [!NOTE]
> Added a small fix to allow avatar to be displayed in roll20


kakaroto:
Good catch!

PR #1322, opened 2025-12-12
docs: Add FAW entry for Roll20 Digital Dice double roll issue

dragooner4788:
- Adds new FAQ section explaining Digital Dice conflict
- Provides instructions on how to disable Digital Dice in Beyon20 Extension
- Clarifies difference from crit damage issue
- Includes troubleshooting steps for persiisten issues


kakaroto:
Thanks for the contribution, however, the entered FAQ entry is incorrect, so I won't add it to the FAQ.
Having or not the digital dice enabled would not be causing rolls to appear twice in roll20. In pretty much 100% of the cases over the years, every time someone complained about rolls appearing twice was because they had two tabs to their game open and hadn't noticed the second tab.

Issue #1329, opened 2026-01-15
Cannot Roll off of monster sheets.


GuywithDice:
When i open a monster sheet, the red boxes i normally can click to roll are no longer showing, ive gone up and down my extension, and have changed nothing, so i know it is not on my end.

If you go into a monsters sheet on dnd beyond, you will notice the boxes are missing.

I am using Chrome, but i dont see them on firefox either.

dmportella:
hmm do you have a monster, i can test? I have tried on sprite 2014 and 2024 and both are working
https://www.dndbeyond.com/monsters/17020-sprite
https://www.dndbeyond.com/monsters/4775845-sprite

GuywithDice:
https://www.dndbeyond.com/monsters/6081719-admiral-kobaryoken-full-cerberus is one of my monsters 
also heres a screenshot, the red boxes are gone as you can see

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/d27f14a3-a11d-4abf-8806-2c509b659030" />

GuywithDice:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/56803ef1-d1e1-4775-b36a-8a556a9231c5" /> this is what i see with sprite

GuywithDice:
<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/34a1053a-4953-48dc-9a6f-1089710d22b2" /> chrome as you can see

dmportella:
On the homebrew page there is no support for digital dice btw.
On the specific url for the monster.

the list page for all homebrew monsters the controls dont show, but if you go to the specific page for a monster it should show

# List page
https://www.dndbeyond.com/homebrew/monsters?filter-type=0&filter-search=copy_of&filter-cr-min=&filter-cr-max=&filter-armor-class-min=&filter-armor-class-max=&filter-average-hp-min=&filter-average-hp-max=&filter-is-legendary=&filter-is-mythic=&filter-has-lair=&filter-author=&filter-author-previous=&filter-author-symbol=&filter-rating=-11

<img width="1211" height="832" alt="Image" src="https://github.com/user-attachments/assets/a9dffabe-abb4-4527-a1e9-92b9aeae8c40" />

# Ape monster
https://www.dndbeyond.com/monsters/2172193-copy_of_ape

[...]

PR #1330, opened 2026-01-15
fix: updated browser dectection to use the user agent

dmportella:
> [!NOTE]
> Previous dectection logic no longer works with newest version of chrome: fixes https://github.com/kakaroto/Beyond20/issues/1329

Previous browser detection no longer works with chrome latest version.

I changed the code to use navigator.userAgent instead, and I test it for regex of each user agent.

All Brave, Opera, Safari will return Chrome
Firefox, firefix and edge will retiurn edge to keep in line with all the code that looks at firefox and chrome on the extension.

I have tested all browsers beside safari and digital dice, monster rolling etc work properly with the change.

### DIFF
--- a/src/common/utils.js
+++ b/src/common/utils.js
@@ -36,7 +36,8 @@ function getBrowser() {
     const ua = navigator.userAgent;
 
     if (/Firefox\//.test(ua)) return "Firefox";
-    if (/Edge\//.test(ua)) return "Edge";
+    // 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36 Edg/143.0.0.0'
+    if (/Edg\//.test(ua)) return "Edge";
     if (
         (/Brave\//.test(ua) || (navigator.brave && typeof navigator.brave.isBrave === "function")) ||
         (/OPR\//.test(ua)) ||


src/common/utils.js
```diff
@@ -30,18 +30,24 @@ function cleanRoll(rollText) {
...
-        if (typeof (browser) !== "undefined") {
-            return "Firefox";
-        } else {
-            return "Chrome";
-        }
-    } else {
-        return "Edge";
+    const ua = navigator.userAgent;
 
+    if (typeof browser !== "undefined" && browser.runtime && browser.runtime.getBrowserInfo) {
```
kakaroto: Shouldn't it call `getBrowserInfo` to check that it's Firefox? What if other browsers implement that API?

src/common/utils.js
```diff
@@ -30,18 +30,24 @@ function cleanRoll(rollText) {
...
+        return "Firefox";
     }
+
+    if (/Edg\//.test(ua)) return "Edge";
```
kakaroto: Should it be `Edg` or `Edge` ?

kakaroto (changes requested):
Thanks, I've noted a few comments for you to review

kakaroto (changes requested):
I checked all the call that calls `getBrowser()` and we only ever compare it to `Firefox`, so none of the other values matter.. might as well have called it `isFirefox`.
However, while testing, I'm not getting it to work on Firefox
image

And yeah, the docs say it should be supported in FF since version 51.
I'm trying to find a way for it to work reliably.

src/common/utils.js
dmportella: yep edge is better

src/common/utils.js
```diff
@@ -30,18 +30,24 @@ function cleanRoll(rollText) {
...
+    if (
+        (/Brave\//.test(ua) || (navigator.brave && typeof navigator.brave.isBrave === "function")) ||
+        (/OPR\//.test(ua)) ||
+        (/Safari\//.test(ua) && !/Chrome\//.test(ua) && !/Chromium\//.test(ua)) ||
+        (/Chrome\//.test(ua))
+     ) return "Chrome"; // if brave, opera, safari or chrome just return chrome
+    
+    return "Unknown";
```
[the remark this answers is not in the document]
dmportella: yep changed it default to chrome if matches nothing

kristiandueholm:
Seems like using UA is not recommended for detecting browser according to MDN docs.

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgent

> Theoretically this information is useful for detecting the browser and serving code to work around browser-specific bugs or lack of feature support. However, this is unreliable and is not recommended for the reasons given in [User-Agent reduction](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/User-agent_reduction) and [Browser detection using the user agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent).

kristiandueholm:
@dmportella Thank you for the rundown!

7337a87ca14b42d47970dc8a338bfafc2875dc3c solved Edge being detected as Chrome for me.

PR #1333, opened 2026-01-18
Strip customization action name

### LINKED ISSUE #1307: Tavern brawler damage reroll doesn't work if unarmed attack has custom damage modified
**Describe the bug**
When a monk has customized default unarmed attack and customization adds damage/to hit bonus, then rolling 1 on damage dice is not automatically rerolled with Tavern Brawler feat. 
Removing customization from an attack brings this back.

**To Reproduce**
1. Create Monk character
2. Select Tavern Brawler Origin feat
3. Roll damages with default unarmed strike
4. Notice how 1s are rerolled automatically!
5. Customize action and add flat to hit and to damage bonuses
6. Roll damages until 1 is rolled

**Expected behavior**
1s are rerolled as with attack without bonuses
**Actual behavior**
1s are not rerolled

**Screenshots**
Customization:
Image
Without customization:
Image
With customization:
Image

kristiandueholm:
This PR attempts to solve the issue of the Tavern Brawler effect not being applied when adding custom damage to Enhanced Unarmed Strike. This issue is described with detail in #1307.

**Cause**

When customizing an action, `action_name` get an asterisk (*) appended to the end. Even though the name is not customized, the check for whether the action name contains some sort of unarmed strike, fails. The original check uses `.includes` on the list, which in turns checks for exact matches. This is quite strict.

**Proposed solution**

~~Inspired by `isMeleeAttack`:~~

```
const isMeleeAttack = [
            "polearm master",
            "pole strike",
            "unarmed strike",
            "tavern brawler strike",
            "psychic blade",
            "bite",
            "claws",
            "tail",
            "ram",
            "horns",
            "hooves",
            "talons",
            "thunder gauntlets",
            "unarmed fighting",
            "arms of the astral self",
            "shadow blade",
            "predatory strike",
            "enhanced unarmed strike",
            "flurry of blows"
        ].some(action => action_name.toLocaleLowerCase().includes(action));
```

~~Essentially check not for full string equality, but just that some string in the list is contained within the action name.~~

Instead I strip asterisks at the end of the action name.

Furthermore, `isMeleeAttack` will be changed to use this normalization scheme as well.

**Test**

Tested by following the procedure in #1307. Before changes the issue was reproduced. After changes, the expected outcome of the damage being rerolled on a 1 occurred.

Closes #1307

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -494,9 +494,7 @@ function handleSpecialMeleeAttacks(damages=[], damage_types=[], properties, sett
     // enhanced unarmed strike
     if(to_hit !== null && 
         character.hasFeat("Tavern Brawler 2024") &&
-        ["enhanced unarmed strike", "unarmed strike", "flurry of blows"].includes(
-            normalizeActionName(action_name)
-        )
+        includesNormalized(["enhanced unarmed strike", "unarmed strike", "flurry of blows"], action_name)
     ) {
         damages[0] = damages[0].replace(/[0-9]*d[0-9]+/g, "$&ro<=1");
         effects.push("Tavern Brawler");


src/dndbeyond/content-scripts/character.js
```diff
@@ -892,6 +895,11 @@ function capitalize(str) {
     return str.replace(/\b\w/g, (c) => c.toUpperCase());
 }
 
+// Transform lowercase, remove trailing asterisk and spaces
```
dmportella: I would move this function to the utils file. https://github.com/kakaroto/Beyond20/blob/master/src/dndbeyond/base/utils.js
Likely use this in other places

kristiandueholm:
Should I introduce a const `normalizedActionName` and use that everywhere? I am unsure of how broad to make this change. Do we have tests to run, to make sure this does not mess stuff up?

I noticed this in `rollAction` which seemed strange to me as well:
```
const isRangedAttack = action_name.includes("Lightning Launcher");
```

dmportella:
I would keep this PR only tavern brawler and the melee change we can fix other places separately

kristiandueholm:
> How are you testing at the moment?

I just make up manual tests in my head of what the expected outcome is of my change.

I have a test character with a lot of rule books associated, so I also make sure that both 2014 and 2024 acts as expected.

PR #1337, opened 2026-01-26
feat: healer 2024 rerolls

### LINKED ISSUE #1336: 2024 Healer-Feat not implemented
**Describe the bug**
Characters with the 2024 Healer Feat do not automatically reroll 1's on their healing spells.

**To Reproduce**
1. Chreate any Character capable of casting healing spells (e.g. Druid) with the Healer Feat (2024 version)
2. Click on "Features & Traits" to have beyond20 read the feats
3. Cast a healing spell (healing word has an easy chance to roll a 1)
4. See that none of the 1's gets rerolled.

**Expected behavior**
1's on healing spells getting rerolled once. (like the foundry command "/r 2d4r=1+3")

**Additional context**
Partial quote of the feat:
"**Healing Rerolls.** Whenever you roll a die to determine the number of Hit Points you restore with a spell or with this feat’s Battle Medic benefit, you can reroll the die if it rolls a 1, and you must use the new roll."

kristiandueholm:
Closes #1336 

**Purpose**

Adds 2024 Healer feat rerolls as requested by #1336. The implementation is not too different from the Tavern Brawler feat.

The Healer 2024 feat is described in DnDBeyond as:

> Healer
> Origin Feat
> 
> You gain the following benefits.
> 
> Battle Medic. If you have a Healer’s Kit, you can expend one use of it and tend to a creature within 5 feet of yourself as a Utilize action. That creature can expend one of its Hit Point Dice, and you then roll that die. The creature regains a number of Hit Points equal to the roll plus your Proficiency Bonus.
> 
> Healing Rerolls. Whenever you roll a die to determine the number of Hit Points you restore with a spell or with this feat’s Battle Medic benefit, you can reroll the die if it rolls a 1, and you must use the new roll.

This PR aims to implement the second benefit *Healing Rerolls*.

**Changes**

- Adds the following checks to `handleSpecialHealingSpells`:
      - Character is version 2024
      - Has healer rerolls setting enabled
      - hasFeat Healer
- Injects reroll with regex if conditions are met
- Adds switch to popup

**Limitations and questions**

1. I am unsure how to interpret spells that heal multiple dice. I was unable to really find any discussions on forums that clarify this. As I understand, since it says "whenever you roll a die", even though we roll multiple, they can all be rerolled.
2. You are not able to rerolls only 1 die, if the setting is enabled. But then again, why would you?
3. Should this have a lock, like great weapon master?
4. Battle Medic, the first benefit Healer grants, has an injected roll button in the feat tab. I was not able to get the rerolls effect to trigger when rolling using this button. I debugged a bit, and I think this is because the roll type is "custom" which seems to be the Dice default? I am unsure where the injected +2 roll button is injected and how it triggers and roll. See image below.
> image

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -895,7 +895,7 @@ const character_settings = {
         "description": "Whenever you roll a die to determine the number of Hit Points you restore with a spell or " +
             "with this feat’s Battle Medic benefit, you can reroll the die if it rolls a 1, and you must use the new roll.",
         "type": "bool",
-        "default": false,
+        "default": true,
     },
     "great-weapon-master": {
         "title": "Feat: Great Weapon Master (Apply to next roll only, unless locked)",


src/dndbeyond/content-scripts/character.js
```diff
@@ -1616,6 +1616,15 @@ function handleSpecialSpells(spell_name, damages=[], damage_types=[], {spell_sou
 }
     
 function handleSpecialHealingSpells(spell_name, damages=[], damage_types=[], {spell_source="", spell_level="Cantrip", castas, settings_to_change}={}) {
+    // Feat: Healer
+    if (
+        character.hasFeat("Healer 2024") &&
+        character.getSetting("healer-rerolls-feat-2024", "false")
+    ) {
+        damages[0] = damages[0].replace(/[0-9]*d[0-9]+/g, "$&ro<=1")
+        damage_types.push("Healer")
```
dmportella: you dont need this damage type, the damage type is already "Healing"

dmportella:
The change is all good now, jsut remove the damage type override on the comment and it should all be oke

Issue #1339, opened 2026-02-07
drop down for Hotkey actions has 'undefined' entries


frudolph77:
**Describe the bug**
The drop down for Hotkey Actions has a bunch of undefined entries. There are much more than only the two from the screenshot

<img width="658" height="608" alt="Image" src="https://github.com/user-attachments/assets/11659032-ebed-44d1-bf00-97ae2e4215a8" />

**Expected behavior**
no undefined entries in the drop down :-) 

**Screenshots**
see above

**Browser Info (please complete the following information):**
 - Browser: Chromium
 - Version: 144

PR #1341, opened 2026-02-08
fix: added hidden flag to lock settings

dmportella:
> [!NOTE]
> fixes: https://github.com/kakaroto/Beyond20/issues/1339 Added hidden setting to lock settings so they dont show on the hotkeys page.


src/common/settings.js
kristiandueholm: Should we maybe update this to reflect the special keys like `hidden`?

dmportella: I never looked at that, we should.

PR #1343, opened 2026-02-18
Use native dice for custom rolls when DDB's Digital Dice are disabled

nox-4O4:
Currently, custom rolls in D&D Beyond still use Digital Dice, even when they are disabled in the Beyond20 settings.

This PR modifies the behaviour of custom rolls when Digital Dice are disabled so that the native Beyond20 dice roller is used instead. This results in faster rolls (almost instant) and ensures that custom rolls are still sent to VTT, regardless of the Digital Dice setting.

Additionally, the custom roll button now respects the Whisper settings.

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -2744,6 +2744,103 @@ function injectSettingsButton() {
     updateHotkeysList(hotkeys_popup);
 }
 
+let previousButton = null;
+function injectCustomRollButton() {
+    if (!DigitalDiceManager.isEnabled()) // nothing to do when digital dice menu does not exist
+        return;
+
+    let rollButton = document.querySelector("button[data-testid=diceRollButton]"),
+        resetButton = document.querySelector("button[data-testid=diceClearButton]");
+
+    // We have to wait until dice menu is completely initialized.
+    if (!rollButton)
+        return;
+
+    const currentlyInjected = previousButton !== null && rollButton.dataset.b20roll !== undefined;
+    if (!settings["use-digital-dice"]) { // inject or update button
+        if (!currentlyInjected) {
+            // store a reference to the previous button so we can revert our changes if the user switches the setting off again
+            previousButton = rollButton;
+
+            // remove existing click event by replacing element with itself
+            const newButton = rollButton.cloneNode(true);
+            rollButton.replaceWith(newButton);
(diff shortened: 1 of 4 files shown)


src/dndbeyond/content-scripts/character.js
```diff
@@ -2553,6 +2553,67 @@ function injectSettingsButton() {
...
+
+        // update label
+        let label = "To: Everyone";
+        if (key_modifiers.whisper || character.getGlobalSetting("whisper-type", WhisperType.NO) == WhisperType.YES)
+            label = "To: GM";
+        else if (character.getGlobalSetting("whisper-type", WhisperType.NO) == WhisperType.QUERY)
+            label = "To: (ask)";
+
+        const labelElement = document.querySelector(".dice-toolbar .dice-toolbar__target-user");
+        if (labelElement.textContent != label)
```
dmportella: possible null crash here, add

this would be better `if (labelElement && labelElement.textContent !== label)`

src/dndbeyond/content-scripts/character.js
```diff
@@ -2553,6 +2553,67 @@ function injectSettingsButton() {
...
+    }
+
+    let menuButton = document.querySelector(".dice-toolbar .dice-toolbar__target-menu-button"),
+        rollButton = document.querySelector(".dice-toolbar .dice-toolbar__target button:not(.dice-toolbar__target-menu-button)");
+
+    if (!menuButton || !rollButton) // buttons are not directly available at page load, so try again at a later mutation event
+        return;
+
+    if (settings['use-digital-dice']) { // revert custom button we injected previously
+        rollButton.replaceWith(previousButton);
```
dmportella: reverting can crash if previousButton is stale / already moved, it is was removed from the dom elsewhere or the toolbar got re-rendered you might be replacing a detached element that no longer has expected structure or events. make sure to check isConnected is false before reinserting.

dmportella (changes requested):
made a few observations other then that looks ok, @kakaroto any thoughts on it?

nox-4O4:
Wow, I really like the look of the new panel.

I updated the PR to work with the new custom rolls.

As the encounters page still uses the old custom dice bar, I've kept support for the previous version there. It's not very pretty, but as I expect them to eventually change the custom dice bar there, too, I think it's a good compromise between clean code and effort spent.

PR #1344, opened 2026-02-20
Roll20: Add support for HP sync with the 2024 character sheet templates

nox-4O4:
## Summary

This PR updates the Roll20 HP sync integration to support the "D&D 5e (New!) – 2024 / 2014" character sheet template, which is the default option for new users.

## Type of change
- [ ] 📝 Documentation
- [ ] 🔧 Build/CI
- [ ] 🚨 Breaking change

> ⚠️ If you used AI tools (e.g., ChatGPT, Copilot, Claude, etc.) to generate or substantially modify **code, tests, documentation, or review comments**, you **must** disclose it here and describe what was AI-assisted.

- [x] I did **not** use AI for this PR.
- [ ] I **did** use AI for this PR (describe below).

> ⚠️ **If AI-generated content is identified in this PR and AI usage was not disclosed, the PR will be rejected.**

## Motivation & context
During testing, I kept the character sheet template at the default setting ("D&D 5e (New!) – 2024 / 2014") and wondered why the HP did not synchronize to Roll20. As I thought that it would be nice for the HP sync to work with default settings as well, I took a stab at implementing this feature.

## What changed?
This change adds a mechanism that updates the HP values using the advanced sheet's functionality. It works by passing update messages to a headless background frame, which is responsible for state management.

The existing functionality for older character sheets remains unchanged.

This makes it possible to use any of the following templates:

| Character Template                    | Short Name (internally used) |
|---------------------------------------|---------------------|
| D&D 5E By Roll20 - 2014 Sheet Version | `ogl5e`             |
| D&D 5E By Roll20 - 2024 Sheet Version | `dnd2024byroll20`   |
| D&D 5E 2014                           | `ogl5e`             |
| D&D 5e (New!) - 2024 / 2014           | `dnd2024byroll20`   |
| D&D 5E (Community Contributed)        | `dnd5e`             |

## How to test
1. Create a new campaign in Roll20 using the "D&D 5e (New!) - 2024 / 2014" character sheet template. Start the campaign in GM mode.
1. Create a new character. Make sure that the character's name in Roll20 matches its name in D&D Beyond.
1. Add the character's token to the map in Roll20. In the token settings, add bars for the `hp` and `hp_temp` attributes.
1. Change the HP, max HP, and temp HP settings on the character sheet in D&D Beyond, and observe them changing in Roll20 accordingly.

### DIFF
--- a/src/roll20/page-script.js
+++ b/src/roll20/page-script.js
@@ -116,12 +116,11 @@ async function updateHP(name, current, total, temp) {
                 },
             });
 
-            // set HP & temp HP using the property setter.
-            await character.characterSheet.headlessRelay.setComputed({
-                characterId: character.id,
-                property: "hp",
-                args: [current],
-            });
+            // set HP using the property setter
+
+            /* **TEMPORARY WORKAROUND**
+               Using the property setter is currently broken on Roll20's end, as they use the wrong key for storing the temporary HP (temp vs tempHP).
+               So, the current workaround is to just forgo the setter and change the store value directly, where we can determine the key ourselves.
 
             // If we wanted to auto-hide the bar when temp is zero, we'd have to manually edit the temporary HP integrant's value and set it to an empty string (ref. changing max HP).
             // The property setter that's used with setComputed automatically converts values to integers, so an empty string remains zero and the bar is still visible.
@@ -131,6 +130,33 @@ async function updateHP(name, current, total, temp) {
                 property: "hp_temp",
                 args: [temp],
             });
+            */
+            const tempHpIntegrantKey = Object.entries(store.integrants.integrants).find(([_, v]) => v.hitpointType === "Temporary")?.[0],
+                tempHPIntegrantUpdatePayload = tempHpIntegrantKey ? { integrants: { integrants: { [tempHpIntegrantKey]: { valueFormula: { flatValue: temp } } } } } : {};
+            character.characterSheet.headlessRelay.postMessage({
+                type: "change",
+                character: {
+                    id: character.id,
+                    attributes: {
+                        store: {
+                            hitpoints: {
+                                temp,
+                                tempHP: temp,
+                            },
+                            ...tempHPIntegrantUpdatePayload,
+                        },
+                    },
+                },
+            });
+            /* **END OF TEMPORARY WORKAROUND** */
+
+            // set HP using the property setter
+            await character.characterSheet.headlessRelay.setComputed({
+                characterId: character.id,
+                property: "hp",
+                args: [current],
+            });
+
         }
         // make sure all tokens resemble the new values
         character.updateTokensForAdvancedSheets();


nox-4O4:
> We never supported the 2024 version of the character sheet in roll20.

Ah, that makes sense. I had suspected that I must be missing something obvious here.

> Currently testing it and I am not able follow the "Set any max HP value in the character sheet on Roll20." ?
> 
> it seems to be missing a step.

I expanded upon the issue description. One has to edit the character sheet directly to change the max HP value.

> by the way was that what you meant as `Set any max HP value in the character sheet on Roll20` ?

Exactly :)

> Ideally, we do not want player to have to follow too many steps to get the HP syncing working. 

I agree – the fewer steps a user has to take before getting the sync working, the better.
I think I've found a way to have the plugin perform the initialization step of setting an initial max HP value itself, so the player wouldn't have to change anything apart from the character name.

I will take a closer look at this tomorrow evening. 

> There is another issue I have found with the 2024 sheet
> 
> the changes you have doesnt seem to support changes to the max hp modifier.

Now, that's curious. I used the max HP modifier extensively during testing, and it worked as intended.
If I can't reproduce the erratic behavior, I'll add some debugging output to help pin down the issue.

I'm marking the PR as draft until I could take a closer look.

nox-4O4:
I've made some progress and updated the PR. Most parts should be working now, but extensive testing is still necessary.

There are still some edge cases that require user intervention, which I hope to smooth out:
- Sometimes, Roll20 does not initialize _any_ attributes at all after character creation. In this case, the player has to manually open the Character Sheet once (but does not need to change anything).  
  I'm not sure what the best solution is here. Perhaps, the plugin could open the Character Sheet dialog automatically? Alternatively, a pop-up telling the user what to do might suffice.

- When a bar on the token is linked to the character's HP before the max HP entry is created, the bar won't register any max HP changes. 
  The user can always reload the page to resolve this issue. I'll look into whether the plugin could unlink and relink the bar to the token automatically. Also, I'll take a look at how Roll20 handles this case.

@dmportella what are your thoughts on this?

dmportella:
> I've made some progress and updated the PR. Most parts should be working now, but extensive testing is still necessary.

Understood, testing is not a huge issue.

> 
> There are still some edge cases that require user intervention, which I hope to smooth out:
> 
> * Sometimes, Roll20 does not initialize _any_ attributes at all after character creation. In this case, the player has to manually open the Character Sheet once (but does not need to change anything).

Is this only the 2024 sheet or does it affect the 2014 sheet?

>   I'm not sure what the best solution is here. Perhaps, the plugin could open the Character Sheet dialog automatically? Alternatively, a pop-up telling the user what to do might suffice.

:/

yea, i dont think we should open the character sheet on creation, ppl will eventually get annoyed with sheets opening all the time when using the extension. when you say open the sheet what do you mean, can you give me the steps so i can try to reproduce? 

> * When a bar on the token is linked to the character's HP before the max HP entry is created, the bar won't register any max HP changes.
>   The user can always reload the page to resolve this issue. I'll look into whether the plugin could unlink and relink the bar to the token automatically. Also, I'll take a look at how Roll20 handles this case.

can we check the max hp is set before trying to set the hp and hp temp?

I am going to test your latest changes tomorrow morning and see what i can do too.

nox-4O4:
> I know the Product Manager of Roll20, worse comes to the worse we can reach out and let them know.

That's good to know, but I don't think it makes sense to reach out to them unless we're certain that the problem is with Roll20. Even then, it would depend on the impact.

> I'm not sure if this is some race condition, a connectivity issue, or something else, but I suspect the issue is on Roll20's end.

To clarify, this is pure speculation. I did not investigate the cause in any great detail.

> * The only reason to use this new sheet is the black and red look-and-feel.
> * There is ZERO value to the user using beyond20 if they use 2024 sheets over the 2014.
> * The REASON to use beyond20 is that so you dont need to use the sheets provided by roll20/foundry.

I do see value in having the default Roll20 settings work out of the box, without requiring the user to configure anything other than the character's name. 
In fact, we had another session today with a DM who is new to Roll20, and setting up the characters and tokens was quite a hassle (or at least more of a hassle than it should have been).

That being said, I realize that the gains are rather low beyond that, even though I personally prefer the updated dice template, though.

For now, I don't want to give up on the PR just yet – at least, not while it's still fun to work on. But I don't put the priority too high, either.

dmportella:
> > I know the Product Manager of Roll20, worse comes to the worse we can reach out and let them know.
> 
> That's good to know, but I don't think it makes sense to reach out to them unless we're certain that the problem is with Roll20. Even then, it would depend on the impact.
> 
> > I'm not sure if this is some race condition, a connectivity issue, or something else, but I suspect the issue is on Roll20's end.
> 
> To clarify, this is pure speculation. I did not investigate the cause in any great detail.
> 
> > * The only reason to use this new sheet is the black and red look-and-feel.
> > * There is ZERO value to the user using beyond20 if they use 2024 sheets over the 2014.
> > * The REASON to use beyond20 is that so you dont need to use the sheets provided by roll20/foundry.
> 
> I do see value in having the default Roll20 settings work out of the box, without requiring the user to configure anything other than the character's name. In fact, we had another session today with a DM who is new to Roll20, and setting up the characters and tokens was quite a hassle (or at least more of a hassle than it should have been).
> 

Indeed hundred percent, I much rather that the user just set their name and that is it, but I also know that it may not be possible which is the nature of the project unfortunately.

> That being said, I realize that the gains are rather low beyond that, even though I personally prefer the updated dice template, though.
> 

The dice template is much better in my opinion. So I am there with you.

> For now, I don't want to give up on the PR just yet – at least, not while it's still fun to work on. But I don't put the priority too high, either.

I was not advocating for you to give up at all. It has been something I been wanting to support for a while now, I had a partially working branch, but at the time the sheet was still not completed (roll20 was still developing it). So I put in a pause (just before xmas last year).

If we can resolve this it will be great. But we have to be pragmatic.

nox-4O4:
A quick update:

> * Sometimes, Roll20 does not initialize _any_ attributes at all after character creation. In this case, the player has to manually open the Character Sheet once (but does not need to change anything).
>   I'm not sure what the best solution is here. Perhaps, the plugin could open the Character Sheet dialog automatically? Alternatively, a pop-up telling the user what to do might suffice.

I think I found a way to manually call the initialization code without the user having to open any dialogs.
I'll take a closer look at it later this week. Depending on why the initialization failed the first time, this may or may not help.

> * When a bar on the token is linked to the character's HP before the max HP entry is created, the bar won't register any max HP changes.
>   The user can always reload the page to resolve this issue. I'll look into whether the plugin could unlink and relink the bar to the token automatically. Also, I'll take a look at how Roll20 handles this case.

The problem seems to be that editing the store directly is a bit too low-level, so some of the other components don't receive the changes. The intended way is probably to first update the character sheet state and then synchronize the changes to the store.

During manual testing I was able to initialize and update the maximum HP this way. The HP bar, which I had linked to the HP attribute _prior to_ creating the maximum HP attribute, still followed the changes accordingly.

I'll try to push an update to the PR in the next few days.

dmportella:
@nox-4O4 testing again before merging and for some reason the hp_temp is not syncing on a campaign using the new 2024 character sheets (with only the 2024 sheets available)

here is the character to test: https://www.dndbeyond.com/characters/162398760

to confirm the character sheet i am using 
image

I recreated the character a few times, did a few test with "editing the sheet directly waiting etc" etc and HP syncs but the temp doesnt seem too

The same character sheet but using 2014 is working as intented.

The d&d community sheets are working too

its just the temp hp on the 2024 version

Issue #1345, opened 2026-02-24
Digital Dice is not Showing up After dndbeyond update 24/02/2026


dmportella:
**Describe the bug**
After the update to the dice roller of dndbeyond, the digital dice no longer shows.

This is because the HTML has changed (dice roller is now a new typescript react component that uses different controls and settings to roll dice. Additionally the Notification Bar that used to show on the bottom right of the screen is not longer there. All roll results now are posted in the gamelog.

**To Reproduce**
1. Go to Any character or Encounter page.
2. With digital dice enabled on the beyond20 extension.
3. Click on Any ability, skill, attack doesnt matter.
4. No digital dice are rolled (however the fallback still works and makes the roll anyway using beyond20 rolls).
5. No Digital Dice is rolled.
**Expected behavior**
The digital dice should roll and post the results in the VTT of choice.

**Screenshots**
To help explain your problem, add screenshots of where you clicked in D&D Beyond, a screenshot of your Beyond20 character settings if applicable and the result you saw in the VTT.
> [!NOTE]
> # New Dice Roller
> <img width="346" height="492" alt="Image" src="https://github.com/user-attachments/assets/49bf05a8-a86a-4098-a4af-226f2031479d" />

**Browser Info (please complete the following information):**
 - Browser (Any Browser)
 - Version (All versions)

dmportella:
> its only if you dont click the beyond 20 dice so if you click that it still rolls to roll20 just doesnt appear in dnd beyond

Yep

morderkaine1:
I have the same issue. When Beyond20 is installed I can only see digital dice when clicking the Dnd Beyond dice roller in the bottom left of he screen but clicking on any skill, attack, save, etc, the digital dice do not show. The extension breaks the digital dice roller.

mashb1t:
same for me, it's only showing the text prompt when rolling dice with the Beyond20 extension enabled.

fixed by https://github.com/kakaroto/Beyond20/pull/1346

dmportella:
all issues should be resolved in the latest release -> https://github.com/kakaroto/Beyond20/releases/tag/v2.18.0

PR #1346, opened 2026-02-25
fix: new digital dice for character pages

dmportella:
## Summary
Updated the D&D Beyond Digital Dice integration to work with the new toast + Game Log UI. The roller now waits for the toast, parses the correct Game Log entry to obtain full dice results, and reliably applies those results back to Beyond20 roll objects.

D&D Beyond Digital Dice no longer uses the legacy `noty` bar, so we can’t parse roll results from notifications anymore. Instead, we parse the transient toast message and the Game Log sidebar entry that contains the full breakdown.

The Digital Dice flow is now fully asynchronous. We also updated the manager to work with the new D&D Beyond UI elements. Several issues were caused by React state/effects not being settled when buttons were clicked too quickly, so small delays were added anywhere we interact with the new dice roller. Users may notice the roller feels slightly slower as a result; the dice roll itself is fast, but the UI/state pipelines in D&D Beyond require additional settling time to behave consistently.

## Type of change
- [ ] 📝 Documentation
- [ ] 🔧 Build/CI
- [ ] 🚨 Breaking change

Used number of tools to propose updates to the DigitalDiceManager/DigitalDice logic, including DOM selector changes for the new DDB Game Log UI, toast-driven timing, notation matching adjustments, and moving/removing message broker triggers that interfered with digital dice rolls. Most of the code was code was human written.

## Motivation & context
- Related issue(s): https://github.com/kakaroto/Beyond20/issues/1345
- Context:
  - D&D Beyond Digital Dice UI changed (old `noty` notifications removed, new toast + Game Log entries introduced).
  - Existing Beyond20 integration broke because it relied on old notification parsing and because early message broker events could interfere with the new digital dice roll flow.
  - Dice notation ordering in the DDB custom roller is not guaranteed to match Beyond20 formula ordering, causing false mismatches during Game Log parsing.

## What changed?
- Switched Digital Dice result handling from legacy notifications to toast-first + Game Log parsing.
- Fixed Game Log breakdown parsing to read actual dice results (e.g., `title="6 + 7"`) instead of accidentally reading SVG `title="D20"`.
- Implemented robust dice assignment from Game Log results back into Beyond20 dice objects and recalculated totals.
- Made Game Log notation matching order-insensitive via canonicalized dice notation to handle DDB’s internal dice ordering (checked the actual typescript code from dndbeyond dice roller to verify how they did things internally).
- Removed `MBPendingRoll` / `MBPendingRoll=null` dispatch from `_makeRoll()` because it breaks digital dice when DDB “roll-to-game-log” is enabled; roll-to-game-log support is deferred until after digital dice parsing completes (via the fulfilled flow).

## How to test
1. Enable D&D Beyond Digital Dice and ensure the dice roller popup is available.
2. Roll a simple custom roll (any ability, skill, attack, spell etc, e.g., `1d20 + 15`) from Beyond20 and verify:
   - a toast appears,
   - the Game Log entry is detected and parsed,
   - the displayed Beyond20 total includes modifiers (e.g., raw `d20=6` becomes `21` with +15).
3. Roll a multi-dice sequence (e.g., `1d20 + 8`, `2d6 + 2`, `1d8`) and verify:
   - parsing succeeds even if the Game Log notation order differs,
   - dice results are assigned correctly and totals are computed correctly.
4. (Regression) Enable DDB “roll-to-game-log” setting and verify digital dice still rolls successfully (no hang/no fallback), and Beyond20 results still compute after parsing.
5. Encounters page still uses the old digital dice.

## Reviewer notes
- Game Log entry selection is toast-driven to avoid opening the Game Log too early (which can suppress toast/log updates).
- Notation matching is canonicalized (order-insensitive) because DDB’s custom roller orders dice based on its own state key order, not formula order.
- `MBPendingRoll` was removed from `_makeRoll()` intentionally because it interferes with the new digital dice pipeline; roll-to-game-log posting should happen after parsing completes (fulfilled stage).

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -90,6 +90,7 @@ See the [Screenshots](screenshots) page for a video tour of some of the features
 * Feat: Elven Accuracy (Via super advantage option)
 * Feat: Flames of Phlegethos
 * Feat: Polearm Master
+* Feat: Healer 2024
 * Aasimar: Radiant Soul
 * Artificer: Arcane Firearm
 * Artificer: Alchemical Savant
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,6 +25,17 @@ const DNDBEYOND_PAGE_DEPS = [
     "src/generic-site/renderer.js",
 ];
 
+const DNDBEYOND_CHARACTER_PAGE_DEPS = [
+    ...ROLL_RENDERER_DEPS,
+    "src/dndbeyond/base/digital-dice-new.js",
+    "src/dndbeyond/base/dice.js",
+    "src/dndbeyond/base/base.js",
+    "src/dndbeyond/base/hotkeys.js",
+    "src/dndbeyond/base/utils.js",
+    // Second renderer on DDB pages to allow rednering roll to the DOM
+    "src/generic-site/renderer.js",
+];
+
 const SRC_FILES = {
     background: [
         "src/common/constants.js",
@@ -121,7 +132,7 @@ const SRC_FILES = {
         "src/dndbeyond/content-scripts/vehicle.js",
     ],
     dndbeyond_character: [
-        ...DNDBEYOND_PAGE_DEPS,
+        ...DNDBEYOND_CHARACTER_PAGE_DEPS,
         "src/dndbeyond/base/spell.js",
         "src/dndbeyond/base/monster.js",
         "src/dndbeyond/base/extras.js",
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -901,6 +901,13 @@ const character_settings = {
         "default": false,
         "hidden": true
     },
+    "healer-rerolls-feat-2024": {
+        "title": "Feat: Healer - Healing Rerolls",
+        "description": "Whenever you roll a die to determine the number of Hit Points you restore with a spell or " +
+            "with this feat’s Battle Medic benefit, you can reroll the die if it rolls a 1, and you must use the new roll.",
+        "type": "bool",
+        "default": false,
+    },
     "great-weapon-master": {
         "title": "Feat: Great Weapon Master (Apply to next roll only, unless locked)",
         "description": "Apply Great Weapon Master -5 penalty to roll and +10 to damage",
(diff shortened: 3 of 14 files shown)


dmportella:
@nox-4O4 those message broker messages are the ones sent by us i believe, after the roll is taken we publish those events, so they are not created by dndbeyond but by us after we get the roll information

nox-4O4:
> those message broker messages are the ones sent by us i believe

That was my first assumption, too. But I don't think that's the case. Given how the message broker is registered, it looks like it hooks into D&D Beyond's internal message broker. Also, I took the above screenshot using the current master branch (without the changes from this PR).

nox-4O4:
The messages that Beyond20 exchanges are sent with `runtime.sendMessage` to/from the background page, whereas the message broker, that `DDBMessageBroker` hooks into, further goes to the `game-log-api-live.dndbeyond.com` websocket (for characters that have joined a campaign, otherwise the messages are dropped):

image

nox-4O4:
@dmportella here you go: https://github.com/nox-4O4/Beyond20/commit/50c9b4d79f8c8b6ab98334adf8a0c38e904a048a

The tricky part is accessing the `rollDice` helper in the first place, as it's buried in the depths of ES Modules. Currently, I'm hijacking the object initialization phase for module access. If you know a better way, please do tell. 😅

Everything else is just retrieving the configured/stored values (dice set, character information, etc.).
While it would be possible to also use some (fragile?) DOM parsing to get the values, I'm using D&D Beyond's API directly, as I believe it to be more stable.

Currently, there is hardly any error handling in the code. (Also, I've left a couple of minor TODOs.)

nox-4O4:
> If we rely on the game log to parse rolls through the message broker, we wont be able to decouple from the gamelog, i.e. we will always have to interact with the gamelog when using the digital dice.

I don't think that's a significant drawback. When Digital Dice are enabled, the game log is used for the results anyway. Besides, we only interact with the message broker, not the game log directly.

dmportella:
On thing i will say is that we need to extra careful with changing the manifest it can cause a ton of issues with the stores and peoples installations

PR #1354, opened 2026-03-17
fix: Digital dice 2026 merged

dmportella:
> [!IMPORTANT]
> merged code with highjack code.


nox-4O4:
Sometimes, clicking the button quickly results in empty rolls being sent to Roll20. I haven't been able to reproduce this reliably. Most of the time, it seems to work fine. Digital Dice were enabled during this test.

image

Also, the roll formula seems to be missing the modifiers when this happens. The test character has a +6 for perception checks, so, a 5 should not be possible.

Issue #1381, opened 2026-05-06
Beyond20: No VTT Found when using with Roll20


butlerofdeath:
**Describe the bug**
When I attempt to roll through my DnDbeyond character sheets the dice will roll in the tab, but the results will not show in Roll20. Instead I now receive a message saying "No VTT found that matches your settings. Open a VTT window, or check that the settings don't restrict access to a specific campaign." Attempting to view Beyond20 settings in the Roll20 tab now displays "Not a D&D Beyond or VTT page" regardless of my addon settings.

Everything was functioning just fine yesterday and I've made no changes to my Beyond20 or Firefox settings in-between.

**To Reproduce**
1. Select any skill or feature to roll while both Roll20 and DnDBeyond are open.
2. Click the skill to attempt to roll it publicly.
3. See error.

**Expected behavior**
Typically the roll result would appear in the chat area of my Roll20 tab.

**Screenshots**

<img width="530" height="302" alt="Image" src="https://github.com/user-attachments/assets/11d89148-f874-40cb-a73c-8ddaced59278" />

<img width="730" height="207" alt="Image" src="https://github.com/user-attachments/assets/d180cb90-3f62-42da-95f3-7de49be08200" />

<img width="624" height="121" alt="Image" src="https://github.com/user-attachments/assets/6c815fb4-7d21-4531-9792-3ddbd43ee3c0" />

**Browser Info (please complete the following information):**
 - Firefox
 - Version 150

**Additional context**
This occurs in both standard and Incognito tabs and across all of my games and character sheets.  Example sheet https://www.dndbeyond.com/characters/164122506/Kg3Wn0

fencer-x:
Also reporting this issue with Firefox 150.0.1 -- with the additional context that this problem does not exist in Chrome 147.0.7727.117. Seems to be a similar issue to #1071.

dmportella:
Thx I will check it out

mathiasose:
Based on what I read on #1071 I changed my URL from https://app.roll20.net/editor to https://app.roll20.net/editor/ (adding the trailing slash) and reloaded D&D Beyond and after that it began working again for me 🤷

coinich:
> Based on what I read on [#1071](https://github.com/kakaroto/Beyond20/issues/1071) I changed my URL from https://app.roll20.net/editor to https://app.roll20.net/editor/ (adding the trailing slash) and reloaded D&D Beyond and after that it began working again for me 🤷

[...]

PR #1386, opened 2026-05-12
roll20: Page Permission request for roll20 editor page.

dmportella:
## Summary

Fixes Beyond20 failing to send rolls when the Roll20 editor tab has no trailing slash (`/editor` vs `/editor/`).

## Type of change

Drafting this PR description

## Motivation & context

When a user opens `https://app.roll20.net/editor` (no trailing slash), the existing `ROLL20_URL = "*://app.roll20.net/editor/*"` and manifest content‑script match pattern never find this tab. Beyond20 cannot deliver rolls and shows "No VTT found."

On Firefox, `currentPermissions.origins` contains `*://app.roll20.net/*` automatically (the manifest's `/editor/` permission is normalized to the origin level), so the host permission is always present - no grant flow or user interaction is needed.

## What changed?
- **Constants (`src/common/constants.js`)**: Added `ROLL20_URL_NO_SLASH = "*://app.roll20.net/editor"` for no‑slash tab detection.
- **Background (`src/extension/background.js`)**:
  - Added `roll20_tabs` tracking array with `addRoll20Tab`/`removeRoll20Tab` helpers.
  - `sendMessageToRoll20` - falls back to `roll20_tabs` when the `ROLL20_URL` query misses the no-slash tab (both limit and non‑limit paths).
  - `onRollFailure` - queries `ROLL20_URL_NO_SLASH` directly instead of iterating all tabs. Checks `hasRoll20Permission` against `currentPermissions.origins`. On Firefox this is always true, so it injects scripts, tracks the tab, and resends the roll automatically - no user prompt needed.
  - `activate-icon` - tracks Roll20 tabs via `addRoll20Tab`.
  - `onTabsUpdated` - removal tracking when the tab URL changes away from Roll20; auto‑injects into no‑slash tabs on `status === "complete"` when host permission exists.
  - `onTabRemoved` - includes `removeRoll20Tab(id)`.

No changes to `popup.js`, `utils.js`, or the manifest. No grant popups, no permission dialogs, no two‑step flows - Firefox handles it transparently because the host permission is already granted at the origin level.

## How to test
1. Open `https://app.roll20.net/editor` (no trailing slash) in a Firefox tab.
2. On any D&D Beyond character sheet, click a Beyond20 roll button.
3. The roll should go through to Roll20 immediately - no prompts, no dialogs.
4. Verify the same works with `https://app.roll20.net/editor/` (with slash).

## Reviewer notes
- Firefox normalizes `*://app.roll20.net/editor/` to `*://app.roll20.net/*` in `permissions.getAll()`, so `hasRoll20Permission` is always true. Chrome MV3 preserves the exact path, so this fix is Firefox-specific; Chrome already works via the existing manifest content scripts for the slash variant.


src/dndbeyond/base/utils.js
```diff
@@ -1025,6 +1025,37 @@ function beyond20SendMessageFailure(character, response) {
         return;
     console.log("Received response : ", response);
     if (["roll", "rendered-roll"].includes(response.request.action)  && ((response.vtt && response.vtt[0] == "dndbeyond") || response.error)) {
+        if (response.roll20TabFound) {
+            // A Roll20 editor tab was found but lacks the content script (no trailing slash).
+            // Prompt the user to grant permission.
+            dndbeyondDiceRoller._prompter.prompt(
+                "Beyond 20",
+                "<p>Beyond20 detected a Roll20 editor tab but doesn't have the necessary permissions to load on it.</p>" +
```
kakaroto: I'd like to change the text to something a bit more user-friendly. Maybe "Beyond20 detected a Roll20 game it cannot send your rolls to. Please authorize Beyond20 to send messages to your Roll20 page"

src/dndbeyond/base/utils.js
```diff
@@ -1025,6 +1025,37 @@ function beyond20SendMessageFailure(character, response) {
...
+            ).then((result) => {
+                if (result) {
+                    chrome.runtime.sendMessage({action: "roll20-permission-granted"}, (resp) => {
+                        if (resp && resp.success) {
+                            if (resp.popup) {
+                                // Chrome: permission dialog popup was opened
+                                alertify.success("A permission dialog has been opened. Please grant access in the popup window, then roll again.");
+                            } else {
+                                // Firefox: permission was granted and scripts injected
+                                alertify.success("Beyond20 is now active on your Roll20 editor tab. Please roll again.");
```
kakaroto: Let's not ask users to "roll again". Instead, we should just do a resend of the original `response.request` message.

src/extension/background.js
```diff
@@ -158,6 +174,27 @@ function removeCustomTab(id) {
...
+function addRoll20Tab(tab) {
+    if (isRoll20TabAdded(tab)) return;
+    roll20_tabs.push(tab);
+    console.log("Added ", tab.id, " to roll20 tabs.");
+}
+
+function removeRoll20Tab(id) {
+    for (let t of roll20_tabs) {
+        if (t.id == id) {
+            roll20_tabs = roll20_tabs.filter(tab => tab !== t);
```
kakaroto: Why loop on `roll20_tabs` just to do a `filter` in the end? If it's to check if the tab is in the list, that won't matter since `filter` will be a noop if the tab isn't there, so it shouldn't be necessary in the first place.
But if it's needed for some reason, it would be nicer to do like this instead of the `for` loop:
```js
if (roll20_tabs.some(tab => tab.id === id)) {
   roll20_tabs = roll20_tabs.filter(tab => tab.id !== id);
}
```

src/extension/background.js
```diff
@@ -179,10 +216,28 @@ function onRollFailure(request, sendResponse) {
                 "error": "Found a Foundry VTT tab that has not been activated. Please click on the Beyond20 icon in the browser's toolbar of that tab in order to give Beyond20 access."
             })
         } else {
-            sendResponse({
-                "success": false, "vtt": null, "request": request,
-                "error": "No VTT found that matches your settings. Open a VTT window, or check that the settings don't restrict access to a specific campaign."
-            })
+            // Check if there's a Roll20 editor tab without trailing slash
+            chrome.tabs.query({}, (allTabs) => {
```
kakaroto: Why query all tabs then check if the url matches the ROLL20_URL_NO_SLASH? 
Why not query for ROLL20_URL_NO_SLASH directly and get an already filtered list of tabs?

src/extension/background.js
```diff
@@ -270,6 +328,49 @@ function onMessage(request, sender, sendResponse) {
...
+                        return;
+                    }
+                    console.log("Opened permission grant popup window", window?.id);
+                    sendResponse({success: true, popup: true});
+                });
+            });
+        });
+        return true;
+    } else if (request.action == "inject-roll20-scripts") {
+        chrome.tabs.get(request.tabId, (tab) => {
```
kakaroto: This should not be necessary. Once the user gives the permission, it should autodetect/inject the scripts without any external request, just like it currently does for all the custom_domains.
The inject should be on `onTabsUpdated`  on that check for `status == complete`.
Only time it would loop through tabs and manually check/inject scripts would be at the moment the user gives authorization.

src/extension/background.js
```diff
@@ -179,10 +216,28 @@ function onRollFailure(request, sendResponse) {
...
             })
         } else {
-            sendResponse({
-                "success": false, "vtt": null, "request": request,
-                "error": "No VTT found that matches your settings. Open a VTT window, or check that the settings don't restrict access to a specific campaign."
-            })
+            // Check if there's a Roll20 editor tab without trailing slash
+            chrome.tabs.query({}, (allTabs) => {
+                const roll20Tab = allTabs.find(tab =>
+                    tab.url && urlMatches(tab.url, ROLL20_URL_NO_SLASH) && isRoll20(tab.title)
```
kakaroto: It checks that there is a roll20 tab then asks the user for permission but doesn't seem to check if the permission had already been previously granted.

kakaroto (changes requested):
Started the review, went well, felt good, but I got distracted and by the time I got back to it, I couldn't finish it anymore, brain has shut down :( 
Sending partial review for now.

src/dndbeyond/base/utils.js
dmportella: yep can do that

kristiandueholm:
Reference #1381

dmportella:
@kakaroto new release it work with out any popup I found a way to get it work on firefox with out any extra permissions being needed

src/extension/background.js
```diff
@@ -71,7 +79,17 @@ function sendMessageToRoll20(request, limit = null, failure = null) {
             failure(true)
         }
     } else {
-        sendMessageTo(ROLL20_URL, request, failure);
+        sendMessageTo(ROLL20_URL, request, (failed) => {
+            console.log("[Roll20] non-limit path: query failed:", failed, "roll20_tabs:", roll20_tabs.length);
+            if (failed && roll20_tabs.length > 0) {
```
kakaroto: Why only send to some roll20 tabs when it fails to send to tabs with the other URL? 
I think it should do
```js
sendMessageTo(ROLL20_URL, request, (failed) => {
	for (let tab of roll20_tabs) {
	    sendMessageWithLog(tab.id, request)
	}
	if (failure) failure(failed && roll20_tabs.length === 0);
});
```

src/extension/background.js
```diff
@@ -179,10 +212,47 @@ function onRollFailure(request, sendResponse) {
...
+                const roll20Tab = roll20Tabs.find(tab => isRoll20(tab.title));
+                if (roll20Tab) {
+                    // Check if permission has already been granted
+                    const hasRoll20Permission = currentPermissions.origins.some(pattern =>
+                        urlMatches("https://app.roll20.net/", pattern)
+                    );
+                    console.log("[Roll20] no-slash tab found:", roll20Tab.url, "hasPermission:", hasRoll20Permission);
+                    if (hasRoll20Permission) {
+                        // Already have permission; inject scripts and resend the roll
+                        injectRoll20Scripts([roll20Tab]);
```
kakaroto: Should not be needed since you already auto inject in the tab navigation monitoring function. This should only check if there is a tab. The FVTT message is to click on the beyond20 icon in toolbar, maybe something similar here for roll20 tab.
That being said, I wonder if we do a "we got new permissions, let's see what tabs exist now that we can inject into" and also, if we can do the request permissions from here directly, we should maybe apply that to the "fvtt tab found" workflow too!

src/extension/background.js
```diff
@@ -179,10 +212,47 @@ function onRollFailure(request, sendResponse) {
...
+                                });
+                            }
+                        });
+                        return;
+                    }
+                    // Found a Roll20 tab that doesn't match the /editor/* pattern
+                    // The content script won't auto-run here due to manifest match pattern
+                    sendResponse({
+                        "success": false, "vtt": null, "request": request,
+                        "error": "Beyond20 detected your Roll20 editor tab but needs permission to enable Beyond20 on it. Would you like to grant access?",
```
kakaroto: I don't see the code anymore for granting the permissions, so this does nothing and `roll20tabUrl` and `roll20TabId` are unused.

kakaroto (changes requested):
A few issues found, and also, looks like you made all the files executable
Image

dmportella:
@kakaroto done changes here, and tested it. it is all working on my side

kakaroto (approved):
LGTM!

PR #1375, opened 2026-04-23
fix: dice regex matches bare D1/D0 in sourcebook text (#1126)

0xguy07:
## Summary

Fixes #1126. The `DICE_REGEXP` in `src/common/utils.js` used `\d*[dD]\d+` for the dice core, which allows *zero* digits before `d`. That makes bare `D0` and `D1` match the regex, so sourcebook labels like the `D1` door in Vecna: Eve of Ruin get swapped out for the Beyond20 roll icon.

This PR narrows the bare form to `[dD](?:[2-9]|\d{2,})` while leaving `NdM` (explicit count) completely untouched. d0 and d1 are never real dice, so this removes the false positive without changing behavior for any real roll notation.

## What still matches (no regressions)

- `d2`, `d3`, `d4`, `d6`, `d8`, `d10`, `d12`, `d20`, `d100`
- `D20`, `D12` (uppercase)
- `1d20`, `2d6+3`, `10d6`, `3d8 + 1d4`
- Explicit `1d1`, `2d1`, `10d1`, `1d0` (preserved for any internal formulas that stand in a fixed value)
- Suffixes: `2d10ro=1`, `d4min2`
- Pure modifiers: `+5`, `-3`

## What no longer matches (the bug)

- Bare `D1`, `d1`, `D0`, `d0`
- `"touching the following doors: A2, B1, C1, D1."` (the exact string from the issue)
- `"Door D1 is locked."`

## Test plan

Regex tested against 23 inputs (bug cases + positive cases + edge cases) — all pass on the new regex, 5 fail on the old.

- [ ] Open the Vecna: Eve of Ruin ritual chamber page on D&D Beyond and confirm `D1` renders as text, not a dice icon
- [ ] Roll a spread of dice on a character sheet (`d20`, `2d6+3`, `1d8 + 2d6`, `d100`) — all still become roll buttons
- [ ] Verify any internal/homebrew formulas that use explicit `1d1` still work (e.g. fixed-damage stand-ins)

Issue #1126, opened 2024-05-21
(Minor) Replacement of D1 false positive


Jonfor:
**Describe the bug**
In the new Vecna: Eve of Ruin book there is a section describing doors with labels on a map. One of these doors is labeled as D1 that is incorrectly getting converted into the Beyond20 diceroll png.

**Warning that there are potential spoilers ahead for the new book!**

**To Reproduce**
1. Go to [veor](https://www.dndbeyond.com/sources/veor/eve-of-ruin#R3RitualChamber)
2. Scroll down to `Characters can teleport to Vecna’s central chamber by touching the ruby gemstone on one of the following doors: A2, B1, C1, D1.`
3. See custom20.png instead of D1

**Expected behavior**
Expected to see D1 instead of the custom20.png

**Screenshots**
![Capture](https://github.com/kakaroto/Beyond20/assets/3246528/915dc44d-370f-4a30-8ab5-d35713203a71)

**Browser Info (please complete the following information):**
This is in Firefox 126.0 Desktop on Windows 10 Pro 22H2 with Beyond20 extension version 2.9.3

**Additional context**
[Here](https://www.dndbeyond.com/sources/veor/eve-of-ruin#R3RitualChamber) is the relevant page

Jonfor:
There are other similar issues on the same page - should I report them all on this issue?

kakaroto:
thanks for the report. I do see the other d1/d2 getting a similar treatment.. I don't know how exactly we'll fix this, but I'll give it a try.

dmportella:
that still valid. obviously it is a difficult one since it clashes with dice notation

PR #1389, opened 2026-06-05
fix: dice regex matches bare D1/D0 in sourcebook text (#1126)

0xguy07:
## Summary
Restrict the bare-die `dN`/`DN` form in `DICE_REGEXP` to N=2–9 or N with two-or-more digits, so impossible bare dice like `D0`/`D1` no longer hijack sourcebook map labels (e.g. `A2, B1, C1, D1` in Vecna: Eve of Ruin). Explicit-count forms (`1d1`, `2d0`, …) are unchanged. A standalone `npm test` script is added covering the bug and existing-behavior regressions.

## Type of change
- [x] 🧪 Tests (adds or updates tests)
- [ ] 📝 Documentation
- [ ] 🔧 Build/CI
- [ ] 🚨 Breaking change

**What was AI-assisted:** The regex change in `src/common/utils.js` and the test cases in `test/dice_regex.test.js` were drafted with Claude (Anthropic, Opus 4.7). I (the PR author) reviewed every change, ran the bundled test script, and confirmed the behavior on the issue's reproducer sentence and on standard dice expressions. No build/CI tooling, documentation, or unrelated code was modified by AI.

This is a resubmission of #1375 — the prior PR was closed after @dmportella asked about AI use; this one discloses it upfront and adds the test coverage that the prior thread noted regex changes need.

## Motivation & context
- Related issue(s): #1126
- Prior PR: #1375 (closed without merge)
- Context: The current `DICE_REGEXP` accepts `\d*[dD]\d+`, which matches bare `D0` and `D1`. Neither is a real die (a 0-sided die is undefined and a 1-sided die has no roll variance), but both occur as map/door labels in published sourcebooks, producing a `custom20.png` icon where the original label should be (issue #1126 screenshot).

## What changed?
- `src/common/utils.js`: tightened the bare-die alternative in `DICE_REGEXP` from `\d*[dD]\d+` to `\d+[dD]\d+|[dD](?:[2-9]|\d{2,})`. Applied in both the primary dice group and the trailing-extra-dice modifier group so compound expressions like `1d8+1d6+3` still parse. Comment added inline explaining the rationale.
- `test/dice_regex.test.js` (new): 25-case standalone Node script. Loads `DICE_REGEXP` directly from `src/common/utils.js` at runtime (no bundler needed) and exits non-zero on any failure.
- `package.json`: added `"test": "node test/dice_regex.test.js"`.

## How to test
1. `git fetch && git checkout fix/d1-d0-false-positive-1126`
2. `npm test` — should print `25/25 passed`.
3. Manual: load the unpacked extension build (`npm run build:chrome` / `:firefox`), open the [Vecna ritual chamber page](https://www.dndbeyond.com/sources/veor/eve-of-ruin#R3RitualChamber), scroll to the doors sentence — `D1` should render as text, not a Beyond20 icon. Sanity check a known-good page (e.g. any monster stat block with `1d20` / `2d6+3`) and confirm dice buttons still appear.

## Reviewer notes
- **Scope is intentionally narrow:** only `D0` and `D1` are excluded. The original issue thread noted `D2` also collides with map labels, but `d2` is a legitimate (if uncommon) die, so excluding it would regress real dice notation. This PR fixes the impossible-die cases and leaves the `D2+` ambiguity as a known trade-off (worth a follow-up if a context-aware approach is wanted).
- **Test framework choice:** I deliberately did not pull in Jest/Mocha — the test file is a single ~150-line Node script with zero new dependencies. Happy to migrate it to whatever you prefer if you'd rather grow this in a different direction.
- **AI use:** disclosed above; commit body also notes it. I'm the human responsible for the change and have verified the behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
         "doc": "docs"
     },
     "scripts": {
+        "test": "node test/dice_regex.test.js",
         "gulp": "gulp",
         "clean": "gulp clean",
         "build": "run-s clean build:*",
--- a/src/common/utils.js
+++ b/src/common/utils.js
@@ -1,4 +1,8 @@
-const DICE_REGEXP = /(^|[^\w])(?:(?:(?:(\d*[dD]\d+(?:ro(?:=|<|<=|>|>=)[0-9]+)?(?:min[0-9]+)?)((?:(?:\s*[-−+]\s*\d+)|(?:\s*[0+]\s*\d*[dD]\d+))*))|((?:[-−+]\s*\d+)+)))($|[^\w])/gm;
+// Bare-die forms `dN`/`DN` are only accepted when N is 2-9 or has two-or-more digits.
+// This excludes `D0`/`D1`, which are not real dice and produced false positives on
+// sourcebook map labels (issue #1126: "doors: A2, B1, C1, D1"). Forms with an explicit
+// count (`0d1`, `1d1`, `2d0`, ...) are unchanged so existing behavior is preserved.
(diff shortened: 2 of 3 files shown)


0xguy07:
@dmportella I ensured I used the correct template this time. I just want to assist in fixing some of the issues which are on the list. We use this extension a lot during our games, and I want to ensure I can help in the ways that I can regarding ensuring that I contribute.

dmportella:
Yep that is good, seems to be working and test will ensure that things work ok, I will them to the build too.

Issue #1388, opened 2026-06-04
versatile weapon always rolls higher dice


delanthear:
**Describe the bug**
If you have beyond20 installed and enabled, and you are using a versatile weapon such as a longsword, it will always roll the max damage die, irrelevant of if you clicked on the smaller die or not.

If beyond20 is not enabled, the correct dice is rolled.

**To Reproduce**
1. Install and enable beyond 20
2. Equip a longsword
3. Click on the D8 damage dice
4. Observe a D10 rolled for damage.

**Expected behavior**
A d8 to be rolled when the D8 button is clicked.

**Screenshots**
https://drive.google.com/file/d/1CP68sDKdmfTuJcdeTqGPzyJC7XuY5Q6O/view?usp=sharing

**Browser Info (please complete the following information):**
 - Affects both Firefox (151.0.3) and Chrome (148.0.7778.179)
 - If applicable, any dev console errors (F12 to open the console) from D&D Beyond and the VTT
From the Inspector console when clicking on the D8 button.
```
Beyond20: New side panel is : b20-item-pane
dndbeyond_character.js:12084 EXECUTING QUICK ROLL!
dndbeyond_character.js:11220 Beyond20: Executing panel : b20-item-pane false true true
dndbeyond_character.js:5519 Damage to crits :  1d10+7 ['1d10']
dndbeyond_character.js:5084 Sending message:  {action: 'rendered-roll', request: {…}, title: 'Sword of Answering', html: '<div class="beyond20-message"><div class="beyond20…span></div></div></div></span></span></div></div>', character: 'Lorazai', …}
dndbeyond_character.js:6416 Received response :  {success: true, vtt: Array(1), error: null, request: {…}}
```

**Additional context**
This seems to have been present for a few updates now.

dmportella:
That is confirmed, I will check the fix for this ty

PR #1390, opened 2026-06-05
fix: versatile weapon quick-roll picks d8 button correctly (#1388)

0xguy07:
## Summary
Replace a fragile `action.previousElementSibling !== null` check in the quick-roll wire-up with a scoped lookup inside `.ddbc-combat-attack__damage`. The previous check incorrectly set `force_versatile = true` for the d8 button on a versatile weapon whenever DDB rendered any label/icon node ahead of the first dice container, so clicking the d8 rolled 1d10 instead of 1d8 (issue #1388).

## Type of change
- [ ] 📝 Documentation
- [ ] 🔧 Build/CI
- [ ] 🚨 Breaking change

**What was AI-assisted:** Root-cause analysis from the issue's console trace and the `closest()` / `querySelectorAll()` rewrite were drafted with Claude (Anthropic, Opus 4.7). I (the PR author) reviewed the patch and the surrounding `activateQuickRolls` / `rollItem` flow before pushing. No test file is included because this code only runs against the live DDB DOM.

## Motivation & context
- Related issue: #1388
- Reporter's console trace shows `Executing panel : b20-item-pane false true true` → `force_versatile = true` was passed even though the user clicked the d8. That flag flows into `rollItem` where `force_versatile = true` forces `versatile_choice = "two"` (longsword 1d10), producing the reported `Damage to crits : 1d10+7`.

## What changed?
- `src/dndbeyond/content-scripts/character.js` (`activateQuickRolls`): when wiring up `actions_damage`, find each action's nearest `.ddbc-combat-attack__damage` ancestor, count the `.integrated-dice__container` children, and set `isVersatile = true` only when this action is the last of more-than-one container. Behavior matches the original intent ("2-handed/versatile is the second dice container") but is no longer affected by any label/icon/wrapper nodes DDB renders alongside the first container.

## How to test
1. `git fetch && git checkout fix/versatile-quick-roll-1388`
2. `npm run build:chrome` (or `:firefox`) and load the unpacked build.
3. On a character with a versatile weapon (Longsword, Quarterstaff, Spear, Warhammer, etc.):
   - Hover the **d8** damage die in the action bar → confirm the rolled damage is `1d8 + mods`, not `1d10`.
   - Hover the **d10** damage die → confirm the rolled damage is `1d10 + mods` (2-handed).
4. On a non-versatile weapon (Shortsword, Greatsword, Dagger…): confirm the single damage button still rolls correctly.
5. With quick-rolls disabled in settings: confirm dice still roll as before (this path is gated on `settings["quick-rolls"]` and is unchanged).

## Reviewer notes
- **DOM hypothesis caveat:** I do not have access to a live DDB session for verification. The fix preserves the original branch behavior (`length > 1 && last container` is equivalent to `previousElementSibling !== null` whenever the dice containers are the only children) so it should be a no-op on DOM shapes where the prior code worked, and only changes behavior on shapes that previously broke. If you have a quick reproducer or DOM snippet handy, it would be useful to confirm.
- Scope intentionally minimal: only the four lines wiring `actions_damage` are touched. `rollItem` / `force_versatile` semantics are unchanged.

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -2988,7 +2988,16 @@ function activateQuickRolls() {
         activateQRAction(action, true, false);
     }
     for (let action of actions_damage.toArray()) {
-        activateQRAction(action, false, true, action.previousElementSibling !== null);
+        // force_versatile = true only when this is the 2-handed damage button,
+        // i.e. the second dice container in the weapon's damage area. Using
+        // previousElementSibling alone breaks when DDB renders any label/icon
+        // node ahead of the first container (issue #1388).
+        const damageArea = action.closest(".ddbc-combat-attack__damage");
+        const containers = damageArea
+            ? damageArea.querySelectorAll(".integrated-dice__container")
+            : [];
+        const isVersatile = containers.length > 1 && containers[containers.length - 1] === action;
+        activateQRAction(action, false, true, isVersatile);
     }
 
     const activateQRSpell = (spell, force_to_hit_only, force_damages_only) => {


dmportella:
Yeah testing its not working I have a bigger fix for this. I will have to reject this PR for now. I had to add some more code to also enforce the 1 handed attack etc.

thank you for this PR tho helped me figure out where to start looking. Also apologies for reviewing so late.

I am testing all the PR's you published and I will be updating those if there is any additional work required.

PR #1400, opened 2026-07-08
fix: force one handed or two handed dmg

dmportella:
## Summary
Fix versatile weapon damage quick-roll so the 1‑handed button rolls 1‑handed damage and the 2‑handed button rolls 2‑handed damage, regardless of DDB label/icon siblings or the user's "versatile-choice" setting.

## Type of change

> AI (Claude) was used to analyze the DOM structure, identify both code paths affected by the same bug, and draft the fix code. The logic and correctness were manually verified.

## Motivation & context
- Related issue(s): #1388
- Context: When clicking the 1‑handed damage button on a versatile weapon (e.g. Longsword 1d8+5), the quick-roll either rolled 2‑handed damage (1d10) or both damages, depending on the DOM layout. Two root causes:

  1. **Wrong sibling detection** – `previousElementSibling !== null` was meant to detect the 2‑handed button, but DDB renders label/icon nodes before the first button, making `previousElementSibling` non-null for both buttons.
  2. **Missing "one‑handed" signal** – `force_versatile` was a boolean (`true` → force 2‑handed, `false` → use user default of `"both"`). The 1‑handed button had no way to request *only* 1‑handed damage.

## What changed?
- **`activateQuickRolls`** (line 2992): Replaced `action.previousElementSibling !== null` with a position check using `parentElement.querySelectorAll(".integrated-dice__container")` — scoped to the button's immediate parent, immune to non-container siblings.
- **`handleCombatAttackIntegratedDie`** (line 3264): Same fix for the integrated‑dice hijack path that had the identical bug.
- **`force_versatile` → tristate**: Changed from `boolean` to `false` / `"one"` / `"two"`. `false` means no force (respect user setting), `"one"` forces 1‑handed, `"two"` forces 2‑handed. This is used in `rollItem` (line 1147) to set `versatile_choice` accordingly.

## How to test
1. Open any DDB character sheet with a versatile weapon (e.g. Longsword, Quarterstaff).
2. Click the 1‑handed damage button (e.g. `1d8+5`) — verify the roll uses 1‑handed damage only.
3. Click the 2‑handed damage button (e.g. `1d10+5`) — verify the roll uses versatile damage only.
4. Repeat with a non‑versatile weapon to confirm its damage rolls are unaffected.

## Reviewer notes
- The fix uses `parentElement` (the `ddbc-combat-item-attack__damage` container) rather than `.closest(".ddbc-combat-attack__damage")` to avoid counting buttons from sibling damage‑type containers (e.g. bonus fire damage on a Flame Tongue Longsword).
- Both code paths (`activateQuickRolls` for Beyond20 tooltips, `handleCombatAttackIntegratedDie` for DDB native dice hijack) are fixed with identical logic.

### DIFF
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -1144,8 +1144,10 @@ async function rollItem(force_display = false, force_to_hit_only = false, force_
                         versatile_choice = "one"
                     if (key_modifiers.versatile_two_handed)
                         versatile_choice = "two";
-                    if (force_versatile) {
+                    if (force_versatile === "two") {
                         versatile_choice = "two";
+                    } else if (force_versatile === "one") {
+                        versatile_choice = "one";
                     }
                     if (versatile_choice == "one") {
                         damages.push(damage);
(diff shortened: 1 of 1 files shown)


dmportella:
tested on all browsers.

PR #1402, opened 2026-07-30
feat: add support for the 2024 Bladesinger Bladesong

jugarrit:
## Summary

Adds a `Wizard: Bladesinger: Bladesong (2024)` character setting implementing the revised Bladesong (Forgotten Realms: Heroes of Faerûn), alongside the existing 2014 option. The two are mutually exclusive per character, following the version-split convention already used for features like Assassinate.

## Type of change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [x] ♻️ Refactor (no functional change)
- [ ] 📝 Documentation
- [ ] 🔧 Build/CI
- [ ] 🚨 Breaking change

All source changes, the commit message, and this PR description were written by Claude (Claude Opus 5, via Claude Code). My involvement was directing the design, reviewing every change, and testing against a live D&D Beyond character sheet.

Specific points where I pushed back and the AI's initial approach was changed, in case they're useful review signal:

- Its first draft auto-deactivated the setting when the character was Incapacitated or attacked two-handed. No feature in this codebase reads conditions, and every `settings_to_change` write is a one-shot "apply to next roll only" toggle, so that was removed as off-pattern.
- It initially excluded unarmed strikes with a negated `includesNormalized(...)` check. Song of Victory — the direct 2014 predecessor — has no such check, so it was dropped rather than introduce a fourth idiom for "is this a weapon".
- Bladework originally rendered as a separate labelled damage entry. Since it *replaces* the ability modifier rather than adding a source, it now merges into the roll.
- The `usesTwoHandedDamage` extraction was my request, to avoid duplicating the Versatile/Two-Handed logic that already existed for Charger 2024.

## Motivation & context

- Related issue(s): #
- Context: The Bladesinger in *Forgotten Realms: Heroes of Faerûn* reworks Bladesong. The 2014 version's benefits (Acrobatics advantage, a concentration bonus) partly overlap with the new one, but **Bladework** — using Intelligence for weapon attack and damage rolls — is new, and D&D Beyond does not model it. There is no control on the sheet to activate Bladesong, so DDB's displayed to-hit is always the plain Strength/Dexterity value.

## What changed?

- New `wizard-bladesong-2024` setting; the existing `wizard-bladesong` is untouched in behaviour.
- `getFeatureVersionName` now treats `FRHoF` as a 2024 rules source and includes `bladesong` in the name list, so the feature parses as `Bladesong 2024`. Added to the **name-gated** branch only — deliberately not alongside `free-rules`, which applies to every feature from that source regardless of name.
- **Agility**: advantage on Acrobatics, now tagged with a `Bladesong` effect on the roll.
- **Focus**: adds the Intelligence modifier to Constitution saves for concentration. Floored at 0 for 2024, since that wording is optional ("you *can* add"); the 2014 bonus keeps its printed minimum of +1.
- **Bladework**: uses Intelligence for the attack and damage rolls of proficient weapons when it beats Strength/Dexterity, merged into the rolls via a new `mergeFlatBonus` helper rather than added as a separate term.
- Song of Victory is scoped to the 2014 Bladesong so it cannot stack with Bladework.
- Extracted `usesTwoHandedDamage`, now shared with the Charger 2024 Great Weapon Fighting check.

Deliberately **out of scope**, as the extension cannot observe or apply them: the AC bonus, the +10 speed, donning armour or a shield, and use tracking / the 1-minute duration.

## How to test

1. `npm install && npm run build`, then load `build/chrome` as an unpacked extension.
2. Open a Bladesinger with the FRHoF subclass and visit **Features & Traits** once so class features are re-parsed.
3. In the Beyond20 popup, confirm **Wizard: Bladesinger: Bladesong (2024)** appears and the plain "Bladesong" option does not. Enable it.
4. Roll **Acrobatics** — advantage, tagged `Bladesong`.
5. Roll an attack with a proficient **one-handed** weapon where Intelligence exceeds Strength/Dexterity — the to-hit and damage should each be higher by the difference, as single values rather than a separate term.
6. Roll a **Constitution** saving throw and answer yes to the concentration prompt — Intelligence is added.
7. Roll with a **two-handed** weapon, or a Versatile weapon with `versatile-choice` set to "both"/"two" — no Bladework applies.
8. Regression: a 2014 (Tasha's) Bladesinger should still show only the original "Bladesong" option and behave as before.

## Reviewer notes

- **`getWeaponAbilityModifier` is the only ability-substitution construct in the codebase, and I'd welcome scrutiny on it.** Every other `getAbility()` call adds a modifier or sets a threshold; this one subtracts, because Bladework replaces the ability D&D Beyond already used. The reason no precedent exists is that DDB natively handles the other substitution features — Hex Warrior, Battle Smith, Pact of the Blade — so Beyond20 never had to. Bladework isn't modelled there, hence the reconstruction. It infers the ability from weapon properties (Dexterity for ranged, better of Strength/Dexterity for Finesse, else Strength), so it will be wrong where something else overrides the ability, e.g. a Monk multiclass using Dexterity for monk weapons.
- **`usesTwoHandedDamage` slightly changes Charger 2024 behaviour.** The old inline check read `versatile-choice` directly and ignored `key_modifiers`; the shared helper follows `rollItem`'s own resolution, so the versatile hotkeys are now honoured. It also guards against `properties["Properties"]` being undefined, which the old expression did not. Both look like fixes to me, but they are behaviour changes in a Bladesong PR and easy to revert if you'd rather.
- The versions are separated by feature name (`Bladesong` vs `Bladesong 2024`), so a character can only ever match one. If `FRHoF` isn't the right marker, `is2024Source` is the single place to adjust.
- No changelog entry, since `docs/Changelog.md` appears to be written by release commits rather than feature PRs. Happy to add one if you'd prefer.
- Concentration still relies on the existing `confirm()` prompt, since D&D Beyond gives no way to distinguish a concentration save from any other Constitution save.

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -877,6 +877,12 @@ const character_settings = {
         "type": "bool",
         "default": false
     },
+    "wizard-bladesong-2024": {
+        "title": "Wizard: Bladesinger: Bladesong (2024)",
+        "description": "Activate your Bladesong: Advantage on Acrobatics checks, Intelligence for weapon attack and damage rolls, and Intelligence added to Concentration saves",
+        "type": "bool",
+        "default": false
+    },
     "empowered-evocation": {
         "title": "Wizard: Evocation Wizard: Empowered Evocation",
         "description": "Your prowess in Evocation lends power to your Evocation spells",
--- a/src/dndbeyond/base/character.js
+++ b/src/dndbeyond/base/character.js
@@ -286,6 +286,8 @@ class Character extends CharacterBase {
     getFeatureVersionName(feat_name, feat_reference) {
         if (!feat_reference) return feat_name;
         let is2024 = false;
+        // Sources using the 2024 rules whose reference doesn't say "2024"
+        const is2024Source = ["2024", "frhof"].some(src => feat_reference.toLowerCase().includes(src));
         if (
             (
                 feat_name.toLowerCase() === "great weapon master" ||
@@ -304,9 +306,9 @@ class Character extends CharacterBase {
                 feat_name.toLowerCase() === "remarkable athlete" ||
                 feat_name.toLowerCase() === "blessed strikes" ||
                 feat_name.toLowerCase() === "improved blessed strikes" ||
-                feat_name.toLowerCase() === "healer"
-            ) && ( 
-                feat_reference.toLowerCase().includes("2024")) ||
+                feat_name.toLowerCase() === "healer" ||
+                feat_name.toLowerCase() === "bladesong"
+            ) && is2024Source ||
                 feat_reference.toLowerCase().includes("free-rules")
          ) {
             is2024 = true;
--- a/src/dndbeyond/content-scripts/character.js
+++ b/src/dndbeyond/content-scripts/character.js
@@ -20,6 +20,59 @@ function addEffect(rollProperties, effect) {
     }
 }
 
+/**
+ * Which Bladesong version is currently active.
+ * @returns {(2014|2024|null)} The active version, or null if the Bladesong is not up
+ */
+function getActiveBladesong() {
+    if (character.hasClassFeature("Bladesong 2024") && character.getSetting("wizard-bladesong-2024", false))
+        return 2024;
+    if (character.hasClassFeature("Bladesong") && character.getSetting("wizard-bladesong", false))
+        return 2014;
+    return null;
+}
+
+/**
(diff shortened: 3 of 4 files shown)


dmportella:
thank you for your contribution, i am reviewing it and testing it, but with cursory look everything seems to look oke.

dmportella:
all good merging it

